PR #100, opened 2012-06-29
new SIP added: Fixing Either

robcd:
Dear Heather,

As threatened on the scala-debate thread by the same name, here's my suggestion for fixing scala.Either.

Best regards,

Rob


SethTisue:
this has actually been the subject of fresh discussion recently, but regardless, I'm closing to close the PR, so it doesn't show up in our queue. there's still strong interest in doing _something_ about Either for 2.13, so @robcd, if you're still around, consider reviving this

heathermiller:
Pinging @dickwall just to keep him in the loop about this (this PR was discussed in the most recent SIP/SLIP meeting btw, so we would like to do something about providing a right-biased variant of Either somehow)

dickwall:
I think left vs right bias is one of the first things to address in fact. For an infix type like Or (e.g. Person Or Error) it would make more sense to be left biased, right bias is the convention through Scala for either at present though because of the Right holding the "right" answer (and of course the conventions in languages like Haskell). I suspect this might be a long journey.

PR #206, opened 2013-05-31
Add the SIP-20 - proposal on lazy val initialization

axel22:


sjrd:
This SIP was last discussed during the July 13, 2016 SIP meeting. As of this date, @DarkDimius is taking over as "author" of this SIP, and @sjrd (myself) was assigned as reviewer.

The committee has the following feedback for the authors:
* The SIP document should be updated with the latest development [from the mailing list](https://groups.google.com/d/topic/scala-internals/4sjw8pcKysg/discussion), including with benchmarks of the new schemes.
* An implementation for scalac should be done.

Both tasks are assigned to @DarkDimius.

The next iteration of this SIP is scheduled 3 months from now, in the October 2016 SIP meeting, where it will be discussed further.

PR #213, opened 2013-06-30
Add Async SIP

phaller:
Adds first version of the Async SIP.

### DIFF
--- a/sips/pending/_posts/2013-06-30-async.md
+++ b/sips/pending/_posts/2013-06-30-async.md
@@ -1,0 +1,298 @@
+---
+layout: sip
+disqus: true
+title: SIP-22 - Async
+---
+
+**By: Philipp Haller and Jason Zaugg**
+
+## Introduction
+
+This is a proposal to add constructs that simplify asynchronous and concurrent programming in Scala. The main constructs, async and await, are inspired by similar constructs introduced in C# 5.0. The main purpose of async/await is to make it possible to express efficient asynchronous code in a familiar direct style (where suspending operations look as if they were blocking). As a result, non-blocking code using Scala’s futures API \[[1][1]\] can be expressed without using higher-order functions, such as map and flatMap, or low-level callbacks.
+
+On the level of types, async and await are methods with simple, intuitive types:
+
+    def async[T](body: => T): Future[T]
+    def await[T](future: Future[T]): T
+
+Here, `Future[T]` refers to the `Future` trait in package `scala.concurrent`. (The system can be adapted to other implementations of future-like abstractions; at the moment the API with the required extension points is internal, though.) The above methods are used as follows:
+
+    val fut = async {
+      slowComputation()
+    }
+
+The async construct marks a block of asynchronous code, and returns a future. Depending on the execution context in the implicit scope (see \[[1][1]\]), the block of asynchronous code is either executed on the current thread or in a thread pool. The async block can contain calls to await:
+
+    val futureDOY: Future[Response] =
+      WS.url("http://api.day-of-year/today").get
+
+    val futureDaysLeft: Future[Response] =
+      WS.url("http://api.days-left/today").get
+
+    val respFut = async {
+      val dayOfYear = await(futureDOY).body
+      val daysLeft = await(futureDaysLeft).body
+      Ok(s"$dayOfYear: $daysLeft days left!")
+    }
+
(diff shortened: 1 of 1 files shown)


xeno-by:
Hello @phaller and @retronym! 

Welcome to the new edition of the SIP process. I'm going to be your reviewer. This means that I'll be the contact person between you and the SIP committee, presenting your progress and providing you with our feedback.

During the SIP meeting on 13 Jul 2016, the committee has discussed your proposal. We think that async/await is a great thing to have in the Scala ecosystem, and that the quality of your implementation is very impressive!

Given our earlier in-person chat, we also considered your request to postpone this SIP to give you more time to work on the implementation of awaits in try/catch blocks. We agreed that this is a good idea and decided to officially postpone your SIP.

According to the new rules (http://docs.scala-lang.org/sips/sip-submission.html), postponement means that we set aside the SIP under some conditions. When those conditions are met, the SIP can be resubmitted. 

As a result, we're closing your SIP now, and here's the list of our requests for this SIP to be resubmitted again:
  - [ ] Enable awaits in try/catch blocks.
  - [ ] Provide better documentation for the user-facing part of your proposal.
  - [ ] Add the "Related Work" section that covers the latest developments of the async/await idea in other languages and explains whether/when you'd like to pursue them.

Looking forward to hearing from you again!

vn971:
Hi! First of all, I'm new to SIP-s internals and I've never hacked scalac myself, so feel free to correct me if I'm making some mistakes or missing something.

Now on topic. Is there any discussion/argumentation on why to change matters at all? 
Scala is strong by being general, by not following immediate needs and trying to find a general solution.

Deciding between scala/async being a separate library and included in stdlib, here are some concerns:

* the current solution of `async` being a library is clean enough. It's not a compiler plugin, It's just a normal dependency. A macro one. So we are not solving any architectural problem by moving.

* `async` has unresolved bugs, at least as I've heard. (Something related to closures?) Maybe I'm spreading the wrong information here though, I haven't heavily used `async` myself, yet. (I wanted to use it in production code, but disagreeing colleagues were a majority.)

* `async` may not be general enough for inclusion to stdlib. It does not support alternative `Future`-like implementations. Only solving one particular need. (No one wants to repeat the story `xml`.)

* moving `async` to stdlib might slow it's development, since releases would have to be in sync with releases of stdlib.

* people might get used to `async` as part of stdlib, which would stop some kinds of agressive changes to `async`. Is the fear of forever supporting a broken functionality a real one?

retronym:
@vn971 The proposal is not to include `async` into `scala-library.jar`, we forsee it remaining a separate module. But it is within the scala namespace, and once it reaches 1.0, will follow the backwards compatibility constraints that that entails. 

`async` has a few limitations we'd like to remove in the future. I would like to support use of `await` in `try/catch`, this is "just" some more work to do, not a fundamental restriction. 

`async` cannot by its nature be nested in a closure, but we could compose the async macro with another macro that inlines some commonly used higher order functions (like `{Option, List}#(map, ...)}` to rewrite calls into imperative control flow before running the async transform. I personally believe that this should not be a blocker for async 1.0.

`async` is internally architected so that you can create different versions of it for alternative future systems. But we only ship a version of the macro that works with `scala.concurrent.Future`. This is an important use case.

vn971:
@retronym thank you for your time.

What is this SIP, then? AFAIunderstood, SIP should be a "Scala improvement process" for the language. So if it's just a library, why call it SIP? What am I missing here?

fwbrasil:
I echo @vn971's concerns, `scala-async` it too specific to be a SIP.

I do think that a more generic solution would be of great benefit to the Scala ecosystem and should be considered as a built-in language feature. The transformation could be used as syntactic sugar to deal with any monad-like interface similarly to for-comprehensions, translating the code to a normal monad composition without a state machine. See [effectful](https://github.com/pelotom/effectful) as an example, it generalizes the transformation for any Scalaz monad.

Just as an example, this is a prototype of what I have in mind:

Library interface
```scala
trait Transformer[M[_]] {

  def value[T](v: T): M[T]
  
  /* Other methods like `handle` and `recover` 
   * would be necessary to desugar some 
   * language constructs */

  def lift[T](body: T): M[T] = macro Macro.lift[M, T]

  @compileTimeOnly("`unlift` must be used within `lift`")
  def unlift[T](m: M[T]): T = ???
}
```

Usage for `Future`
```scala
  val m = new Transformer[Future] {
    def value[T](v: T) = Future.successful(v)
  }

  import t._

  val f1: Future[Boolean] = ???
  val f2: Future[Int] = ???

  val a: Future[Int] = lift {
    val b = unlift(f1)
    if (b) unlift(f2) else 11
  }
```

`a` would be translated to:
```scala
val a: Future[Int] =
  f1.flatMap { b =>
    if(b) f2
    else t.value(11)
  }
```

Notes:

- The execution context implicit resolution would happen naturally after the macro expansion.
- Only `flatMap` and `map` would be required to be implemented by the monad instance, other methods should be defined by the `Transformer` instance.

SethTisue:
> So if it's just a library, why call it SIP

Back in 2013, SLIPs hadn't been split off from SIPs yet.

If or when this proposal progresses any further, it will be as a SLIP.

PR #265, opened 2013-10-28
Moved to standard notation in performance characteristics docs

SamuelMarks:
Notation used is now the standard computer-science one; a.k.a.: "Big-O" Notation from the Bachmann-Landau family.

Also allowed me to remove the legend table; as the notation is standard. Additionally added in `<sup>` for the more specific amoritized complexities and the like. (often represented with `*` notation, but this'll suffice)


SethTisue:
closing because this appears abandoned. if that's wrong, feel free to address Heather's feedback and reopen

SamuelMarks:
@heathermiller what's the command you use to generate the documentation? - Maybe they've updated their `gem` since…

PR #295, opened 2014-02-03
Added a new SIP about possible extensions of futures and promises.

EddieJamsession:
SIP-23: Extension of Futures and Promises


alexandru:
Hello, I found out about this SIP and I have some suggestions that I've written here: https://gist.github.com/alexandru/3de75c350c710161dc80

jvican:
Since the new [SIP process has been implemented](http://docs.scala-lang.org/sips/sip-submission.html), this current proposal doesn't apply anymore as a SIP. SIPs are supposed to propose changes in the main utilities that the compiler depends on. Thus, I close it. The discussion will continue [in the slip repository for the moment being](https://github.com/scala/slip/issues/7).

PR #303, opened 2014-02-19
Clear up blocking _in_ a future vs blocking _on_ a future.

backuitist:


SethTisue:
this is an edit to SIP-14. could one or more of the SIP's authors review it? @phaller @axel22 @heathermiller @viktorklang @viktorklang @rkuhn @vjovanov

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,30 @@ Invoking the `future` construct uses a global execution context to start an asyn
 Support for extending the Futures API with additional utility methods is planned.
 This will allow external frameworks to provide more specialized utilities.
 
+
 ## Blocking
 
+### in a Future
```
axel22: Please put a capital `I` in `In` here.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,30 @@ Invoking the `future` construct uses a global execution context to start an asyn
...
+      blocking { blockingStuff() }
+    }
+
+Is equivalent to
+ 
+    Future { blockingStuff() }
+
+The blocking code may also throw an exception. In this case, the exception is forwarded to the caller.
+
+### on a Future
```
axel22: Please capitalize.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,30 @@ Invoking the `future` construct uses a global execution context to start an asyn
...
+implement `blocking` by means of `ManagedBlocker`, some just do nothing:
+
+    implicit val ec = ExecutionContext.fromExecutor(
+                        Executors.newFixedThreadPool(4))
+    Future {
+      // here blocking serves only for documentation purpose
+      blocking { blockingStuff() }
+    }
+
+Is equivalent to
```
axel22: "Can be equivalent to:"

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,30 @@ Invoking the `future` construct uses a global execution context to start an asyn
...
 This will allow external frameworks to provide more specialized utilities.
 
+
 ## Blocking
 
+### in a Future
+
+As seen with the global `ExecutionContext`, it is possible to notify an `ExecutionContext` of a blocking call with the `blocking` construct.
+The implementation is however at the complete discretion of the `ExecutionContext`. While some `ExecutionContext` such as `ExecutionContext.global`
+implement `blocking` by means of `ManagedBlocker`, some just do nothing:
```
axel22: ...by means of _a_ `ManagedBlocker`.

overviews/core/_posts/2012-09-20-futures.md
axel22: I would maybe write `Blocking inside a future`.

overviews/core/_posts/2012-09-20-futures.md
axel22: I would maybe write `Blocking outside the future`.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,30 @@ Invoking the `future` construct uses a global execution context to start an asyn
 Support for extending the Futures API with additional utility methods is planned.
 This will allow external frameworks to provide more specialized utilities.
 
+
 ## Blocking
```
axel22: Please add a short one-two lines introductory paragraph - something like the following:

    Futures are generally asynchronous and do not block the underlying execution threads.
    However, in certain cases, it is necessary to block.
    We distinguish two forms of blocking the execution thread:
    invoking arbitrary code that blocks the thread from within the future,
    and blocking from outside another future, waiting until that future gets completed..

axel22:
Thanks a lot for your contribution and effort to clear this up!

LGTM overall, please address the few details I commented.

Btw, there is another futures doc where this should be added:

http://docs.scala-lang.org/overviews/core/futures.html

SethTisue:
> there is another futures doc where this should be added

great point... perhaps the SIP itself should be left alone, as a historical document — it's actually more important to update the document you link to

SethTisue:
@axel22 sorry, I was confused and managed to confuse you as well. this text derives from SIP-14, but it is not the SIP itself. the PR does address the right page.

axel22:
@SethTisue even better, then, less work for the @backuitist !

overviews/core/_posts/2012-09-20-futures.md
backuitist: I think that would be confusing to say "can be". My point is that in this situation (ExecutionContext backed by an ExecutorService) the `blocking` block isn't doing anything. So from the JVM standpoint it _is_ equivalent. It's like saying: `(); 42` is equivalent to `42`.
Maybe I could write `Is equivalent at runtime to`? Or `has the same effect as`?

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,38 @@ Invoking the `future` construct uses a global execution context to start an asyn
...
+We distinguish two forms of blocking the execution thread:
+invoking arbitrary code that blocks the thread from within the future,
+and blocking from outside another future, waiting until that future gets completed.
+
+
+### Blocking inside a Future
+
+As seen with the global `ExecutionContext`, it is possible to notify an `ExecutionContext` of a blocking call with the `blocking` construct.
+The implementation is however at the complete discretion of the `ExecutionContext`. While some `ExecutionContext` such as `ExecutionContext.global`
+implement `blocking` by means of a `ManagedBlocker`, some just do nothing:
```
viktorklang: Perhaps be clear why this is the case:

"some will do nothing, like Executors.newFixedThreadPool, as shown below:"

backuitist: This paragraph is about `ExecutionContext` and `Executors.newFixedThreadPool` isn't one. You're talking about `ExecutionContext.fromExecutor(Executors.newFixedThreadPool(x))` which in this sentence is overkill IMO.

axel22: You could put that long expression into a separate line - like this:

    Some execution contexts, such as the fixed thread pool:
    
        ... long expression ....

    will do nothing, as shown in the following:

backuitist: It's a bit redundant IMO but if you think that helps I'll add it.

SethTisue:
@viktorklang ready to merge?

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -615,8 +615,42 @@ Invoking the `future` construct uses a global execution context to start an asyn
...
+implement `blocking` by means of a `ManagedBlocker`, some execution contexts such as the fixed thread pool: 
+
+    ExecutionContext.fromExecutor(Executors.newFixedThreadPool(x))
+
+will do nothing, as shown in the following:
+
+    implicit val ec = ExecutionContext.fromExecutor(
+                        Executors.newFixedThreadPool(4))
+    Future {
+      // here blocking serves only for documentation purpose
```
viktorklang: Rereading the PR,
I think it is important to let the reader know that generally speaking: code *using* an ExecutionContext should not know implementation details regarding the ExecutionContext, and therefore any code that invokes `blocking` does so to let the ExecutionContext deal with the situation in the best way it knows how to. I.e. it does not serve as documentation purpose. (If the readers think that it is somehow optional and only serves as documentation, then it's not really going to be used.)

Does that line of thinking make sense?

viktorklang:
@SethTisue Thanks for the ping. I have one final comment (added).

overviews/core/_posts/2012-09-20-futures.md
backuitist: Hi Viktor,
I was actually quoting you when I wrote that `blocking` serves for documentation purpose: https://groups.google.com/d/msg/scala-user/LKiVD2PGvew/CWufs2Z8K4UJ

>> If I use `ExecutionContext.fromExecutor(Executor)` the blocking construct will be made useless.

> It's not useless. not only can you install your own BlockContext but it serves as "documentation" so that if someone changes the EC implementation, your code can be made to run better.

viktorklang: Hi Bruno,

Yes, in the case that the implementation behind the EC (or the currently installed BlockContext to be more precise) does not do anything, it is "effectively" documentation, but given that the EC typically is passed in, it may not be like that for all invocations.

My reservation is against making it sound like it is something that only serves documentation purposes, even if the EC implementation is known, because if the implementation changes in the future, it will be hard to track the blocking if that wasn't added when the code was written.

Does that make sense? :)

backuitist: I haven't really changed my point of view on this matter since last year. IMO if you want to block you don't do it through a `blocking` call but rather you'd use an EC that has been designed for that, such as a TPE with a large pool.

> if the implementation changes in the future
Do you really believe that we'll see in the (foreseeable) future an EC implementation for TPE that can leverage `blocking`?

In the mean time, this feature is dangerous and the least you can do is warn your users. I believe this PR addresses that perfectly.
Maybe @axel22 or @SethTisue could have a word on it?

axel22: I don't have a very strong opinion. Maybe we can omit the line which says that it's for documentation, and just say that it informs the EC so that it can optionally adjust its parallelism.

viktorklang: >I haven't really changed my point of view on this matter since last year. IMO if you want to block you don't do it through a blocking call but rather you'd use an EC that has been designed for that, such as a TPE with a large pool.

I disagree with that argument—if every library utility should define its own EC for blocking, what has been solved? ECs should to the largest extent be required from the outside in the form of an implicit parameter, such that it can be threaded through.
Code should be written without knowing the concrete implementation, so any method that needs to block should ideally call `blocking` such that if it is executed within an EC. It is simply not possible to wrap the invocation of any such method in a blocking-block from the outside, as the blocking will occur on another thread. Does that make sense?

>Do you really believe that we'll see in the (foreseeable) future an EC implementation for TPE that can leverage blocking?

I have a PoC?
https://github.com/akka/akka/commit/5e5e38747449ad1798f8aae4a1b5e5989cac5eda

>In the mean time, this feature is dangerous and the least you can do is warn your users. I believe this PR addresses that perfectly.

Which feature are you referring to? My only remaining concern is that readers get the impression that it is about documentation, it's not, it's about demarcation :)

backuitist: > I disagree with that argument—if every library utility should define its own EC for blocking, what has been solved? 

You're right, the library should not instantiate its own EC, but that's not what I said. Libraries needing a blocking EC can merely declare it as a dependency, and if you want to leverage implicits you can do something along those lines:

    trait BlockingEC { def delegate : ExecutionContext }
    def doBlockingAndNonBlockingStuff(implicit blockingEc: BlockingEC, ec: ExecutionContext)

I find this safer than:

    /** IMPORTANT: ExecutionContext must support `blocking` or this method will crash your app with deadlocks */
    def doBlockingAndNonBlockingStuff(implicit ec: ExecutionContext)

As a matter of fact I believe that, as a general rule, implicit parameters are fine as long as they have a very specific meaning. `blocking` feels like a hack to work around the fact that we only have one kind of ExecutionContext, not 2.
But that's just my opinion, and for now I'm just suggesting that we make it clear that `blocking` might not work as expected under certain conditions. 

> I have a PoC?

I took a quick look. That's nice, but that's not yet in the scala-library and, as far as I understand, that only works if you create your TPE-based EC through your ThreadPoolConfig. If someone has a TPE at hand (provided by some other code) and wants to wrap it in an EC through `fromExecutor` then you're stuck.

> Which feature are you referring to?

`blocking`

viktorklang: @backuitist 

>You're right, the library should not instantiate its own EC, but that's not what I said. Libraries needing a blocking EC can merely declare it as a dependency, and if you want to leverage implicits you can do something along those lines:
>[...]
>As a matter of fact I believe that, as a general rule, implicit parameters are fine as long as they have a very specific meaning. blocking feels like a hack to work around the fact that we only have one kind of ExecutionContext, not 2.
But that's just my opinion, and for now I'm just suggesting that we make it clear that blocking might not work as expected under certain conditions.

I can understand that argument, I think we discussed having multiple ones initially but came to the conclusion that it was disruptive enough to have the ripple-effect of having to add `(implicit ec: ExecutionContext)` up the call chain as soon as something async needed to be done down the call chain. Adding a new dimension where one also needed to add yet another implicit as soon as blocking was involved would have serious impact on binary and source compatibility. (Imagine having to add a blocking call, now all your signatures would change, so people would just do unmanaged blocking instead, which would mean that they'd block on the provided EC, without any means of parrying for it).

My long-standing plan has been to create a thread pool that would work well for both CPU-bound work á la FJP but also blocking work á la TPE, transitioning workers between those modes based on `blocking`/`BlockContext`. What would be fantastic about that is that code that uses `blocking` would simply work with the new EC without any modification.

>I took a quick look. That's nice, but that's not yet in the scala-library and, as far as I understand, that only works if you create your TPE-based EC through your ThreadPoolConfig. If someone has a TPE at hand (provided by some other code) and wants to wrap it in an EC through fromExecutor then you're stuck.

Of course, it is not a virus that infects other's object instances :)

backuitist: > My long-standing plan has been to create a thread pool that would work well for both CPU-bound work á la FJP but also blocking work á la TPE, transitioning workers between those modes based on blocking/BlockContext. What would be fantastic about that is that code that uses blocking would simply work with the new EC without any modification.

That's fair, but how about leaving this PR as it is now and updating it once the scala-library gets upgraded with your silver bullet EC?

viktorklang: @backuitist Nice try ;)

Since there is nothing that prevents, nor precludes that such a threadpool already exists and is used in production daily, I think this PR should be amended such that it is clear that the primary thing about `blocking` is about demarcation rather than mere documentation :)

Everything else looks fantastic, it's a great contribution!

backuitist: @viktorklang I surrender! ;)

How should we go about it? Do like @axel22 proposal? Would that work for you to remove that line?

viktorklang: Deal, I think @axel22's proposal is a good middle ground.

SethTisue:
thank you @backuitist for the submission and for getting it past our tough pack of reviewers :-)

PR #304, opened 2014-02-19
Execution context

backuitist:
I've integrated some of your comments ;)

### DIFF
--- a/overviews/core/_posts/2012-09-20-futures.md
+++ b/overviews/core/_posts/2012-09-20-futures.md
@@ -13,9 +13,9 @@ languages: [ja]
 
 Futures provide a nice way to reason about performing many operations
 in parallel-- in an efficient and non-blocking way. The idea
-is simple, a `Future` is a sort of a placeholder object that you can
-create for a result that does not yet exist. Generally, the result of
-the `Future` is computed concurrently and can be later collected.
+is simple, a [Future](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future) 
+is a sort of a placeholder object that you can create for a result that does not yet exist. 
+Generally, the result of the `Future` is computed concurrently and can be later collected.
 Composing concurrent tasks in this way tends to result in faster, asynchronous, non-blocking parallel code.
 
 By default, futures and promises are non-blocking, making use of
@@ -35,6 +35,134 @@ promises are created through such `ExecutionContext`s. For example, this makes i
 in the case of an application which requires blocking futures, for an underlying execution 
 environment to resize itself if necessary to guarantee progress.
 -->
+
+A typical future look like this:
+
+    val inverseFuture : Future[Matrix] = Future {
+      fatMatrix.inverse() // non-blocking long lasting computation
+    }(executionContext)
+
+
+Or with the more idiomatic
+
+    implicit val ec : ExecutionContext = ...
+    val inverseFuture : Future[Matrix] = Future {
+      fatMatrix.inverse()
+    } // ec is implicitly passed
+
+
+Both code snippets delegate the execution of `fatMatrix.inverse()` to an `ExecutionContext` and embody the result of the computation in `inverseFuture`.
+
+
+## Execution Context   
+
+Future and Promises revolve around [ExecutionContexts](http://www.scala-lang.org/api/current/index.html#scala.concurrent.ExecutionContext), responsible for executing computations.
+
+An `ExecutionContext` is similar to an [Executor](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html): it is free to execute  computations in a new thread, in a pooled thread or in the current thread (although this is discouraged - more on that below).
+
+The `scala.concurrent` package comes out of the box with an `ExecutionContext` implementation, a global static thread pool.
+It is also possible to adapt an `Executor` into an `ExecutionContext`, and you are of course free to implement your own. 
+
+
+### The Global Execution Context
+
+`ExecutionContext.global` is an `ExecutionContext` backed by a [ForkJoinPool](http://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html). It should be sufficient for most situations but requires some care.  
+A `ForkJoinPool` manages a limited amount of threads (the maximum amount of thread being refered to as parallelism level). That limit can be exceeded in the presence of blocking computation (the pool must be notified though). 
+
+By default the `ExecutionContext.global` set the parallelism level of its underlying fork-join pool to the amount of available processors ([Runtime.availableProcessors](http://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#availableProcessors%28%29)).
+This configuration can be overriden by setting one (or more) of the following VM attribute:
+
+  * scala.concurrent.context.minThreads - defaults to `Runtime.availableProcessors`
+  * scala.concurrent.context.numThreads - can be a number or a multiplier (N) in the form 'xN' ;  defaults to `Runtime.availableProcessors`
+  * scala.concurrent.context.maxThreads - defaults to `Runtime.availableProcessors`
+
+The parallelism level will be set to `numThreads` as long as it remains within `[minThreads; maxThreads]`.
+
+As stated above the `ForkJoinPool` can increase the amount of threads beyond its `parallelismLevel` in the presence of blocking computation. As explained in the `ForkJoinPool` API, this is only possible if the pool is explicitly notified:
+
+    import scala.concurrent.Future
+    import scala.concurrent.forkjoin._
+
+    // the following is equivalent to `implicit val ec = ExecutionContext.global`
+    import ExecutionContext.Implicits.global
+
+    Future {
+      ForkJoinPool.managedBlock(
+        new ManagedBlocker {
+           var done = false
+
+           def block(): Boolean = {
+             try {
+               myLock.lock()
+               // ...
+             } finally {
+              done = true
+             }
+             true
+           }
+
+           def isReleasable: Boolean = done
+        }
+      )
+    }
+
+
+Fortunately the concurrent package provides a convenient way for doing so:
+
+    import scala.concurrent.Future
+	  import scala.concurrent.blocking
+
+    Future {
+      blocking {
+        myLock.lock()
+        // ...
+      }
+    }
+
+Note that `blocking` is a general construct that will be discussed more in depth [below](#in_a_future).
+
+Last but not least, you must remember that the `ForkJoinPool` isn't designed for long lasting blocking operations. Even when notified with `blocking` the pool might not spawn new workers as you would expect, and when new workers are created they can be as many as 32767. 
+To give you an idea, the following code will use 32000 threads:
+
+    implicit val ec = ExecutionContext.global
+
+    for( i <- 1 to 32000 ) {
+      Future {
+        blocking {
+          Thread.sleep(999999)		
+        }
+      }
+    }
+
+
+If you need to wrap long lasting blocking operations we recommend using a dedicated `ExecutionContext`, for instance by wrapping a Java `Executor`.
+   
+
+### Adapting Java Executor
+
+Using `ExecutionContext.fromExecutor` you can adapt a Java `Executor` into an `ExecutionContext`.  
+For instance:
+
+    ExecutionContext.fromExecutor(new ThreadPoolExecutor( /* your configuration */ ))
+
+
+### Synchronous Execution Context
+
+One might be tempted to have an `ExecutionContext` that runs computations within the current thread:
+
+    val currentThreadExecutionContext = ExecutionContext.fromExecutor(
+      new Executor {
+        def execute(runnable: Runnable) { runnable.run() }
+    })
+
+This should be avoided as it introduce non-determinism in the execution of your future.
+
+    Future { doSomething }(ExecutionContext.global).map { doSomethingElse }(currentThreadExecutionContext)
+
+`doSomethingElse` might either execute in `doSomething`'s thread or in the main thread, and therefore be either asynchronous or synchronous.
+As explained [here](http://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/) a callback shouldn't be both.
+
+
 
 ## Futures
 
--- a/overviews/core/_posts/2012-09-20-futures.md
+++ b/overviews/core/_posts/2012-09-20-futures.md
@@ -11,11 +11,11 @@ languages: [ja]
 
 ## Introduction
 
-Futures provide a nice way to reason about performing many operations
-in parallel-- in an efficient and non-blocking way. The idea
-is simple, a [Future](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future) 
-is a sort of a placeholder object that you can create for a result that does not yet exist. 
-Generally, the result of the `Future` is computed concurrently and can be later collected.
+Futures provide a way to reason about performing many operations
+in parallel-- in an efficient and non-blocking way. 
+A [`Future`](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future)
+is a placeholder object for a value that may not yet exist.
+Generally, the value of the Future is supplied concurrently and can subsequently be used.
 Composing concurrent tasks in this way tends to result in faster, asynchronous, non-blocking parallel code.
 
 By default, futures and promises are non-blocking, making use of
@@ -36,16 +36,16 @@ in the case of an application which requires blocking futures, for an underlying
 environment to resize itself if necessary to guarantee progress.
 -->
 
-A typical future look like this:
+A typical future looks like this:
 
-    val inverseFuture : Future[Matrix] = Future {
+    val inverseFuture: Future[Matrix] = Future {
       fatMatrix.inverse() // non-blocking long lasting computation
     }(executionContext)
 
 
-Or with the more idiomatic
+Or with the more idiomatic:
 
-    implicit val ec : ExecutionContext = ...
+    implicit val ec: ExecutionContext = ...
     val inverseFuture : Future[Matrix] = Future {
       fatMatrix.inverse()
     } // ec is implicitly passed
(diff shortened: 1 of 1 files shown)


SethTisue:
I believe you want to be updating http://docs.scala-lang.org/overviews/core/futures.html, not the SIP page?

SethTisue:
I got confused and thought this targeted the wrong file, but I see now it is right.

SethTisue:
review by @viktorklang and @axel22?

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
 environment to resize itself if necessary to guarantee progress.
 -->
 
+A typical future look like this:
+
+    val inverseFuture : Future[Matrix] = Future {
```
axel22: We normally write the type ascription immediately after the identifier, without the space:

    val inverseFuture: Future[Matrix] = ...

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+      fatMatrix.inverse() // non-blocking long lasting computation
+    }(executionContext)
+
+
+Or with the more idiomatic
+
+    implicit val ec : ExecutionContext = ...
```
axel22: Same here.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+      fatMatrix.inverse()
+    } // ec is implicitly passed
+
+
+Both code snippets delegate the execution of `fatMatrix.inverse()` to an `ExecutionContext` and embody the result of the computation in `inverseFuture`.
+
+
+## Execution Context   
+
+Future and Promises revolve around [ExecutionContexts](http://www.scala-lang.org/api/current/index.html#scala.concurrent.ExecutionContext), responsible for executing computations.
```
axel22: around [`ExecutionContext`s](...)

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -11,11 +11,11 @@ languages: [ja]
...
 ## Introduction
 
-Futures provide a nice way to reason about performing many operations
-in parallel-- in an efficient and non-blocking way. The idea
-is simple, a `Future` is a sort of a placeholder object that you can
-create for a result that does not yet exist. Generally, the result of
-the `Future` is computed concurrently and can be later collected.
+Futures provide a way to reason about performing many operations
+in parallel-- in an efficient and non-blocking way. 
+A [Future](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future)
```
axel22: Not at all, but you probably want:

    A [`Future`](...)

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+
+Both code snippets delegate the execution of `fatMatrix.inverse()` to an `ExecutionContext` and embody the result of the computation in `inverseFuture`.
+
+
+## Execution Context   
+
+Future and Promises revolve around [ExecutionContexts](http://www.scala-lang.org/api/current/index.html#scala.concurrent.ExecutionContext), responsible for executing computations.
+
+An `ExecutionContext` is similar to an [Executor](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html): it is free to execute  computations in a new thread, in a pooled thread or in the current thread (although this is discouraged - more on that below).
```
axel22: Please break the long line to several lines.

axel22: (although executing the computation in the current thread is discouraged -- more on that below).

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+The `scala.concurrent` package comes out of the box with an `ExecutionContext` implementation, a global static thread pool.
+It is also possible to adapt an `Executor` into an `ExecutionContext`, and you are of course free to implement your own.
```
axel22: It is also possible to convert an `Executor` into an `ExecutionContext`.
    Finally, users are free to extend the `ExecutionContext` trait to implement their own execution contexts,
    although this should only be done in rare cases.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+An `ExecutionContext` is similar to an [Executor](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html): it is free to execute  computations in a new thread, in a pooled thread or in the current thread (although this is discouraged - more on that below).
+
+The `scala.concurrent` package comes out of the box with an `ExecutionContext` implementation, a global static thread pool.
+It is also possible to adapt an `Executor` into an `ExecutionContext`, and you are of course free to implement your own. 
+
+
+### The Global Execution Context
+
+`ExecutionContext.global` is an `ExecutionContext` backed by a [ForkJoinPool](http://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html). It should be sufficient for most situations but requires some care.
```
axel22: I don't think there is a license problem with pointing back to Oracle's Java documentation.

Please break these two sentences into two lines.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+An `ExecutionContext` is similar to an [Executor](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html): it is free to execute  computations in a new thread, in a pooled thread or in the current thread (although this is discouraged - more on that below).
+
+The `scala.concurrent` package comes out of the box with an `ExecutionContext` implementation, a global static thread pool.
+It is also possible to adapt an `Executor` into an `ExecutionContext`, and you are of course free to implement your own. 
+
+
+### The Global Execution Context
+
+`ExecutionContext.global` is an `ExecutionContext` backed by a [ForkJoinPool](http://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html). It should be sufficient for most situations but requires some care.  
+A `ForkJoinPool` manages a limited amount of threads (the maximum amount of thread being refered to as parallelism level). That limit can be exceeded in the presence of blocking computation (the pool must be notified though).
```
axel22: "referred"

axel22: ...as *parallelism level*).

axel22: ... of blocking computations...

axel22: Instead of the comment in parenthesis, I would write a new sentence:

    The number of concurrently blocking computations can exceed the parallelism level
    only if each blocking call is wrapped inside a `blocking` call (more on that below).
    Otherwise, there is a risk that the thread pool in the global execution context is starved,
    and no computation can proceed.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+The `scala.concurrent` package comes out of the box with an `ExecutionContext` implementation, a global static thread pool.
+It is also possible to adapt an `Executor` into an `ExecutionContext`, and you are of course free to implement your own. 
+
+
+### The Global Execution Context
+
+`ExecutionContext.global` is an `ExecutionContext` backed by a [ForkJoinPool](http://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html). It should be sufficient for most situations but requires some care.  
+A `ForkJoinPool` manages a limited amount of threads (the maximum amount of thread being refered to as parallelism level). That limit can be exceeded in the presence of blocking computation (the pool must be notified though). 
+
+By default the `ExecutionContext.global` set the parallelism level of its underlying fork-join pool to the amount of available processors ([Runtime.availableProcessors](http://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#availableProcessors%28%29)).
```
axel22: ...sets...

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+This configuration can be overriden by setting one (or more) of the following VM attribute:
```
axel22: ...attributes:

overviews/core/_posts/2012-09-20-futures.md
axel22: I'm having second thoughts about mentioning FJ pool here.
The current `ExecutionContext.global` implementation, even though it's unlikely to change,
should be mentioned at most in the ScalaDoc API.
Otherwise, our documentation is too flaky, and it becomes hard to maintain the crossreferences.

This doc should discuss only the high-level usage and meaning of execution contexts,
not the specific implementation.
It is also unclear if we should mention `minThreads`, `numThreads` and `maxThreads` configuration settings,
since they are not mentioned explicitly in the API docs.

axel22: I would suggest that you refactor this section, so that all mention of `ForkJoinPool` is put in a separate section,
entitled e.g. "Execution Contexts based on Fork/Join Pools".

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+    Future {
+      blocking {
+        myLock.lock()
+        // ...
+      }
+    }
+
+Note that `blocking` is a general construct that will be discussed more in depth [below](#in_a_future).
+
+Last but not least, you must remember that the `ForkJoinPool` isn't designed for long lasting blocking operations. Even when notified with `blocking` the pool might not spawn new workers as you would expect, and when new workers are created they can be as many as 32767.
```
axel22: is not

axel22: Please break into several lines.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+      blocking {
+        myLock.lock()
+        // ...
+      }
+    }
+
+Note that `blocking` is a general construct that will be discussed more in depth [below](#in_a_future).
+
+Last but not least, you must remember that the `ForkJoinPool` isn't designed for long lasting blocking operations. Even when notified with `blocking` the pool might not spawn new workers as you would expect, and when new workers are created they can be as many as 32767. 
+To give you an idea, the following code will use 32000 threads:
```
axel22: This is perhaps too detailed, and concerns standing `ForkJoinPool` implementations more than execution contexts. I would omit this snippet.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+          Thread.sleep(999999)		
+        }
+      }
+    }
+
+
+If you need to wrap long lasting blocking operations we recommend using a dedicated `ExecutionContext`, for instance by wrapping a Java `Executor`.
+   
+
+### Adapting Java Executor
```
axel22: Adapting _a_ Java `Executor`

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+Using `ExecutionContext.fromExecutor` you can adapt a Java `Executor` into an `ExecutionContext`.
```
axel22: Using the `...` method, you can adapt ...

Consider using `convert` instead of `adapt`.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+### Synchronous Execution Context
+
+One might be tempted to have an `ExecutionContext` that runs computations within the current thread:
+
+    val currentThreadExecutionContext = ExecutionContext.fromExecutor(
+      new Executor {
+        def execute(runnable: Runnable) { runnable.run() }
+    })
+
+This should be avoided as it introduce non-determinism in the execution of your future.
```
axel22: ...it introduces...

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+    Future { doSomething }(ExecutionContext.global).map { doSomethingElse }(currentThreadExecutionContext)
```
axel22: Consider breaking into several lines for readability, and maintainability.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+
+`doSomethingElse` might either execute in `doSomething`'s thread or in the main thread, and therefore be either asynchronous or synchronous.
```
axel22: The `doSomething` method might either...

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+As explained [here](http://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/) a callback shouldn't be both.
```
axel22: Even though the blogger is in this case an active Scala community member, I'd argue against linking information to blogs, however accurate it may be.

Instead, I'd first open a documentation section in ScalaDocs about concurrency and/or programming paradigm terms in general, and then link that (since, we, as a Scala community then get to maintain it and ensure that it is current and accurate).

axel22: ...should not ....

axel22:
Thanks a lot for taking the effort to improve the documentation to this extent!
Please address my comments before we can merge this in.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -11,11 +11,11 @@ languages: [ja]
...
 
-Futures provide a nice way to reason about performing many operations
-in parallel-- in an efficient and non-blocking way. The idea
-is simple, a `Future` is a sort of a placeholder object that you can
-create for a result that does not yet exist. Generally, the result of
-the `Future` is computed concurrently and can be later collected.
+Futures provide a way to reason about performing many operations
+in parallel-- in an efficient and non-blocking way. 
+A [Future](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future) 
+is a sort of a placeholder object that you can create for a result that does not yet exist.
```
viktorklang: I'd suggest:

`is a placeholder object for a value that may not yet exist.`

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -11,11 +11,11 @@ languages: [ja]
...
-Futures provide a nice way to reason about performing many operations
-in parallel-- in an efficient and non-blocking way. The idea
-is simple, a `Future` is a sort of a placeholder object that you can
-create for a result that does not yet exist. Generally, the result of
-the `Future` is computed concurrently and can be later collected.
+Futures provide a way to reason about performing many operations
+in parallel-- in an efficient and non-blocking way. 
+A [Future](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future) 
+is a sort of a placeholder object that you can create for a result that does not yet exist. 
+Generally, the result of the `Future` is computed concurrently and can be later collected.
```
viktorklang: Perhaps: `Generally, the value of the `Future` is supplied concurrently and can subsequently be used.`

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
 environment to resize itself if necessary to guarantee progress.
 -->
 
+A typical future look like this:
```
viktorklang: That's not how a Future looks, I think it is important to distinguish between Future-the-object and Future.apply-the async-computation-resulting-in-a-Future-the-object.

I.e.

val f = Future.successful(())
vs
val f = Future(())

overviews/core/_posts/2012-09-20-futures.md
viktorklang: I recommend making the example take the EC as an implicit parameter to foster a culture of getting the EC passed in rather than defining it locally.

overviews/core/_posts/2012-09-20-futures.md
viktorklang: `or, which is discouraged, in the current thread (more on that below)`

overviews/core/_posts/2012-09-20-futures.md
viktorklang: Jus ta heads-up: `as many as 32767` is not true for 2.12: https://github.com/scala/scala/blob/2.12.x/src/library/scala/concurrent/impl/ExecutionContextImpl.scala#L115

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
+    for( i <- 1 to 32000 ) {
+      Future {
+        blocking {
+          Thread.sleep(999999)		
+        }
+      }
+    }
+
+
+If you need to wrap long lasting blocking operations we recommend using a dedicated `ExecutionContext`, for instance by wrapping a Java `Executor`.
```
viktorklang: FJP is a Java Executor so I am unsure how this recommendation will help. Thoughts?

overviews/core/_posts/2012-09-20-futures.md
viktorklang: Or perhaps "lift" or "wrap"?

overviews/core/_posts/2012-09-20-futures.md
viktorklang: I'd suggest adding a comment after the `execute` method declaration to clearly state not to do this, so that careless copy-pasting still retains the fact that it is broken.

overviews/core/_posts/2012-09-20-futures.md
viktorklang: I'd suggest to encourage the reader to get their EC from an implicit parameter for the normal case.

overviews/core/_posts/2012-09-20-futures.md
phaller: It might be helpful to point out that `blocking` provides an opportunity to the underlying thread pool to spawn spare workers *temporarily*, i.e., spare workers terminate when possible. Moreover, the behavior is typically much better-behaved than with, e.g., a JDK `Executors.newCachedThreadPool()`.

overviews/core/_posts/2012-09-20-futures.md
phaller: Perhaps what was meant here is to create a *dedicated* execution context just for executing blocking code, so that non-blocking code running on a separate execution context is not influenced negatively.

overviews/core/_posts/2012-09-20-futures.md
axel22: Agree - the more idiomatic version below could be sufficient.

overviews/core/_posts/2012-09-20-futures.md
```diff
@@ -36,6 +36,134 @@ in the case of an application which requires blocking futures, for an underlying
...
 -->
 
+A typical future look like this:
+
+    val inverseFuture : Future[Matrix] = Future {
+      fatMatrix.inverse() // non-blocking long lasting computation
+    }(executionContext)
+
+
+Or with the more idiomatic
```
axel22: A colon should be at the end.

overviews/core/_posts/2012-09-20-futures.md
axel22: Even better!

SethTisue:
@backuitist interested in getting this across the finish line...?

backuitist:
Sorry, that went off my radar.

overviews/core/_posts/2012-09-20-futures.md
backuitist: That was indeed the idea.

overviews/core/_posts/2012-09-20-futures.md
backuitist: Thanks for the link Viktor. That change fixes the problem in 2.12 but you might want to keep these comments for your 2.11 users? Maybe once 2.12 is out you can amend that documentation and say that this is now under control (through `scala.concurrent.context.maxExtraThreads`)?

overviews/core/_posts/2012-09-20-futures.md
backuitist: This section is named `The Global Execution Context`. IMO you ought to know how your `ExecutionContext.global` behaves. I would agree with you if the underlying implementation of the global EC wasn't a `ForkJoinPool`.

overviews/core/_posts/2012-09-20-futures.md
backuitist: @phaller 
> the behavior is typically much better-behaved than with, e.g., a JDK Executors.newCachedThreadPool()
Do you mean that blocking calls are usually infrequent so that you're better off with an EC that clears up spare workers as fast as possible?
What about having to perform frequent blocking calls? Wouldn't it be more efficient to manage them in a dedicated EC with warm threads?

overviews/core/_posts/2012-09-20-futures.md
backuitist: That's too much work for me. I'll leave the link but you're of course free to change this :)

overviews/core/_posts/2012-09-20-futures.md
backuitist: I disagree, references of `ExecutionContext.global` are all over that documentation. So you might be right in theory, but in practice people will be using (and abusing) the global EC. Documenting how this works and flagging potential issues isn't low-level details.

backuitist: > I would suggest that you refactor this section, so that all mention of ForkJoinPool is put in a separate section,
entitled e.g. "Execution Contexts based on Fork/Join Pools".

That's actually what I did by wrapping FJP details into the `The Global Execution Context` section.

SethTisue:
@axel22 @viktorklang lgty now?

axel22:
I would fix that one last colon, but it looks good to me.

SethTisue:
yay, across the finish line! thank you @backuitist for your great patience, thank you reviewers. (@viktorklang, it's of course not too late for further changes.)

viktorklang:
@SethTisue :-)

PR #323, opened 2014-04-02
Added sip on automatic argument type inference

MikaelMayer:
Automatic argument type inference would allow a lightweight syntax for scala. Namely, we could write:

```Scala
 object Test {
 	    import pkg.{customName: ClassName}
         // Do something with customName of type ClassName
         def method1(customName) = ???
         
         import pkg.{Class2Name => personnalizedName: CN}
         // def method2(className: ClassName, className2: ClassName, personnalizedName: CN)
         def method2(className, className2, personnalizedName) = ???
         
         // case class Innerclass(customName: ClassName)
         case class InnerClass(customName)
         
         //              (className: ClassName) => className.methodInClassName()
         val anonymous = className => className.methodInClassName()
 
         // Declares an implicit type for variable cname in scope.
         implicit type cname: ClassName
         
         // Use of variable name cname.
         def method3(cname, i: Int) = cname.methodInClassName(i)
     }
```

### DIFF
--- a/sips/pending/_posts/2014-04-02-implicit-argument-types.md
+++ b/sips/pending/_posts/2014-04-02-implicit-argument-types.md
@@ -1,0 +1,127 @@
+---
+layout: sip
+disqus: true
+title: SIP-23 - Implicit argument types
+
+---
+
+**By: Mikael Mayer**
+
+## Abstract ##
+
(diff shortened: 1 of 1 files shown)


SethTisue:
This seems moribund, so I'm provisionally closing.

PR #331, opened 2014-04-16
Update "Parenthesis () vs {}"

LannyRipple:
Provide reasoning, workaround, references.


SethTisue:
Yeah, I don't think this is worded well enough (e.g. "fiddly bits" is too vague) to merge as-is, and needs a bit too much work for me to just fix it and merge, either. Lanny, are you interested in doing another version? If so, feel free to reopen; I'm only closing this on the assumption that it may be abandoned.

PR #344, opened 2014-05-09
Propose clarifying constructor invocation

sksamuel:


SethTisue:
I started a thread on this at https://groups.google.com/d/msg/scala-language/rtkamhSxv-Q/09Y_NnMbMocJ

SethTisue:
Closing for lack of sufficient consensus, in my judgment.

SethTisue:
Nonetheless, it's a little odd that the style guide doesn't touch on the issue at all. Right now the style guide tends to remain silent on such issues. I'd be fine with a revised text that mentioned this — and/or other such points — and says that preferred styles vary.

sksamuel:
I think the 1 year it took to get any interest in this means I don't think there is any, and it could just be left unspecified :)

PR #346, opened 2014-06-27
Add SIP-23: Literal-based singleton types

folone:
Hello folks,

Looks like we've discussed all the details on the mailing lists ([scala-language](https://groups.google.com/forum/#!topic/scala-language/9VEnFZImyJI) and [scala-sips](https://groups.google.com/forum/#!topic/scala-sips/YRHd8WW0V40)), and there seem to be a general agreement on what this feature has to look like. So I decided to be bold, and go ahead with porting the [draft document](https://docs.google.com/document/d/1V5Xc3Up_hTuhFNtsZMYJwvCG0IHviq0UMWTjWANwvZ4/edit#) to markdown, incorporating all the suggestions from mailing lists as well as comments from the document itself.

Just read through the [SIP Submission Process document](http://docs.scala-lang.org/sips/sip-submission.html), and it looks like I've messed up some intermediate steps (for example I've posted to scala-language instead of scala-debates). Hope this is okay. Also not sure if I can have number 42 for this SIP. Hope this is okay too :)

Please let me know what you think.

/cc @xeno-by @retronym @adriaanm @paulp @odersky @milessabin @gkossakowski @heathermiller

### DIFF
--- a/sips/pending/_posts/2014-06-27-42.type.md
+++ b/sips/pending/_posts/2014-06-27-42.type.md
@@ -6,6 +6,13 @@ title: SIP-23 - Literal-based singleton types
 
 **By: George Leontiev, Eugene Burmako, Jason Zaugg, Adriaan Moors, Paul Phillips**
 
+<span class="label warning">Note: This SIP is considered a "good idea", but is a work-in-process.</span>
+<span class="label warning">There is no guarantee that it will not be dropped if at some point it is decided that its foundation is not good enough.</span>
+
+<span class="label success">Champion: Adriaan Moors</span>
+
+<span class="label success">Last update: July 9, 2014</span>
+
 ## Motivation
 
 Singleton types bridge the gap between the value level and the type level and hence allow the exploration in Scala of techniques which would typically only be available in languages with support for full-spectrum dependent types.


adriaanm:
Hi, we discussed this SIP in the SIP meeting last week. I have a few questions & suggestions to get this going again:

# feature interactions
The sip is great with examples and motivation, but needs a bit more work in the feature interactions area (always the hardest part, but really important!)

  - do we need the lowering macro in the stdlib? if so, can it be integrated with, say SingletonTypeTag somehow?
  - since we're adding a type with new meaning, we should specify its behavior with respect to asInstanceOf, isInstanceOf and type patterns. Since the first two are specified in terms of the pattern, let's start there.  Currently, `x asInstanceOf y.type` does not mean the same thing as `x: y.type` in a pattern, since the `asInstanceOf` does not imply `x eq y`. Existing code will break if we enforce this.
  - Also note that literal paths in singleton types do not have an `eq` method, so we're using `==`, which is fine, but needs to be specified. Maybe some day we can factor out the top of the hierarchy, as dotty is doing -- would be good to ping them about this.
  - the logic behind widening types inferred for val signatures is crucial to make this SIP convenient/useful -- lets add a description to the specification how this works
  - quasiquotes need to be updated

# misc
  - the sip description currently refers throughout to the implementation. I think it would be better to first focus on the concept and then discuss implementation in a separate appendix (it's too tempting to drive the spec from the implementation needs otherwise)

DarkDimius:
> since we're adding a type with new meaning, we should specify its behavior with respect to asInstanceOf, isInstanceOf and type patterns. Since the first two are specified in terms of the pattern, let's start there. Currently, x asInstanceOf y.type does not mean the same thing as x: y.type in a pattern, since the asInstanceOf does not imply x eq y. Existing code will break if we enforce this.

Also, note that `this.type` already exists and casting to it is used in some code patterns.

milessabin:
+ Several years of experience with shapeless and shapeless-based libraries suggests that yes we do need the ability to drop from the type to the value level. Adding language level support here removes the need for macros elsewhere.
+ I'm open to suggestions, but I don't see how this could be integrated cleanly with `TypeTag`s. We need a value, which will only be available for singleton types whereas `TypeTags` are defined both singleton and non-singleton types.
+ I don't follow what you mean by "we're adding a type with new meaning" ... are you referring specifically to `ValueOf`? `ValueOf` is intended to have as little semantics as possible ... in effect it's just a newtype to allow us to materialize a unique value without polluting the implicit scope for the underlying type.
+ Agreed on describing the narrowing/widening logic. I modified the behaviour from where you left it to make it as small a change from the status quo as possible, but the status quo (esp. wrt the significance of `final`) wasn't particularly well documented, so the more the better.
+ Can you say what needs updating in quasiquotes? I think they should function as intended, and respect the setting of the flag in the running compiler.

paulp:
> I don't follow what you mean by "we're adding a type with new meaning" ... are you referring specifically to ValueOf?

I believe he's referring to LiteralType.

milessabin:
> I believe he's referring to LiteralType.

Ah, right. Both `LiteralType` and `FoldableConstantType` are internal to the implementation. I agree they should be documented as part of the implementation, but I don't think they should be mentioned in the SIP or the SLS.

As types they aren't new ... they're singleton types of particular values and have been represented in Scala's type system, albeit without syntax, for aeons.

paulp:
You can read "we're adding a type with new meaning" as "we're adding a type syntax with previously inexpressible meaning."

PR #351, opened 2014-06-30
Update index.md

kwladyka:
```Error:(23, 23) missing parameter type for expanded function ((x$3) => x$3.$times(2))
(1 to 5).map { val x=_*2; println(x); x };}```
                  ^

I think it should be:
```(1 to 5).map { val x =(_: Int)*2; println(x); x }```


SethTisue:
I think you have correctly identified a mistake in this document, but the fix isn't right, either. This prints `<function1>`, rather than printing a number.

SethTisue:
replaced by #438

PR #352, opened 2014-06-30
Update index.md

kwladyka:
miss brackets


SethTisue:
You've correctly identified a mistake in the document, but with your change, the code no longer matches the explanatory text — it isn't a "trailing underscore" anymore. Instead of changing `_` to `(_)`, I'd suggest inserting a space before the `_`.

SethTisue:
replaced by #438

PR #364, opened 2014-07-21
Add Gemfile.lock

tkawachi:
Gemfile.lock should be committed to force all users to use the exact same version of the dependent libraries.


SethTisue:
I'm a Ruby idiot. Can someone with more Ruby knowledge than me review this?

@tkawachi, sorry there was no feedback for this on so long. Is is still up to date?

tkawachi:
@SethTisue  I've updated Gemfile.lock to the newest.

jessitron:
+1
we want people to be able to duplicate the build with the same dependency versions.

PR #405, opened 2015-01-27
Worded link to views link better

EdgeCaseBerg:
While reading through the documentation I came across this
sentence and found that it didn't fit quite right. "more about 
views in later" is not a complete sentence. This has been 
fixed.


heathermiller:
Thanks – totally right on this one. Much appreciated!

PR #410, opened 2015-02-02
Update links on overviews page to point to the correct scaladoc pages.

dickwall:
Links to Using and Authoring Scaladoc were incorrect after recent tweaks, so updated them. Also changed the order (using before authoring) and the descriptions to be better.

Review by @heathermiller please


heathermiller:
Ah, good catch! Thank you!

PR #412, opened 2015-02-22
updated syntax from "future{ ... }" (deprecated) to "Future{ ... }" in the examples

seralf:
Hi I updated the example in order to adopt the syntax:

    Future{ ... }

in the examples, as suggested by the annotation:

    @deprecated("Use `Future { ... }` instead.", "2.11.0")

on the constructor itself


heathermiller:
Thank you! (Sorry for the heinous delay on this – somehow I didn't get any notifications!) LGTM!

PR #415, opened 2015-03-04
added refreshed scala logo and tweaked CSS

rasummer:
@heathermiller @dickwall Added refreshed Scala logo and tweaked the CSS.


heathermiller:
Thanks Richard!

PR #417, opened 2015-04-11
SIP-??? created pending: Implicits in for comprehensions

haroldl:
I'd like to propose SIP-24: Implicits in for comprehensions.

### DIFF
--- a/sips/pending/_posts/2015-04-10-implicits-in-for-comprehensions.md
+++ b/sips/pending/_posts/2015-04-10-implicits-in-for-comprehensions.md
@@ -135,7 +135,7 @@ generator clause, expanding the left-hand side of the generator into a tuple.
                        .withFilter((a0, ..., an) => g)) ...
 
 In the above example, either `p1`, `p2` or both may be marked as implicit. If
-`p1` is implicit, then it should be scoped such that it is available in `e1`,
+`p1` is implicit, then it should be scoped such that it is available in
 `e2`, `p2`, `g` and in the body of the for loop/comprehension. If `p2` is
 implicit, then it should be scoped such that it is available in `g` and in the
 body.
--- a/sips/pending/_posts/2015-04-10-implicits-in-for-comprehensions.md
+++ b/sips/pending/_posts/2015-04-10-implicits-in-for-comprehensions.md
@@ -31,7 +31,9 @@ heavy use of for comprehensions (in the `do` syntax style of Haskell monads).
 In those cases, the alternatives are ugly: break the code into several for
 comprehensions or explicitly pass the implicit parameter to other methods.
 
-We would like to write the following (where f3 and f4 have an implicit param),
+We would like to write the following (where f4 and f5 have an implicit param),
+and maybe there is code in the body of the `yield` that also would use
+the implicit value v3:
 
     def f4(implicit v: T) = ...
     def f5(implicit v: T) = ...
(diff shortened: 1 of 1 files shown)


haroldl:
Thank you both for taking a look at this!

I'm not sure how this is making `Option` special. I'm generating AST that maps to code that wraps a value in Some, or else returns None, and then flattening it away so the `Option[T]` values never reach the outside code.

I'm not actually using flatten, I'm translating it to ".flatMap { x => x }" where x is fresh in order to avoid adding any requirements on the source data types for generators. Today they are only required to provide `flatMap`, `map` and `withFilter` and I didn't want to change that.

My next steps are to read the quasiquotes and comprehensive comprehensions docs mentioned by @retronym and figure out how that impacts my work and try making the changes suggested by @adriaanm. This may take me a few days.

I've also updated the code:

I've squashed my commits and moved them to a new branch sip24a based on 2.12.x to be more modern (and added some more tests).

I've changed the `nme.Some` / `nme.None` to use `definitions.SomeModule` / `definitions.NoneModule` to avoid picking up shadowing definitions of `Some` / `None`.

retronym:
> I'm generating AST that maps to code that wraps a value in Some, or else returns None, and then flattening it away so the Option[T] values never reach the outside code.

This depends on the `flatMap` method supporting return values of type `Option[_]`. That holds for the collections (given the impicit `Option.optionToIterable`), but doesn't generalize to other monadic types:

```scala
trait M[A] {
  def map[B](a: A => B): M[B]
  def flatMap[B](a: A => M[B]): M[B]
  def filter(f: A => Boolean): M[A]
}

object Test {
  def guard = true
  def testOk  (m: M[String]) = m.filter(_ => guard).map(x => x)
  def testFail(m: M[String]) = m.flatMap(x => if (guard) Some(x) else None)
}
```

```
qscalac sandbox/test.scala
sandbox/test.scala:10: error: type mismatch;
 found   : Some[String]
 required: M[?]
  def testFail(m: M[String]) = m.flatMap(x => if (guard) Some(x) else None)
                                                             ^
sandbox/test.scala:10: error: type mismatch;
 found   : None.type
 required: M[?]
  def testFail(m: M[String]) = m.flatMap(x => if (guard) Some(x) else None)
                                                                      ^
two errors found
```

The current encoding using `filter`/`withFilter` only requires a common understanding of `Boolean`.

haroldl:
If people are okay with broadening the scope per the suggestion from @densh then the desugaring rules become much simpler and the `Option` + `flatMap(x => x)` piece just goes away.

Otherwise, I'll look for alternatives that do not depend on `Option` though there may need to be some other compromise such that the carrier type needs to implement something else in addition to what is currently required (which would be some method/feature of the existing collections) in order to support implicits in for comprehensions.

retronym:
+1 to the generalization

dickwall:
Have emailed and tweeted Harold to glean continuing interest in pursuing this. Will likely move to dormant in the next SIP/SLIP committee meeting if nothing heard back.

haroldl:
I've closed this PR since I haven't found time to pursue the generalization suggested by @retronym and I think that that is the right way to proceed.

Jacke:
FYI nowadays you can get the same functionality by using https://github.com/oleg-py/better-monadic-for#define-implicits-in-for-comprehensions-or-matches

PR #419, opened 2015-04-24
Fix an example that doesn't compile

lloydmeta:
The example on this page was wrong according to https://issues.scala-lang.org/browse/SI-9280

### DIFF
--- a/tutorials/tour/default-parameter-values.md
+++ b/tutorials/tour/default-parameter-values.md
@@ -16,8 +16,8 @@ In Java, one tends to see a lot of overloaded methods that only serve to provide
 
     public class HashMap<K,V> {
       public HashMap(Map<? extends K,? extends V> m);
-      /** Create a new HashMap with default capacity (16) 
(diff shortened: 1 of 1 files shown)


gourlaysama:
Nice catch. I think it might be more useful here to use a float literal rather than changing the type to `Double`:
```scala
class HashMap[K,V](initialCapacity: Int = 16, loadFactor: Float = 0.75f)
```

lloydmeta:
Done :+1:

gourlaysama:
Ah, I just realized there are two more float literals that need to be added for the example to compile (line 60 and 64, `0.8f`).

lloydmeta:
Nice catch fixed 60 and 64 as well.

gourlaysama:
LGTM, thanks a lot! :sparkles:

/cc @heathermiller for merging :)

heathermiller:
Sorry for the heinous delay on this! (not receiving notifications for some reason :/ )

heathermiller:
LGTM!

PR #423, opened 2015-05-15
Update 2012-11-03-value-classes.md

briangordon:
The "implicit" keyword needs to be present in order to allow the expression `3.toHexString` as promised.


retronym:
Thanks!

PR #424, opened 2015-06-03
Fixes to architecture-of-scala-collections

vsalvis:
* Fixed typos
* Fixed all links (anchors have dashes instead of underscores)
* TraversableLike is a trait, not a class
* Reordered some sections so that titles make more sense
* Restructured outline (level #### is currently displayed with both a dot and a square, so removed it)
* Added a few clarifications


heathermiller:
LGTM! Thanks Vera!

PR #425, opened 2015-06-04
Update deprecated constructs in code example newTermName -> TermName newTypeName -> TypeName

at15:
`newTermName` and `newTypeName` are deprecated, and the code examples in reflection and macros are still using them, so I did a search and replace in editor.  And the Japanese version is updated as well.


heathermiller:
Thanks a lot! LGTM!

PR #427, opened 2015-06-06
Improve English in Methods with Numerous Arguments section

dcbriccetti:
“put all arguments on a line by themselves” is ambiguous. It could mean that all arguments go on a single line, by themselves, apart from what may be on other lines.


heathermiller:
LGTM thanks!

PR #428, opened 2015-06-18
Sip/trait parameters

odersky:
New SIP: Trait Parameters


heathermiller:
Would you like to allow comments on these 2 SIPs?

odersky:
@heathermiller Yes, of course. We want to review them.

heathermiller:
Ok, sure thing. I'll pull this in now and I'll add comments in the front matter in a follow-up commit (`discus: true` is the toggle switch for comments)

adriaanm:
We agreed in the meeting last week that this proposal is desirable, but it would be useful to discuss a few more details:

  - interaction with context bounds encoding (and enumerate other potential interactions)
  - scoping (e.g., is a trait parameter in scope for the self type?)
  - binary compatibility / api evolution guidelines: adding a trait parameter has an impact on all subclasses that's slightly different from adding an abstract member because of the restrictions in passing arguments around, so it would be good to discuss this common scenario
  - are all modifier combinations legal?

PR #431, opened 2015-07-06
Update control-structures.md

propensive:
Reflect current thinking on non-usage of curly braces in case clauses.

### DIFF
--- a/style/control-structures.md
+++ b/style/control-structures.md
@@ -35,9 +35,7 @@ Remember the following guidelines:
 -   `for` - Omit braces if you have a `yield` clause. Otherwise,
     surround the contents with curly-braces, even if the contents are
     only a single line.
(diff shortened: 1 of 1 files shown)


dickwall:
Well, I agree, LGTM, however will give this a day or so to see if there are any dissenting opinions (pardon me, is this Contradiction? Oh, I was looking for Arguments, they're down the hall on the left...)

SethTisue:
Pretty sure there was a mailing list thread on this not long ago and the mob was unanimously in favor of this change. I have never seen a codebase that used the old rule.

dickwall:
Right on - merging it in then.

SethTisue:
for the record, here's the thread: https://groups.google.com/d/msg/scala-debate/ZH2XiimsC0k/Jghw8pKUCl4J

PR #433, opened 2015-07-07
Jettison slip

dickwall:
Since SLIPs now have a home on github, this PR reverts much of the stuff previously added to include SLIPs in the contrib pages. They are still mentioned, but only to redirect people to the correct resources in github and elsewhere. Some changes to the SIP side (including adding a sip-template to replace the slip-template) are included. Mostly it's removal of stuff though and should be a quick review.

Please review by @heathermiller and/or @SethTisue


SethTisue:
LGTM

PR #435, opened 2015-07-15
Moved to standard notation in performance characteristics docs

SamuelMarks:
(replaces old PR => https://github.com/scala/scala.github.com/pull/265)

> Notation used is now the standard computer-science one; a.k.a.: "Big-O" Notation from the Bachmann-Landau family.

> Also allowed me to remove the legend table; as the notation is standard. Additionally added in `<sup>` for the more specific amoritized complexities and the like. (often represented with * notation, but this'll suffice)

### DIFF
--- a/overviews/collections/performance-characteristics.md
+++ b/overviews/collections/performance-characteristics.md
@@ -13,73 +13,68 @@ The previous explanations have made it clear that different collection types hav
 
 Performance characteristics of sequence types:
 
-|               | head | tail | apply | update| prepend | append | insert |
-| --------      | ---- | ---- | ----  | ----  | ----    | ----   | ----   |
-| **immutable** |      |      |       |       |         |        |        |
-| `List`        | C    | C    | L     | L     |  C      | L      |  -     |
-| `Stream`      | C    | C    | L     | L     |  C      | L      |  -     |
-| `Vector`      | eC   | eC   | eC    | eC    |  eC     | eC     |  -     |
-| `Stack`       | C    | C    | L     | L     |  C      | C      |  L     |
-| `Queue`       | aC   | aC   | L     | L     |  L      | C      |  -     |
-| `Range`       | C    | C    | C     | -     |  -      | -      |  -     |
-| `String`      | C    | L    | C     | L     |  L      | L      |  -     |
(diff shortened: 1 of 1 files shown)


overviews/collections/performance-characteristics.md
```diff
@@ -13,73 +13,68 @@ The previous explanations have made it clear that different collection types hav
...
-| **aC**  | The operation takes amortized constant time. Some invocations of the operation might take longer, but if many operations are performed on average only constant time per operation is taken. |
-| **Log** | The operation takes time proportional to the logarithm of the collection size. |
-| **L**   | The operation is linear, that is it takes time proportional to the collection size. |
-| **-**   | The operation is not supported. |
+|                  |                                                                                                                                                           |
+| -----            | ----                                                                                                                                                      |
+| **<sup>†</sup>** | Assuming bits are densely packed.                                                                                                                         |
+| **\***           | The operation takes effectively constant time, but this might depend on some assumptions such as maximum length of a vector or distribution of hash keys. |
```
dickwall: Seems to render as \* instead of *, at least in the preview

dickwall:
Personally I understand the urge to move to big O, but I don't think the result is easier to read or absorb. We teach these characteristics often in training, and the old "non-standard" form is very easy for people to digest. I want to see what others think before integrating this, my own leaning is to leave it alone.

TimothyKlim:
>Notation used is now the standard computer-science one; a.k.a.: "Big-O" Notation from the Bachmann-Landau family.
+1

SethTisue:
@dickwall started some discussion on this at https://gitter.im/scala/scala?at=55a9553f854106d36bd1afe2 . there doesn't seem to be a consensus in favor of the change, so I suggest we close this. (sorry @SamuelMarks — we do appreciate your well-intentioned work)

dickwall:
Yes, but please do keep 'em coming. Volunteer effort is always welcome, even if it is not always integrated. Maybe before a potential future change, drop in at scala/scala on gitter and just let people know what you are thinking of doing, in this case it became apparent that this is a trickier subject - O(1) for Vector for example is not correct, but then neither is O(Log2 N) since the maximum number of descents is 7, making it effectively constant

PR #436, opened 2015-07-18
reorder Tour of Scala sections more logically

SethTisue:
thanks to Stephen Compall for suggesting this; his original
pull request (#307) couldn't be merged because too much time
had passed, but I have followed his suggested ordering here

I manually verified that all of the "next" and "previous" links are
correct.


SethTisue:
review by @heathermiller

tutorials/tour/abstract-types.md
```diff
@@ -5,11 +5,10 @@ title: Abstract Types
 disqus: true
 
 tutorial: scala-tour
-num: 2
-outof: 35
```
heathermiller: Why'd you remove the `outof` here? I see there's a second one (below) which doesn't agree with this one.

SethTisue: There are only 33 pieces to the tutorial, not 35. I don't even really know what effect `outof` has but I figured it should go in the first section.

heathermiller: Haha funny that we have 33 not 35. No comment there (I must be responsible). `outof` is used for generating indicies and TOCs... Liquid markup is not the smartest templating engine – last I checked you couldn't populate a data structure in it with a subset of pages you want, so we have to do a very dumb "x out of y" articles kind of thing to know when to stop looping through all pages in the site. Yeah. I know.

heathermiller: I was just putting the `outof` in the first file alphabetically in the folder since, based on the filename, it wasn't clear which number which article got.

heathermiller: Believe it or not this is actually somehow a little robust. It at least means no broken links...

heathermiller:
So I think we're good to merge this. I'd just like to merge and then revert the commit with the previous/next buttons and then do a follow up commit removing these bits from the YAML frontmatter afterwards. I don't think that the benefits provided by the next/previous buttons is worth the complication. I think the TOC is far more often used to navigate between tour articles anyway.

PR #437, opened 2015-07-21
Clarifying where ActorSystems should reside.

vjovanov:
Users requested this clarification in:
  https://github.com/scala/actors-migration/issues/9


heathermiller:
Thanks, LGTM :)

PR #441, opened 2015-07-27
archive old GitHub wiki in commit history

SethTisue:
one commit adds it, second commit removes it again


SethTisue:
(still-relevant content was merged into the scala/scala README in https://github.com/scala/scala/pull/4636)

PR #445, opened 2015-08-13
Move "Common methods" before "Automatically imported methods"

alexeyr:
This moves the more common case earlier, and looking up documentation is discussed
before what to do when it isn't found. Also makes the Scaladoc link more visible, and
mentions third-party libraries explicitly.


SethTisue:
LGTM

alexeyr:
Please don't merge yet, I want to add another commit.

alexeyr:
@SethTisue Done.

SethTisue:
thanks Alexey for these nice improvements

PR #447, opened 2015-08-21
remove references to CLR/.NET

SethTisue:
review by ???, perhaps @dickwall?

I used `git grep CLR` and `git grep NET` to find the places needing attention.


dickwall:
LGTM although I have only the vaguest idea of what the spanish versions are saying, and the kanji ones I have no clue (maybe they are saying "I expect you English speakers think this sentence is about .NET but in fact it's very technically about Java instead") :-). Anyway - I say ship it and let people write in if the translations make no sense any more.

Dick

PR #449, opened 2015-08-24
Replace flags with endonyms to link to localization

eed3si9n:
Fixes scala/scala.github.com#448. 

This change replaces series of `a img` elements with `ul.langbar li a`. The localized name for each language is kept in a common YAML file called `_data/languages.yml`. This file could be used to keep more localized text down the line.

I tried to make the language bar blend into the existing design of the pages. Here are some samplings.

### Cheatsheet
scalacheat_-_scala_documentation

### Guides and Overviews

guides_and_overviews_-_scala_documentation

Overview pages

string-interpolation


heathermiller:
LGTM :)

PR #450, opened 2015-08-24
Adds Chinese translation (Subsumes #280)

eed3si9n:
Subsumes #280 by fixing file naming and localization mechanics.

The credit for the actual translation work goes to the following 21 individuals listed on https://code.csdn.net/DOC_Scala/chinese_scala_offical_document

- 陈骏	jacty0219	Jacty
- 陈幸	Meteor2520	meteor2520
- 董泉	dqsweet	dqsweet
- 何乃梧	yuyi20112011	yuyi20112011
- 黄越勇	aptweasel	brucehuang
- 赖正兴	laizx	laizx
- 李奕飞	fancylee	SarahCla
- 林君	a455642158	玲渱胾
- 刘国锋	iceongrass	iceongrass
- 吕浩志	lvhaozhi	smile_lhz
- 聂雪珲	blueforgetmenot	blueforgetmenot
- 潘栋华		pandonghua
- 潘义文	Caidaoqq	Paner
- 王金岩	i9901028	VV一笑
- 王雨施		m00nlight
- 熊杰	xiaoxiong345064855	jjxiong
- 杨志斌	qwewegfd	qwewegfd
- 张冰	usen521	Flicy
- 张明明	a775901421	zhangmingming
- 张欣	kevenking@gmail.com	kevenking
- 周逸灵	pastgift	

According to Google Translate the page says:

> From October 2013, CSDN CODE began organizing volunteers to translate official documents Scala. > Plan translated documents Scala main content overview section of the official website, includes the following sections:  ....
>
> After an open call screening, we finally organized a dozen volunteers to carry out this translation work. We invite to the well-known domestic Scala community Scala研学社 ("Scala research society") two teachers 连城 and 尹绪森 as a consultant and translation proofreading work. To express my heartfelt thanks to Scala Society Institute!
>
> Special thanks also to the hard work is paying in the translation work, the majority of the volunteer interpreters friends, they are:
> (Hereinafter sorted by last name spelling)

## screenshots

index_-_scala_documentation

collections_-_intro_-_scala_documentation

PR #451, opened 2015-08-24
Fixes file names of Chinese translations.

eed3si9n:
Currently the first link on http://docs.scala-lang.org/zh-cn/overviews/index.html, which is http://docs.scala-lang.org/zh-cn/overviews/collections/introduction.html, is broken. This PR fixes that. Ref #450

### note

When I renamed them using Mac last night the file system didn't catch some of the renames that only differs in case. You could `git mv` first to some bogus file name or use a Linux box.


SethTisue:
thanks!

PR #460, opened 2015-10-12
fix: add a new line before the bullet points

ThibaudDauce:
I think this change will fix http://docs.scala-lang.org/tutorials/tour/implicit-parameters.html


SethTisue:
good catch. thanks

PR #461, opened 2015-10-12
fix: fix typo in view example

ThibaudDauce:
a should be A (it's a type)
try to fix the bullet points too with a new line.


SethTisue:
great, thank you!

PR #462, opened 2015-10-14
Update TraitsTest sample to extend App instead of Application

esamson:
`scala.Application` has been deprecated since version 2.9.0 and is no
longer available as of 2.11.0.

### DIFF
--- a/ko/tutorials/tour/traits.md
+++ b/ko/tutorials/tour/traits.md
@@ -26,7 +26,7 @@ language: ko
         obj.isInstanceOf[Point] &&
         obj.asInstanceOf[Point].x == x
     }
-    object TraitsTest extends Application {
+    object TraitsTest extends App {
       val p1 = new Point(2, 3)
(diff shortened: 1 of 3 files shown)


ko/tutorials/tour/traits.md
```diff
@@ -41,4 +41,4 @@ language: ko
     true
     true
 
-윤창석, 이한욱 옮김
\ No newline at end of file
+윤창석, 이한욱 옮김
```
SethTisue: how come this is showing up in the diffs?

esamson: Previous version did not have a newline character at the end.
I blame **vim** for automatically adding one.

SethTisue: Ah... normally GitHub indicates that visually a specific way. I guess the Unicode confused it.  No worries, then.

SethTisue:
good catch. thank you!

PR #466, opened 2015-11-06
Scala Tour: call them "implicit conversions" not "views"

SethTisue:
the old terminology is a blast from the distant past

review by @heathermiller


soc:
It LGMT. :-)

SethTisue:
@heathermiller the only reason I'm hesitating to just hit "merge" on this myself is that I just removed the Spanish translation of the page instead of updating it, since I don't know any Spanish.

heathermiller:
All good. Thanks :) LGTM

PR #468, opened 2015-11-12
front page tweaks

SethTisue:
mostly in the area of removing "under construction" signs
that have been there for several years.  it is what it is.

review by @heathermiller


heathermiller:
Thank you! LGTM

PR #470, opened 2015-11-25
Differ between PascalCase and camelCase

paulclarkin:
"camelCase" with a Capital First Letter is known as PascalCase (c2.com/cgi/wiki?PascalCase).

### DIFF
--- a/style/naming-conventions.md
+++ b/style/naming-conventions.md
@@ -14,7 +14,7 @@ few important exceptions to this guideline (as given below).
 
 ## Classes/Traits
 
-Classes should be named in the camelCase style with the very first
+Classes should be named in the PascalCase style - camelCase with the very first
 letter of the name capitalized:
 
     class MyFairLady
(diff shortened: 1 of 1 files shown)


SethTisue:
not convinced this is a good change. I don't recall ever hearing or seeing the term "PascalCase" before. Wikipedia only mentions it in passing and says its used "in Microsoft documentation". I see and hear "camel case" used to refer interchangeably to both `FooBar` and `fooBar` style naming. and Pascal itself is a historical relic these days so it doesn't seem like a great term to me.

paulclarkin:
Maybe it's my .NET background but its a familiar term to me. It is used in Microsoft documentation because one of the Pascal designers worked on earlier .NET versions (http://blogs.msdn.com/b/brada/archive/2004/02/03/67024.aspx) and it removes the ambiguity between `CamelCase` and `camelCase`.

I prefer to remove the ambiguity, but that's up to you.

heathermiller:
I second Seth. I've never seen this referred to as PascalCase before.

dwijnand:
FWIW it's the term I also use to be less ambiguous than "camel case" and more terse than "upper-case camel case". But I guess it's not ubiquitous enough.

It's a shame there's no strong candidate substitute for "lower-case camel case".

If anything this pull request shows there's some difference in how camel case is referred to in the docs.

SethTisue:
I took a new stab at at least regularizing the existing usage in #474. @paulclarkin, sorry we didn't take the change, but we appreciate your attention to this.

PR #475, opened 2015-12-18
Code of Conduct is on main Scala site now

SethTisue:
it was added there by https://github.com/scala/scala-lang/pull/382

in this repo, there were no links to update.

the exact redirection method used here is based on advice I found at
http://stackoverflow.com/a/506004/86485

I also looked at how the redirect was done for overview/macros.html ,
but the use of `if` there
1) prevents it from working when testing the site locally
2) doesn't seem necessary afaict

review by @heathermiller


heathermiller:
LGTM!

PR #476, opened 2015-12-18
SI-7900 Seed doc for REPL

som-snytt:
A placeholder to list features of common interest.


hamishdickson:
Big :+1: on this @som-snytt 

I was hoping we could add a bit in about simple navigation/history editing etc and loading things into the repl. I'll create a gist or something with what I was thinking to see if you think it's worth adding

heathermiller:
Thanks @som-snytt, this is certainly useful!! LGTM

PR #477, opened 2016-01-04
Java 8 can have "traits" too, the way they are defined here

lukaseder:


SethTisue:
it would be possible to expand further on how Java 8's default methods and methods in Scala traits are similar but have important differences as well. regardless, the change as written seems to me like strictly an improvement, so I suggesting merging; WDYT @heathermiller?

lukaseder:
I agree that this section deserves more detail, specifically because the Java 8 EG members have repeatedly expressed their disapproval of comparing Java 8 interfaces with Scala's traits (although I personally fail to see the significant de-facto difference).

From my side, I just wanted to point out to the fact that the current text is incorrect.

heathermiller:
LGTM

SethTisue:
thank you Lukas!

lukaseder:
Sure, no worries

PR #479, opened 2016-01-06
fix the 404 issue

zhouhualei:
fix the incorrect link in index page


heathermiller:
Hi, thanks for your contribution! However, I don't see any 404 issue. When I go to the Chinese overviews index, all links work for me. Further, links should always go to html documents, not markdown documents (Jekyll converts md to html, there are no md documents left after Jekyll has been run), so this can't be correct. So as far as I can tell, I think this PR must be making some kind of error...

zhouhualei:
ok, got it.

Issue #480, opened 2016-01-07
Example exhibits integer overflow/incorrect result silently


sfaleron:
In https://github.com/scala/scala.github.com/blob/master/overviews/parallel-collections/overview.md, under "Some Examples", the fold/sum example overflows.

One might choose a smaller range, or point out the gotcha for the unwary.

    scala> (1 to 1e6.toInt).map(_.toLong).sum
    res42: Long = 500000500000

SethTisue:
nice catch! can I interest you in submitting a pull request with a fix?

sfaleron:
I'll give it a go.  I'm not familiar with pull requests, but I need to learn :)

SethTisue:
cool. help is available at https://gitter.im/scala/contributors

sfaleron:
Anything further I need to do? Did I miss anything?

SethTisue:
merged!

PR #482, opened 2016-01-07
Shrank range in fold/sum example to correct overflow error.

sfaleron:
Proposed fix for issue #480


SethTisue:
good catch, thank you for the contribution

sfaleron:
no problem, it's nice to have the opportunity to give back to open source!

heathermiller:
:+1: Thank you @sfaleron!!

PR #484, opened 2016-01-11
Fix extension methods example in the ja documentation

harry0000:
I fixed an issue that did not exist a `implicit` keyword in the "Extension methods" example of the ja documentation.

### DIFF
--- a/ja/overviews/core/value-classes.md
+++ b/ja/overviews/core/value-classes.md
@@ -47,7 +47,7 @@ title: 値クラスと汎用トレイト
 
 `RichInt` から抜粋した以下のコードは、それが `Int` を拡張して `3.toHexString` という式が書けるようにしていることを示す:
 
-    class RichInt(val self: Int) extends AnyVal {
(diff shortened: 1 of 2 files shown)


SethTisue:
good catch! would you mind fixing `zh-cn/overviews/core/value-classes.md` as well?

harry0000:
I Fixed zh-cn documentation too. It seems to have been corrected all about this problem.

However zh-cn documentation has more probrems with the markdown syntax. There is a problem that the text is displayed as a code snippet because of that there is no line break after syntax highlighting.

[Examples of Limitations #examples-of-limitations](http://docs.scala-lang.org/overviews/core/value-classes.html#examples-of-limitations)

[Examples of Limitations(zh-cn) #section-8](http://docs.scala-lang.org/zh-cn/overviews/core/value-classes.html#section-8)

Later I will report this problem as issues.

SethTisue:
thanks Harry, the fixes are much appreciated.

PR #491, opened 2016-01-18
Add @static SIP

DarkDimius:
Add @static SIP

### DIFF
--- a/sips/pending/_posts/2016-01-11-static-members.md
+++ b/sips/pending/_posts/2016-01-11-static-members.md
@@ -4,9 +4,9 @@ title: SIP 25 - @static fields and methods in Scala objects(SI-4581)
 disqus: true
 ---
 
-__Dmitry Petrashko and Sébastien Doeraene__
+__Dmitry Petrashko, Sébastien Doeraene and Martin Odersky__
 
-__first submitted TODO 2016__
+__first submitted 11 January 2016__
 
 ## Motivation ##
 
@@ -18,18 +18,19 @@ Some JVM and JavaScript frameworks require classes to have specific static field
 
 For example, classes extending `android.os.Parcelable` are required to have a static field named `CREATOR` of type `android.os.Parcelable$Creator`.
 
-Another example is using an [`AtomicReferenceFieldUpdater`](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.html).
+Another example is using an [`AtomicReferenceFieldUpdater`](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.html).
 
 On the JavaScript side, one example is [Relay Route Definitions](https://facebook.github.io/relay/docs/guides-routes.html), whose subclasses must define static fields such as `queries`.
 Static methods and fields for JavaScript classes are one of the very few things (if not the only thing) that Scala.js "cannot do" at the moment, at least not declaratively.
 
 ## Overview ##
 
-In order for method or field to be considered static it needs to be defined in an `object` and annotated `@static`.
+In order for a method or field to be considered static it needs to be defined in an `object` and annotated `@static`.
 There is no special syntax proposed to access these members, they are accessed as if they were a member of defining objects with all appropriate access requirements for accessing them.
 
 For example:
 
+```scala
 {% highlight scala %}
 class Foo
 
@@ -41,10 +42,12 @@ object Foo {
 println(Foo.x)
 println(Foo.bar(12))
 {% endhighlight %}
+```
 
-Intuively, the presence of the `@static` annotation ensures that a field/method is declared as a static member of the companion class.
+Intuitively, the presence of the `@static` annotation ensures that a field/method is declared as a static member of the companion class.
 For the JVM, the above would therefore look to other Java code as if it had been declared with the following Java code:
 
+```java
 {% highlight java %}
 class Foo {
   public static int x = 5;
@@ -53,10 +56,12 @@ class Foo {
   }
 }
 {% endhighlight %}
+```
 
 In Scala.js, the `@static` annotation has no semantic effect in Scala objects, as they are not visible from JavaScript anyway (it could be used for optimizations).
 It has a semantic effect on Scala.js-defined JS classes, for example:
 
+```scala
 {% highlight scala %}
 @ScalaJSDefined
 class Foo extends js.Object
@@ -67,9 +72,11 @@ object Foo extends js.Object {
   @static def bar(y: Int): Int = x + y
 }
 {% endhighlight %}
+```
 
 would look to JavaScript code as if it had been declared with the following JavaScript code:
 
+```javascript
 {% highlight javascript %}
 class Foo extends Object {
   static bar(y) {
@@ -78,22 +85,42 @@ class Foo extends Object {
 }
 Foo.x = 5; // in ES6, there is no declarative syntax for static fields yet
 {% endhighlight %}
+```
+
+## Comparison with mirror classes ##
+
+Scalac currently generates static forwarders for fields and methods in top-level objects:
+
+```scala
+{% highlight scala %}
+object O {
+  val d = 1 
+  object I {
+    val f = 1
+  }
+}
+{% endhighlight %}
+```
+
+Under the proposed scheme users will be able to opt-in to have the field `f` defined in the inner object `I` emited as a static field. 
+In case `O.d` is annotated with `@static` the field will be crated as a static field `d` in `class O`.
+If not annotated, it will be created in the companion module with a static forwarder `d` in `class O`.
 
 ## Restrictions ##
 
-The following rules ensure that method can be correctly compiled into static member on both JVM and JavaScript:
+The following rules ensure that methods can be correctly compiled into static members on both JVM and JavaScript:
 
 1. Only objects can have members annotated with `@static`
 
 2. The fields annotated with `@static` should preceed any non-`@static` fields. This ensures that we do not introduce surprises for users in initialization order.
 
-3. The right hand side of a method or field annotated with `@static` can only refer to members of globally accessible objects and `@static` members. In particular, for non-static objects `this` is not accesible. `super` is never accessible.
+3. The right hand side of a method or field annotated with `@static` can only refer to top-level classes, members of globally accessible objects and `@static` members. In particular, for non-static objects `this` is not accesible. `super` is never accessible.
 
 4. If a member `foo` of an `object C` is annotated with `@static`, the companion class `C` is not allowed to define term members with name `foo`. 
 
 5. If a member `foo` of an `object C` is annotated with `@static`, the companion class `C` is not allowed to inherit classes that define a term member with name `foo`.
 
-6. Only @static methods and vals are supported in companions of traits. Java8 supports those, but not vars, and JavaScript does not have interfaces at all.
+6. Only `@static` methods and vals are supported in companions of traits. Java8 supports those, but not vars, and JavaScript does not have interfaces at all.
 
 ## Compilation scheme ##
 No modificaiton of the typer is planned. The current proposed scheme piggybacks on already existing scoping restrictions in the typer, thus requiring `@static` methods to be defined in `objects`.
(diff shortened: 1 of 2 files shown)


sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+class Foo extends Object {
+  static bar(y) {
+    return x + y;
+  }
+}
+Foo.x = 5; // in ES6, there is no declarative syntax for static fields yet
+{% endhighlight %}
+```
+
+## Comparisson with mirror classes ##
```
dwijnand: Comparison

DarkDimius:
The implementation was merged into Dotty.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+
+We would like to allow methods and fields to be compiled as static. This is usable for interop with Java and other JVM languages, as well as with JavaScript, and is convenient for optimizations.
+
+## Use Cases
+
+Some JVM and JavaScript frameworks require classes to have specific static fields and/or methods.
+
+For example, classes extending `android.os.Parcelable` are required to have a static field named `CREATOR` of type `android.os.Parcelable$Creator`.
+
+Another example is using an [`AtomicReferenceFieldUpdater`](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.html).
```
janekdb: s/7/8 ?

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+
+On the JavaScript side, one example is [Relay Route Definitions](https://facebook.github.io/relay/docs/guides-routes.html), whose subclasses must define static fields such as `queries`.
+Static methods and fields for JavaScript classes are one of the very few things (if not the only thing) that Scala.js "cannot do" at the moment, at least not declaratively.
+
+## Overview ##
+
+In order for method or field to be considered static it needs to be defined in an `object` and annotated `@static`.
```
janekdb: `for method or field` -> `for a method or field`.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+  @static val x = 5
+  @static def bar(y: Int): Int = x + y
+}
+
+println(Foo.x)
+println(Foo.bar(12))
+{% endhighlight %}
+```
+
+Intuively, the presence of the `@static` annotation ensures that a field/method is declared as a static member of the companion class.
```
janekdb: `Intuitively`

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+{% endhighlight %}
+```
+
+Under the proposed scheme users will be able to opt-in to have the field `f` defined in the inner object `I` emited as a static field. 
+In case `O.d` is annotated with `@static` the field will be crated as a static field in `class O`.
+If not annotated, it will be created in the companion module with a static forwarder in `class O`.
+
+## Restrictions ##
+
+The following rules ensure that method can be correctly compiled into static member on both JVM and JavaScript:
```
janekdb: `... that method can be ... into static member ...` -> `... that methods can be ... into static members ...`

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Compilation scheme ##
+No modificaiton of the typer is planned. The current proposed scheme piggybacks on already existing scoping restrictions in the typer, thus requiring `@static` methods to be defined in `objects`.
+If implemented in the dotty code base, the following modifications would be needed:
+ - extend `RefChecks` to check restrictions 1, 2, 4, 5 and 6. This can be done in a separate mini-phase;
+ - extend `LamdaLift.CollectDependencies` to be aware that accessing a member annotated `@static` should not trigger capturing the object that contains this member;
+ - extend `LambdaLift` to trigger an error if a method annotated with `@static` method cannot be lifted to the top level scope;
+ - extend `GenBCode` to emit static fields and methods in companion classes and forwarders to them in companion modules.
+
+## Overriding&Hiding ##
```
janekdb: `Overriding&Hiding` -> `Overriding & Hiding` ?

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+Java allows classes to define static methods with the same name and signature as a static method of a superclass. In order to define the semantics of such cases, the Java Specification introduces the notion of [hiding](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.2).
+
+This is required because in Java calling a `static` method on a class instance is supported.
+This proposal does not need to introduce this notion as we do not support such calls.
+
+## Comparison with [@lrytz's proposal](https://gist.github.com/lrytz/80f3141de8240f9629da) ##
+Lucas Rytz has proposed a similar SIP, but his SIP requires changes to the typer to ensure that `@static` fields do not capture `this`, as in his proposal `@static` fields are defined in the class, rather than its companion object.
```
janekdb: `Lukas`

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+It also does not address the question of `@static` members in inner objects and inheritance/hiding of those methods in subclasses.
+
+## Open questions ##
+ - @static lazy val
+
+## See Also ##
+ * [SI-4581](https://issues.scala-lang.org/browse/SI-4581) is a request for a `@static` annotation
+ * [Scala.js issue #1902](https://github.com/scala-js/scala-js/issues/1902) is a request for defining static fields in Scala.js-defined JS classes
+ * [Another proposal by @lrytz](https://gist.github.com/lrytz/80f3141de8240f9629da)
```
janekdb: Intentionally repeated?

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+object O {
+  val d = 1 
+  object I {
+    val f = 1
+  }
+}
+{% endhighlight %}
+```
+
+Under the proposed scheme users will be able to opt-in to have the field `f` defined in the inner object `I` emited as a static field.
```
lrytz: maybe mention that you need a companion inner class I.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+  val d = 1 
+  object I {
+    val f = 1
+  }
+}
+{% endhighlight %}
+```
+
+Under the proposed scheme users will be able to opt-in to have the field `f` defined in the inner object `I` emited as a static field. 
+In case `O.d` is annotated with `@static` the field will be crated as a static field in `class O`.
```
lrytz: this is a bit confusing given the phrase above, it sounds like "the field" refers to `I.f`

sips/pending/_posts/2016-01-11-static-members.md
DarkDimius: The current implementation, if companion is missing, is going to generate static field in the `O.I$`.
This is used internally to implement lazy vals since https://github.com/lampepfl/dotty/pull/1226.
I'm not sure if I want to spec this, as this is subject to change.

sips/pending/_posts/2016-01-11-static-members.md
DarkDimius: Yes.

kjsingh:
why can't we have static as a keyword?

sjrd:
Because adding new keywords breaks existing code that used them as identifiers. Adding keywords must be done *very* sparingly, only when there is no other option.

kjsingh:
I guess libs doing interop with Java code will be avoiding that.

jvican:
@DarkDimius is there any change you want to do here before the SIP review?

DarkDimius:
> I guess libs doing interop with Java code will be avoiding that.

Why would they? You may want some data to be seen as Java static field, and you can use `@static` for this.

DarkDimius:
>  @DarkDimius is there any change you want to do here before the SIP review?

@jvican, I've applied those changes now and cleared git history. Feel free to merge if it looks good.

jvican:
Great changes, let's merge it for next week's discussion.

xuwei-k:
https://github.com/scala/scala.github.com/issues/635

adriaanm:
- I don't think it's a good idea to let code gen depend in non-obvious ways on an annotation. Can we break this down further? How about.... 
  - Always emit object members as class statics, as suggested by Martin (solves initialization mismatch that killed scala/scala#894, and keeps static members out of class, which would be problematic in our type system)
  - Deal with name clashes using an `@exportedName` annotation? 
  - Allow emitting a `public static` field without accessor using `@suppressAccessors`?

The proposal should go into more detail on:
  - What about binary compat (adding/removing `@static` annotation)
  - Interaction with initialization semantics?

Some relevant bugs:
 - https://issues.scala-lang.org/browse/SI-6189
 - https://issues.scala-lang.org/browse/SI-6294
 - https://issues.scala-lang.org/browse/SI-6236

retronym:
I'd like to see the `Compilation scheme` fleshed out. Currently, it talks about how this would fit into the compiler pipeline, but lacks details about how things would be compiled to bytecode.

Here's an example I'm interested in:

```scala
package p1;  object C { @static val foo: AnyRef = C; val bar: AnyRef = C }
```

If I translate this according to the intuitive description herein, taking a few guesses at the details, I start with:

```
public class Test {
  public static void main(String[] args) {
    System.out.println(C$.MODULE$.foo()); // null
  }
}

class C {
  public static final Object foo;
  static {
    foo = C$.MODULE$; // initializer moved to CompanionClass.<clinit>
  }
}
final class C$ {
  public static C$ MODULE$;
  private final Object bar;
  public static Object foo() { return C.foo; }
  static {
     Object x = C.foo; // or UNSAFE.ensureClassInitialized(classOf[C])
     new C$();
  }
  private C$() {
    super();
    MODULE$ = this;
    bar = this;
  }
}
```

Running this prints `null`. I guess we should move the call to `C.<init>` after the super constructor call and the assignment of the module var:

```
final class C$ {
  public static C$ MODULE$;
  private final Object bar;
  public static Object foo() { return C.foo; }
  static {
     new C$();
  }
  private C$() {
    super();
    MODULE$ = this;
    Object x = C.foo; // or UNSAFE.ensureClassInitialized(classOf[C])
    bar = this;
  }
}
```

Which avoids the NPE.

But in that case, triggering `C.<clinit>` before `C$.<clinit>` would reverse the order that the initalizer for `foo` and the super constructor of the module class is run.

Both of these variations have another drawback: References to `C` in the RHS of static initializers introduce a cyclic dependency between `C.<clinit>` and `C$.<clinit>`. This is deadlock prone if a a pair of threads simultaneously statically initialize the class and module class.

So maybe we instead should put leave all the initalization code in the module class?

```java
class C {
  public static Object foo; // can't be final
}
final class C$ {
  public static C$ MODULE$;
  private final Object bar;
  public static Object foo() { return C.foo; }
  static {
     new C$();
  }
  private C$() {
    super();
    MODULE$ = this;
    C.foo = this;
    bar = this;
  }
}
```

This looks more promising for thread-safe initialization, and preserving semantics. But we are no unable to declare `foo` as final, which is a bit of a drag, because one reason you want to make a field static is to hold a method handle "constant", and hotspot only is able to fully inline invocations through that handle if the field is marked `final`. (Or, if it has the JDK internal annotation, `java.lang.invoke.Stable`, which isn't in our box of tricks.)

lrytz:
We could also consider omitting the accessor methods for `@static` fields and emitting direct accesses to the field.

DarkDimius:
```java
class C {
  public static Object foo;
  static {
    foo = C$.MODULE$;
  }
}
final class C$ {
  public static C$ MODULE$;
  private final Object bar;
  public static Object foo() { return C.foo; }
  static {
     new C$();
  }
  private C$() {
    super();
    MODULE$ = this;
    Object x = C.foo; // or UNSAFE.ensureClassInitialized(classOf[C])
    bar = this;
  }
}
```

Is the proposed desugaring.

@retronym 
>I guess we should move the call to `C.<init>` after the super constructor call and the assignment of the module var:

I guess you've meant `C.<clinit>`. Than that's true.

> But in that case, triggering `C.<clinit>` before `C$.<clinit>` would reverse the order that the initalizer for foo and the super constructor of the module class is run.

`C$.<clinit>` isn't user defined. Did you mean `C$.<init>`?

In case you did, the order in which initializer for foo and the super constructor of the module class is run indeed may change under the proposed  sip. It's discussed here https://github.com/scala/scala.github.com/pull/658/files#diff-9284321b7900073f358a16a971c06881R159 , do you want me to expand this discussion?

> This is deadlock prone if a a pair of threads simultaneously statically initialize the class and module class.

If we always trigger initialization of `C.<clinit>` from `C$.<init>` we will introduce ordering here that should help. But in case user-code inherently has cycles in inititaliziation, I don't think we can do anything.

Thanks for the questions, I'll expand the "Compilation scheme" section.

retronym:
>  do you want me to expand this discussion?

Yes, it would be useful to talk about the super constructor thing explicitly,
perhaps with an example showing how the order could vary depending on how
whether the class or module was initialized first.

> If we always trigger initialization of C.<clinit> from C$.<init> we will
> introduce ordering here that should help. But in case user-code inherently
> has cycles in inititaliziation, I don't think we can do anything.

I still think that we're introducing a new form of cycle by moving
the code into 'C$'-s static initializer, but I guess that's what the user
is asking for with this annotation, so we can argue that it is the user's
responsibility to code around this possibility.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -0,0 +1,151 @@
...
+{% endhighlight %}
+```
+
+Intuitively, the presence of the `@static` annotation ensures that a field/method is declared as a static member of the companion class.
+For the JVM, the above would therefore look to other Java code as if it had been declared with the following Java code:
+
+```java
+{% highlight java %}
+class Foo {
+  public static int x = 5;
```
kjsingh: shouldn't this be final?

SethTisue: yes

PR #492, opened 2016-01-19
fix typo in case classes documentation

boonedox:


heathermiller:
Thanks a lot for the fix! LGTM :)

PR #493, opened 2016-01-26
Fix markdown syntax of the value-classes.md in the zh-cn translation

harry0000:
Fixes scala/scala.github.com#485


SethTisue:
thank you! very much appreciated

PR #500, opened 2016-02-24
Update links to the Scala Language Specification

smarter:
The PDF spec hasn't been updated since 2.9.


SethTisue:
thanks Guillaume!

PR #501, opened 2016-03-07
add reuse warning about buffered iterators

brahn:
The [scaladocs for Iterators]( http://www.scala-lang.org/api/current/index.html#scala.collection.Iterator) has a very useful warning on the method `buffered`:

> Reuse: After calling this method, one should discard the iterator it was called on,
> and use only the iterator that was returned. Using the old iterator is undefined,
> subject to change, and may result in changes to the new iterator as well.

This PR adds that warning (along with some explanatory text) to the corresponding discussion on http://docs.scala-lang.org/overviews/collections/iterators.html#buffered-iterators

(I got bitten by looking at the latter without seeing the former.)

### DIFF
--- a/overviews/collections/iterators.md
+++ b/overviews/collections/iterators.md
@@ -13,7 +13,7 @@ An iterator is not a collection, but rather a way to access the elements of a co
 
 The most straightforward way to "step through" all the elements returned by an iterator `it` uses a while-loop:
 
-    while (it.hasNext) 
+    while (it.hasNext)
       println(it.next())
 
 Iterators in Scala also provide analogues of most of the methods that you find in the `Traversable`, `Iterable` and `Seq` classes. For instance, they provide a `foreach` method which executes a given procedure on each element returned by an iterator. Using `foreach`, the loop above could be abbreviated to:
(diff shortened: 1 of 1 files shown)


SethTisue:
a great many methods on `Iterator` have this issue; `buffered` isn't really special in this regard. what do you think about taking what you've done here and making the warning more general...?

overviews/collections/iterators.md
```diff
@@ -174,3 +174,24 @@ Every iterator can be converted to a buffered iterator by calling its `buffered`
...
+For example, if you wish to pass an iterator and its first element to a function, call the function as follows:
+
+    // Correct
+    val it: Iterator[Int] = ...
+    val bit = it.buffered
+    myFunction(bit.head, bit)
+
+    // Incorrect
+    val it: Iterator[Int] = ...
+    myFunction(it.buffered.head, it)
```
som-snytt: in #551 I just reiterate, "don't re-use the original iterator", which is the theme of most of the page.

I also place an ad for `Iterator.single(head) ++ rest`. It makes me wonder if `BufferedIterator.toIterator` should do that; except that it doesn't have much overhead, which is another point I added.

SethTisue:
superseded by #551. thanks @brahn for calling it to the community's attention

PR #503, opened 2016-03-16
Fixed a typo

t-mochizuki:
In the example it looked like using `reflectField`.


SethTisue:
good catch, thank you!

PR #505, opened 2016-03-22
Minor improvements

adamwy:
Removed a link to the Google Moderator website since it seems to be dead. 

Also some small fixes in macro/quasiquote guides.


SethTisue:
nice, thanks!

PR #507, opened 2016-03-24
Do not call markup "Markdown"

denis-soundcloud:
The Scaladoc documentation calls the markup format “Markdown”, which is incorrect, as the markup is not Markdown (and does not even resemble it!).

This removes the word “Markdown” and changes it to “markup”.


heathermiller:
Good point :) Thanks @denis-soundcloud!

PR #509, opened 2016-03-26
SI-9684 Don't push JavaConversions

som-snytt:
Because deprecated, because confusing.


heathermiller:
LGTM! Thanks @som-snytt :)

PR #510, opened 2016-03-28

asgs:
Added a backquote for a class name called Complex so as not to confuse it with the adjective complex.


SethTisue:
good catch. thanks!

Issue #512, opened 2016-04-01
Java's naming convention for constants is not UpperCamelCase


solimant:
On the "Naming Conventions" doc page, it says in the [Constants, Values, Variable and Methods](http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods) that: 

> Constant names should be in upper camel case... (similar to Java’s static final members)

Java does not follow upper camel case for constants naming convention, according to the [Code Conventions for the Java TM Programming Language](http://www.oracle.com/technetwork/java/codeconventions-135099.html):

> The names of variables declared class constants and of ANSI constants should be all uppercase with words separated by underscores ("_").

SethTisue:
it's defining what is "considered a constant" in this context, not asserting something about the Java convention for constants.

nonetheless, if that wasn't clear to you, that probably means the text should be altered to be clearer. a PR with a proposed improvement would be welcome.

PR #513, opened 2016-04-02
omit confusing text about Java's naming convention

solimant:
Java does not follow upper camel case for constants naming convention, according to the [Code Conventions for the Java™ Programming Language](http://www.oracle.com/technetwork/java/codeconventions-135099.html):

> The names of variables declared class constants and of ANSI constants should be all uppercase with words separated by underscores ("_")

Related issue: #512

### DIFF
--- a/style/naming-conventions.md
+++ b/style/naming-conventions.md
@@ -249,7 +249,7 @@ easily transform even the simplest code into symbolic soup.
 
 Constant names should be in upper camel case. That is, if the member is
 final, immutable and it belongs to a package object or an object,
-it may be considered a constant (similar to Java's `static final` members):
+it may be considered a constant:
 
     object Container {
       val MyConstant = ...


mpociecha:
Obviously this text is related to 'what the constant is' and not to the naming convention. What is not clear to you?

Btw, almost everyone knows what the convention in Java is.

solimant:
Hi Michał, I am new to Scala, and had come to that documentation page to follow proper naming convention in my Scala code. Initially, I was following Java naming convention for constants, but realized that in Scala, the convention is different. And that was when I stumbled upon that part of the sentence. I now get the meaning -- after you elaborated -- that it is referring to how a constant is defined in Scala, compared to how it is defined in Java as ```static final```. I must admit that the wording is confusing because the person reading the text would have their mind tuned on "naming conventions". I will close this pull request and suggest a reword, if that is preferable.

PR #517, opened 2016-04-08
fix #515 seq method lengthCompare now said to take Int

### LINKED ISSUE #515: lengthCompare is said in doc to take seq as param but does not
Here:
http://docs.scala-lang.org/overviews/collections/seqs.html
This example is given:

xs.lengthCompare ys  "Returns -1 if xs is shorter than ys, +1 if it is longer, and 0 is they have the same length. Works even if one if the sequences is infinite."

(BTW: the example should not have a dot in operator syntax)

When I do this:
$ scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_66).

scala> val xs = (1 to 10).toVector
xs: Vector[Int] = Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)

scala> xs lengthCompare 10
res0: Int = 0

scala> xs lengthCompare xs
<console>:13: error: type mismatch;
 found   : Vector[Int]
 required: Int
       xs lengthCompare xs
                        ^
Is the method signature wrong or is the overview doc wrong? Should the parameter be an Int or another sequence?
I can try to fix it, if it is the overview doc, with some guidance of how to...

bjornregnell:


heathermiller:
LGTM, thanks @bjornregnell!

PR #527, opened 2016-05-14
Added a note on implicit return

asgs:
Please modify the sentences if necessary. However, I'd like this to be explicitly stated, so that the beginners are aware.


SethTisue:
thanks! I changed the wording somewhat in 62d1fa7db36aa0d4e729027b51afebf993b88932

PR #530, opened 2016-05-23
the first public draft of the inline/meta SIP

xeno-by:
/cc @odersky @adriaanm @retronym


sips/drafts/sip-inlinemeta.md
```diff
@@ -0,0 +1,191 @@
...
+ - parameters of inline methods, e.g.
+
+        inline def pow(b: Double, inline n: Int): Double = {
+          if (n == 0) 1
+          else pow(b, n - 1)
+        }
+
+The previous rule of regarding a `final val` with no explicit type as
+a compile-time constant (inherited from Java) is dropped. Instead we
+write such `val`s now as `inline`.
```
dwijnand: Please note that another thing that `final val` does is make its type more precise:

```scala
scala> val x = 1
x: Int = 1

scala> final val x = 1
x: Int(1) = 1
```

So you would be throwing that out as well..

xeno-by:
I am retracting this draft from public discussion. 

Over the last few months, I have created an extensive description of the new macro system, expanding on the points briefly touched in this pull request and filling in the blanks. Before the end of this month, I will finish the write-up and post it as a pre-SIP at the scala-sips mailing list, as required by the new process (http://docs.scala-lang.org/sips/sip-submission.html). I will also post a link to the new discussion here.

xeno-by:
I have just published a follow-up proposal at https://gist.github.com/xeno-by/9d7a709b1ba7c2ee64cfedcc5d264bd5. In order to centralize the discussion, post your questions, feedback and ideas to http://gitter.im/scalameta/sips.

odersky:
@axel22 Dotty's implementation of SIPs knows about the position in the inlined method and in the call. For example, here's a typical error message:

    C_1.scala:5: error: class D in package p cannot be accessed
        val d = new D()     // error (when inlined): not accessible
                    ^

    ... this location is in code that was inlined at Test_2.scala:5:
    
        c.inl()
             ^
    one error found

axel22:
@odersky That looks great. I'm looking forward to trying it out!

PR #536, opened 2016-07-02
Fixed small typo

travismontoya:
I think "...since they correspond to the type of the body..." sounds correct, rather than "...since they corresponds to the type of the body"


heathermiller:
Yep! You're absolutely right. Thank you for catching it and fixing it!

PR #550, opened 2016-07-30
Swap star-space for space-star

som-snytt:
Everyone is confused why the style page recommends the odd
column for asterisks.

### DIFF
--- a/style/scaladoc.md
+++ b/style/scaladoc.md
@@ -22,10 +22,10 @@ experienced users, but can be invaluable for newcomers).
 The general format for a Scaladoc comment should be as follows:
 
     /** This is a brief description of what's being documented.
-      *
-      * This is further documentation of what we're documenting.  It should
-      * provide more details as to how this works and what it does. 
-      */
+     * 
+     *  This is further documentation of what we're documenting.  It should
+     *  provide more details as to how this works and what it does. 
+     */
     def myMethod = {}
 
 For methods and other type members where the only documentation needed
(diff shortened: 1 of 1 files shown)


soc:
What's the reason for changing this again? I think when this was discussed the last time, it was only decided not to touch all the existing code, not to give up on the improvement in general.

som-snytt:
I was bored waiting for Windows 10 to finish installing.

Today: https://groups.google.com/forum/#!topic/scala-user/q-Vw03zcIVs

Here is the discussion: http://lampscalaw3dev.epfl.ch/old/node/11291.html

> I think the conclusion was not to enforce the right-aligned style. I would be happy if the style guide would not require it either.
> 
> Cheers
> 
>  -- Martin

with reference to previous: https://groups.google.com/forum/#!msg/scala-debate/LMKC2Pw6JT8/iHaUbUOy4pwJ

Not that we have to make Martin happy.

SethTisue:
I support merging this on the grounds that it's a clear improvement to at least mention both styles as they are both in wide use and there's no clear consensus. (I don't really care if the style guide gently encourages one over the other.)

SethTisue:
thanks A!

eddsteel:
In the future could there be more publicity about changes like this? It's common for organizations to have style guides that implicitly recommend whatever the official guideline is, and tools to enforce it/reformat code. Even small changes like this can have a big impact, especially when there's no clear consensus and the "standard" as documented here is flipped from one to the other. 

Scaladoc formatting this is an unfortunate combination of insignificant enough that people "just use the official standard" and significant enough that tools/docs are now "wrong" and need to be changed.

som-snytt:
There is no recommendation per se, so it would be helpful to illustrate all three common comment formats, plus single-line style. Maybe add language to the effect, "There is no recommended comment style per se."

[sample abuse of comment](https://github.com/adriaanm/scala/blob/5a6acca2d064d79b25d9fd5911477aa1d930a52b/src/repl-frontend/scala/tools/nsc/interpreter/shell/Reporter.scala#L78-L81):

```
  /** Whether very long lines can be truncated.  This exists so important
    *  debugging information (like printing the classpath) is not rendered
    *  invisible due to the max message length.
    */
```

SethTisue:
> In the future could there be more publicity about changes like this

@eddsteel it's very difficult to turn "user X [you] wishes they had known about change Y [this]" into an improved general approach for making announcements. practically any change in any of the repositories under the scala organization is something at least some people will want to know about, and there is an absolute torrent of such changes, many more than we can reasonably announce other than via GitHub notifications. so, in other words, I feel for you, but I'm also not really seeing how your suggestion is actionable.

a better place to discuss this, if you want to take it up further, would be contributors.scala-lang.org, since this isn't really about _this_ change in particular.

eddsteel:
Thanks! I started [a discussion](https://contributors.scala-lang.org/t/quiet-changes-to-the-style-guide/923), we can continue there if you're interested.

I like your suggestion @som-snytt but it would still require updates to several tools and public style guides, which is why I think the impact of this change was perhaps underestimated.

PR #551, opened 2016-07-30
Enhance doc of iterator re-use

som-snytt:
Add more words about not reusing an iterator after calling
a method on it. Clarify what happens with `duplicate` and
with `BufferedIterator`.

Iterator implementation characteristics may guide one's
choice of idiom.

This is intended to supersede #501 .

Drop symbolic fold from table. Add text about `++` in context of buffering and efficiency.


som-snytt:
http://stackoverflow.com/questions/38713347/concatenation-of-iterators

soc:
LGTM.

PR #553, opened 2016-08-01
Fix HTML redirects on scaladoc pages

ashawley:
These two redirects are broken:

- http://docs.scala-lang.org/overviews/scaladoc/basics.html
- http://docs.scala-lang.org/overviews/scaladoc/usage.html

Add new layout, "redirected", based on blog post by Kanishk Kunal at
<https://superdevresources.com/redirects-jekyll-github-pages/>.

- `_layouts/redirected.html`

Convert two HTML-based redirects to use new layout for redirects:

- `overviews/scaladoc/basics.md`
- `overviews/scaladoc/usage.md`

### DIFF
--- a/_layouts/redirected.html
+++ b/_layouts/redirected.html
@@ -1,0 +1,16 @@
+---
+layout: default
+---
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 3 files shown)


heathermiller:
Ah, great, thanks for catching this and taking care of it for us :)

PR #561, opened 2016-08-26
correct the name of future method

li-zhirui:
The name of future constructor should be `Future`, but it was written as
`future` by mistake.


SethTisue:
replaced by #562

PR #562, opened 2016-08-26
correct the method name for creating Future

li-zhirui:
The Future is constructed using `Future`, but name was written as
`future` by mistake.


SethTisue:
thank you, good catch! the lower case `f` here is left over from how things worked in an old Scala version.

SethTisue:
tweaked further in 4085611303405d100e92c68bca8876578e5d02dc

PR #568, opened 2016-09-09
class C(x: R) NOT same as class C(private val x: R)

yanisIk:
It says on the [cheatsheet](http://docs.scala-lang.org/cheatsheets/) that 
`class C(x: R)` is the same as `class C(private val x: R)` which is not true.

In this case (`class C(x: R)`), x is just a parameter that the constructor can receive and is only a local variable inside the constructor and not a member of the class.


SethTisue:
I agree the inaccuracy should be fixed, but I'm not sure this is the right way to address it. I think the spirit behind saying that `(x: R)` is like `private val x: R` was to convey that unlike a Java constructor, `x` is in scope in the entire class body — making it extremely close kin to `private[this] val x: R`.

I don't think it's accurate wording to say that `x` is "is only a local variable inside the constructor", because it can be referenced in method bodies in the class, not only in code in the class body that will become part of the constructor.

yanisIk:
I understand what this is trying to show, but I think this should not be in the cheatsheet, the cheatsheet should be simple and not ambiguous.
For example, I just began to learn Scala and I was thinking the whole time that not specifying the modifier was the same as specifying private and I just recently found out that it's not true and I'm probably not the only one falling into this.

And thanks for the correction, I didn't know that it was also available to the methods and I see now why it is considered almost like a private member.

SethTisue:
interested in getting this across the finish line...?

SethTisue:
since this seems abandoned I've opened https://github.com/scala/scala.github.com/issues/620 on it

PR #570, opened 2016-09-14
[context-bound-docs] change deprecated `ClassManifest` to `ClassTag`

dk14:
http://stackoverflow.com/questions/39486704/what-is-classmanifest-in-scala/39486938#39486938

I also added an example about migration from view bounds. I think, maybe we should also add a warning about implicit view abuse (like `implicit def(a: Any): String = a.toString`)


dk14:
Linked issue: https://github.com/scala/scala.github.com/issues/210

SethTisue:
thank you!

PR #572, opened 2016-09-16
Fix duplicate words

HairyFotr:
Trust me on the bosnian one, I'm also from the balkans :)


heathermiller:
Thanks! LGTM (and I trust you on the Bosnian one :P)

PR #573, opened 2016-09-19
shorten long name to fit into the provided space

### LINKED ISSUE #421: Lines of code are miswrapped in one coding style example
The example at 
http://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments

is supposed to show when lines should be wrapped. However, lines 10 and 11 are wrapped again in my browser by the code display, and that makes the example unnecessarily hard to understand. Instead of:

```scala
val myOnerousAndLongFieldNameWithNoRealPoint = foo(someVeryLongFieldName,
```
you see
```scala
val myOnerousAndLongFieldNameWithNoRealPoint =
foo(someVeryLongFieldName,
```
If you're careful enough, you can notice that the continuation line lacks a line number, but it'd be better to avoid the wrapping.

The layout fixes the line length to which the code is wrapped — extending that would be a more invasive change. Hence it might be best to simply shorten the line.

jarrodwb:
fixes #421

### DIFF
--- a/style/indentation.md
+++ b/style/indentation.md
@@ -73,7 +73,7 @@ spaces to achieve alignment. In such cases, the invocation itself should
 be moved to the next line and indented two spaces:
 
     // right!
-    val myOnerousAndLongFieldNameWithNoRealPoint = 
+    val myLongFieldNameWithNoRealPoint =
       foo(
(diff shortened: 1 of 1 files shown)


SethTisue:
Thanks!

PR #574, opened 2016-09-21
Remove Erik and Andrew from the Committee

jvican:
Erik is leaving because of time availability.

Andrew is leaving because of personal reasons.


jvican:
Merging, will report on this in the next meeting.

PR #577, opened 2016-09-22
fixed a bunch of dead links

jarrodwb:
My editor, Atom, is also making a bunch of changes related to
whitespace. :-(

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -88,7 +88,7 @@ The rest of the document should, of course, be written in [Markdown](http://en.w
 
 At the moment, `RELEVANT-CATEGORY` corresponds to only a single category, "core," because we are currently focusing on building up documentation of core libraries. However, expect more categories here in the future.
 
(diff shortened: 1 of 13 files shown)


SethTisue:
Thanks, this is terrific, the work is really appreciated.

I suggest we merge even with the whitespace changes, but for next time it would be better if you disabled the feature — it looks there's hints on that at https://github.com/atom/whitespace/issues/10

heathermiller:
(I'm guilty of also being a white space squasher, and I *love* it in pull requests. But I realize I might just be a crazy person)

heathermiller:
In any case, **THANK YOU @jarrodwb**!!

jarrodwb:
Sure @SethTisue. I will take more care about it for the next one.

I do prefer files with spaces swashed but I do not feel strongly about it if I did not write the original file.

PR #580, opened 2016-09-22
a printable version of the style guide

### LINKED ISSUE #299: Add a single-page (printable) version of the Scala Style Guide
The Scala Style Guide is an important document. At the moment it appears on many separate pages, which is great for browsing but not so good for printing out or for reading right through. It would be great if there was a single-page version available for these purposes.

I'm very happy to do the work necessary to make this happen, but I'm not sure where to start. Please can someone point me at the code that generates HTML from the .md pages in /style?

Thanks.

--KW 8-)

jarrodwb:
You can of course print to a pdf in most browsers I know of.
Fixes #299

I was not sure where to put the link for the one page version of the
style guide so I just put it on the index page.


jarrodwb:
closed from lack of interest.

PR #581, opened 2016-09-23
Make accepted clause clearer

jvican:
Committee asks SIP authors to update the spec and implementation after
the final verdict.


dwijnand:
The "release date" is confusing, at least to me. Shouldn't it be more like a "target Scala version", eg. "the committee proposes this be targetted for Scala 2.12.1"?

jvican:
Good suggestion @dwijnand.

PR #583, opened 2016-09-23
Final improvement in `accepted` clause

jvican:
After some thought, I believe that the compiler maintainers have a
better judgement of when proposals can be merged, especially because of
the compatibility guarantees that the Scala compiler contract enforces.

In this way, I propose that the Compiler team chooses the most
appropriate version to include the accepted proposal.


heathermiller:
When changing the wording of the SIP document, which most people have already agreed with, I think it's best not to merge until someone (me, Seth, just anyone else on the SIP committee) has reviewed your changes.

xeno-by:
LGTM.

jvican:
I will try to be more rigorous in the future as well, @heathermiller.

PR #586, opened 2016-09-26
Add improvements to completed and pending SIPs

jvican:
Fix styles, vote-status, meeting dates and formatting errors.


_includes/allsids.txt
```diff
@@ -8,9 +8,15 @@
...
       {% elsif post.vote-status == 'postponed' %}
-        <span class="label warning">Postponed</span>
+        <span class="label light-warning">Postponed</span>
+      {% elsif post.vote-status == 'under review' %}
+        <span class="label warning">Under review</span>
+      {% elsif post.vote-status == 'numbered' %}
+        <span class="label notice">Numbered</span>
       {% elsif post.vote-status == 'not accepted' %}
-        <span class="label danger">Not Accepted</span>
+        <span class="label danger">Rejected</span>
```
heathermiller: Haha, so historical fact: we originally used the phrase  Martin was concerned that a big red "rejected" badge was too negative so a grey "not accepted" badge was proposed.

jarrodwb: Why was it changed back to Rejected? I was surprised by the language. Just curious.

heathermiller: I don't know. This is why I pointed out the historical background. @jvican made the call.

jarrodwb: Being rejected is one of the greatest fears for a lot of people.

heathermiller: Euphemism is the wrong word. It is less harsh to say "not accepted" rather than "rejected." Both are accurate, one just has a less negative connotation, which after you spend a lot of time proposing a change only to find that it wasn't accepted, may be what one needs. Less negativity. Maybe we should go back to "not accepted," Jorge? It was a strong preference of many a few years ago to do it this way. And I would also lean on keeping "not accepted" personally, myself.

SethTisue: I think "not accepted" is better, too

jvican:
Just pushed the changes. Review by @heathermiller | @SethTisue.

SethTisue:
LGTM

PR #587, opened 2016-09-26
Fix #585: Add `jekyll-redirect-from`

### LINKED ISSUE #585: URL change when SIP is accepted or rejected breaks links everywhere
so e.g. SIP-12 is now http://docs.scala-lang.org/sips/rejected/uncluttering-control.html , which means that everyone who had linked to the old URL http://docs.scala-lang.org/sips/pending/uncluttering-control.html now has a broken link on their hands

if we are going to change the URLs, we should redirect. probably better/easier not to change them in the first place

reported by @the-Arioch at https://github.com/scala/scala.github.com/issues/555 /cc @jvican

jvican:


jvican:
Merging to see if it works.

PR #588, opened 2016-09-26
Finish off sips redirection

jvican:
It works locally. Let's try it on live.


jvican:
Should work now. Let's find out.

PR #590, opened 2016-09-28
Correct return type for Iterator in lazyMap example of Views section

CarlEkerot:
The function in the example maps a function T => U on the given Iterable
of type T, and should return an Iterable of type U.

---

This was mentioned in one of the comments over at http://docs.scala-lang.org/overviews/collections/views.html. I figured I'd make a pull request since I too stalled on this example.


SethTisue:
thanks Carl!

Jarrod: yeah, that'd make a great project for somebody

PR #592, opened 2016-10-01
Remove view bounds from FAQ

raboof:
As they're deprecated (scala/scala#2909).

I updated the 'context and view bounds' section and renamed it to 'context bounds', unfortunately it appears github doesn't recognize this in the diff..


soc:
+1. Did the same in https://github.com/soc/scala-lang/commit/548d4a41d6f55211403d94cc55692aeafc0fb568 last November.

SethTisue:
excellent, thanks!

PR #593, opened 2016-10-03
get tut to work with Travis

### LINKED ISSUE #591: Start using tut to improve quality of tutorials.
I am interested in converting the documentation site to use a `sbt` build process so that the code in the tutorials is checked at compile time. This could be done by using [tut](https://github.com/tpolecat/tut).

The documentation site started four or five years ago and is based only on `jekyll`. Things have progressed since then.

I don't think this would take to much effort to do. The only time consuming part would be to update the tutorials so that they pass the compiler. I suspect there will be quite a few bugs and typos in the tutorials.

This issue is partly inspired by #590 but there have been other issues and even some pull requests that could have been avoided if `tut` had used. For instance maybe #488 would be an example.

jarrodwb:
Fix #591 

I went with @SethTisue idea of letting Travis do all the work and not using tut for building the site.

Currently this will take all the files in the repo and pass them to tut. Tut will validate all code blocks that tut recognizes. If there is an error then the Travis build will fail. If tut does succeed in validating all the files then Travis will build the site like before. None of the output files of tut are discarded.

At the moment none of the tutorials use code blocks that tut recognizes. At some point in the future the old tutorials could be modified to allow for the checks. All future tutorials should use the correct type of code blocks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,5 @@ install:
 
 script:
 - ./scripts/run-tut.sh
+- rm -r tut-tmp
 - bundle exec jekyll build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-# opt-in to Travis new infrastructure
-sudo: false
+sudo: true
 
 language: ruby
 rvm: 2.0.0
@@ -8,7 +7,9 @@ rvm: 2.0.0
 # will kick in and use `vendor`, ignoring our BUNDLE_PATH
 # declaration in `.bundle/config`
 install:
+- ./scripts/install-coursier.sh
 - bundle install
 
 script:
+- ./scripts/run-tut.sh
 - bundle exec jekyll build
--- a/contribute.md
+++ b/contribute.md
@@ -39,7 +39,7 @@ It's statically generated from [Markdown](http://en.wikipedia.org/wiki/Markdown)
 
 The markdown syntax being used supports [Maruku](http://maruku.rubyforge.org/maruku.html) extensions, and has automatic syntax highlighting, without the need for any tags.
 
-Additionally [tut](https://github.com/tpolecat/tut) is used during pull requests to valid Scala code blocks. To use this feature you must use the backtick notation as documented by tut. Note that only validation are done. The output files from tut are not used in the building of the tutorial. Either use `tut` or `tut:fail` for your code blocks.
+Additionally [tut](https://github.com/tpolecat/tut) is used during pull requests to validate Scala code blocks. To use this feature you must use the backtick notation as documented by tut. Note that only validation is done. The output files from tut are not used in the building of the tutorial. Either use `tut` or `tut:fail` for your code blocks.
 
 ## Submitting Docs
 
--- a/contribute.md
+++ b/contribute.md
@@ -39,6 +39,8 @@ It's statically generated from [Markdown](http://en.wikipedia.org/wiki/Markdown)
 
 The markdown syntax being used supports [Maruku](http://maruku.rubyforge.org/maruku.html) extensions, and has automatic syntax highlighting, without the need for any tags.
 
+Additionally [tut](https://github.com/tpolecat/tut) is used during pull requests to valid Scala code blocks. To use this feature you must use the backtick notation as documented by tut. Note that only validation are done. The output files from tut are not used in the building of the tutorial. Either use `tut` or `tut:fail` for your code blocks.
+
 ## Submitting Docs
 
 For one to contribute a document, one must simply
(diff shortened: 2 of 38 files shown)


SethTisue:
> At some point in the future the old tutorials could be modified to allow for the checks

this looks cool — but I'm not sure it makes sense to merge unless/until it includes at least one example where tut is actually doing its job? for two reasons, 1) so we can know it's even working (by inserting an error into the example), 2) as an example for others to follow

SethTisue:
also, is there is somewhere in this repo this could/should be documented?

jarrodwb:
What tutorial would you like me to modify?

SethTisue:
pick one or more sections of the language tour, I guess? judging from the amount of Disqus comments those pages get, at least, they seem to get an especially large amount of traffic

jarrodwb:
@SethTisue I implemented your suggestions and the PR is ready for review.

soc:
@jarrodwb Is there a way to configure tut to use a different code block (like ` ```scala `?)

jarrodwb:
I am not sure. Tut has a gitter channel though.

jarrodwb:
I am not sure. Tut has a gitter channel though.

contribute.md
```diff
@@ -39,6 +39,8 @@ It's statically generated from [Markdown](http://en.wikipedia.org/wiki/Markdown)
 
 The markdown syntax being used supports [Maruku](http://maruku.rubyforge.org/maruku.html) extensions, and has automatic syntax highlighting, without the need for any tags.
 
+Additionally [tut](https://github.com/tpolecat/tut) is used during pull requests to valid Scala code blocks. To use this feature you must use the backtick notation as documented by tut. Note that only validation are done. The output files from tut are not used in the building of the tutorial. Either use `tut` or `tut:fail` for your code blocks.
```
SethTisue: s/valid/validate/
s/are done/is done/

SethTisue:
@tpolecat does this look reasonable to you?

SethTisue:
@jarrodwb have you verified that Travis fails if there's an error in one of the `tut` sections?

tpolecat:
Why are you not using the output?

.travis.yml
```diff
@@ -1,5 +1,4 @@
-# opt-in to Travis new infrastructure
-sudo: false
+sudo: true
```
jarrodwb: I need to be able to use `sudo` and similar commands. The main thing is to be able to `chmod` coursier. This is disallowed on the container version of Travis. The `sudo: true` puts it all on a virtual machine. At least this is my understanding.

jarrodwb:
@SethTisue Yes. I am sure that Travis fails when tut throws an exception. I tested it out locally and also on this PR. There a few commits where I submitted broken code and then fixed it again.	990085e and 3a052b6

jarrodwb:
@tpolecat The main reason is that the current project is pretty old and it would take a lot of changes to use tut to preprocess the tutorials. The current tutorials don't use GFM code blocks and were written without the idea of having the output of code blocks inserted into the site. Also `sbt-site` and `sbt`  are currently not being used to generate the site. I would have had to completely redo the directory structure to be able to use the more modern approach. In the future hopefully we can go in this direction. If a new tutorial is written and added to the site then it would make sense to use the output of tut. Currently we are only dealing with rather old tutorials that are only being maintained and not expanded.

One other major reason that @SethTisue asked me to do in this way. ;) See #591

SethTisue:
@tpolecat our main concern about using the output was that the deployment pipeline would have to change, as discussed at #591. we might still want to go on and do that, but we were hoping to get this much integrated and merged first before worrying too much about a more ambitious option

.travis.yml
SethTisue: we can't even `chmod +x` things we just downloaded ourselves in our own home directory? that surprises me, I wouldn't expect superuser privileges to be necessary. are you sure?

tpolecat:
@SethTisue this looks reasonable to me, as long as the final rendering engine can cope with the `'''tut` markup … I didn't see anything in the PR that rewrites those. There is an old verify-only mode as a [feature request](https://github.com/tpolecat/tut/issues/59) that you can feel free to upvote. It might have made this easier for you.

SethTisue:
> as long as the final rendering engine can cope with the '''tut markup

when I tested this locally it didn't seem to faze Kramdown. @jarrodwb, you have seen the same in your testing? perhaps you could check and see if our `bundle exec...` setup is still in sync with whatever GitHub Pages is using these days? because if it isn't, we could get a nasty surprise after this was merged.

jarrodwb:
Everything works for me locally. I will double check GitHub Pages and see what I can come up with.

jarrodwb:
Looks good to me. When we use the github-pages gem we get a local version of the same setup that GitHub uses. If this gem works they way it is advertised then we are good to go. :)

https://github.com/github/pages-gem

jarrodwb:
@SethTisue Any chance we can get this merged in?

SethTisue:
🎉 thanks Jarrod!

PR #594, opened 2016-10-04
suggested points of contact for contributors

SethTisue:
review by @heathermiller (you don't want to be the single suggested
point of contact anymore, I'm guessing?)


SethTisue:
@heathermiller I'm going to go ahead and merge, we can always revert or make further changes

PR #596, opened 2016-10-07
`Stack` does not support constant time `append`

Atry:


SethTisue:
good catch, thanks

xuwei-k:
BTW, `immutable.Stack` is deprecated.
- https://github.com/scala/scala/commit/3cc99d7b4aa43b1b06cc837a55665896993235fc#diff-cf59d0ff92484ae8ccd469ca6f5b00d0R49
- https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/immutable/Stack.scala#L49

PR #597, opened 2016-10-07
Revert .travis.yml back to sudo: false w/ workaround

dwijnand:
For some reason calling chmod this way TravisCI is fine with.

This way the build is back to using the faster container-based infrastructure.


jarrodwb:
Cool! I learned something new today. Thanks.

The Travis log looks good me. https://travis-ci.org/scala/scala.github.com/builds/165956607

SethTisue:
thanks Dale!

Issue #547, opened 2016-07-17
Improvement/Rewrite for Case Class tutorial/tour


jtcwang:
I'm talking about [this page](http://docs.scala-lang.org/tutorials/tour/case-classes.html).

Case class is probably one of the best features of Scala and heart-winner for anyone interested to Scala, but I feel like the current documentation doesn't do it justice. The comment section of that article echos my feeling that it's not really beginner friendly.

Here are the improvements that I think we can make, assuming that the majority of our audience is from an imperative background:

* Better example
  * Most of the people looking at that page won't be familiar with parsing. Switching over to real world 'objects' may help a lot so they don't have to wonder what `App` and `Fun` means
* Show the equivalent java code (e.g. generated `hashCode`, `equals` etc etc)
  * Show how succinct Scala is!

If people is happy with my proposal, I'll go ahead and make it happen.

SethTisue:
I support both of your proposed changes.

jtcwang:
I didn't end up showing the equivalent Java code. I'm not sure whether it'll make the post too long-winded. I can add it if you think it's helpful.

SethTisue:
can you make a pull request and let's discuss there?

jtcwang:
Oh sorry. The PR has already been made (#599). I'll move my comment there.

PR #599, opened 2016-10-09
Reworked case class tutorial #547

jtcwang:
Reworked Case Class tutorial with a more approachable example, and some more explanation/examples on why they are awesome. Feel free to butcher my post. I'm not the best with words :stuck_out_tongue:


tutorials/tour/case-classes.md
```diff
@@ -10,80 +10,125 @@ tutorial-next: pattern-matching
...
-Fun("x", Fun("y", App(Var("x"), Var("y"))))
+val title = emailFromJohn.title
+println(title) // prints "Greetings From John!"
 ```
 
-The constructor parameters of case classes are treated as public values and can be accessed directly.
+With case classes, you cannot mutate their fields directly. Instead, you make a copy using the `copy` method.
+As seen below, you can replace just some of the fields:
 
 ```tut
```
jarrodwb: Since this code block is designed to fail then `tut` should be `tut:fail`. This will allow tut to give an error if the code block compiles. It is a little weird.

https://github.com/tpolecat/tut

Also check the Travis logs if they fail. They will show you where to look to fix the problem. :-)

https://travis-ci.org/scala/scala.github.com/builds/166207587

tutorials/tour/case-classes.md
```diff
@@ -10,80 +10,125 @@ tutorial-next: pattern-matching
...
-val x = Var("x")
-println(x.name)
+emailFromJohn.title = "Goodbye From John!" // This is a compilation error. We cannot assign another value to val fields, which all case classes fields are by default.
+
+val editedEmail = emailFromJohn.copy(title = "I am learning Scala!", body = "It's so cool!")
+
+println(emailFromJohn) // prints "Email(john.doe@mail.com,Greetings From John!,Hello World!)"
+println(editedEmail) // prints "Email(john.doe@mail.com,I am learning Scala,It's so cool!)"
```
jarrodwb: Since this code block is designed to fail then `tut` should be `tut:fail`. This will allow tut to give an error if the code block compiles. It is a little weird.

https://github.com/tpolecat/tut

Also check the Travis logs if they fail. They will show you where to look to fix the problem. :-)

https://travis-ci.org/scala/scala.github.com/builds/166207587

jtcwang: Thanks. I'll fix this up.

jtcwang: @jarrodwb I have fixed this a TravisCI run is passing.

jarrodwb:
Hey Jacob. Thanks for getting Travis to pass.

Unfortunately I don't have the authority to merge your PR into master. It seems like Seth is the main person that reviews PRs and merges them in. Another option that maybe a good idea is to recruit some people from the Gitter channels to do a review . There are generally a lot of people in the scala/scala channel. Scala/contributors is another more official option. They may even give some pointers on the content of the tutorial.

Keep up the good work! :+1: 

@SethTisue @jtcwang

tutorials/tour/case-classes.md
```diff
@@ -10,80 +10,128 @@ tutorial-next: pattern-matching
...
 
 ```
-Var(x) == Var(x) => true
-Var(x) == Var(y) => false
+They are equal!
+SMS is: SMS(12345, Hello!)
 ```
 
-It only makes sense to define case classes if pattern matching is used to decompose data structures. The following [object](singleton-objects.html) defines a pretty printer function for our lambda calculus representation:
+With Case Classes, all the necessary machinery required for using them in **Pattern Matching** is automatically generated by the compiler.
```
heathermiller: Just wondering; is it really necessary to point out that there's stuff generated by the compiler? Just thinking that mentions of this could be removed here, as this sort of information is generally not required to use/understand case classes.

tutorials/tour/case-classes.md
```diff
@@ -10,80 +10,128 @@ tutorial-next: pattern-matching
...
-Scala supports the notion of _case classes_. Case classes are regular classes which export their constructor parameters and which provide a recursive decomposition mechanism via [pattern matching](pattern-matching.html).
+Scala supports the notion of _case classes_. Case classes are just regular classes that are:
 
-Here is an example for a class hierarchy which consists of an abstract super class `Term` and three concrete case classes `Var`, `Fun`, and `App`.
+* Immutable by default
+* Decomposable through [Pattern Matching](pattern-matching.html)
+* Compared by structural equality instead of by reference
+* Succinct to instantiate and operate on
+
+This is achieved through auto-generated code by the Scala compiler.
```
heathermiller: See comment below

tutorials/tour/case-classes.md
jtcwang: Yeah I think I put that in because it was really useful to know that pattern matching isn't some magical thing Scala did which only applies to case classes (coming from a Java-ish background). However you are correct - if they read this tutorial on case classes before pattern matching (which is the right order, imo), then the implementation detail of pattern matching doesn't need to be mentioned.

tutorials/tour/case-classes.md
```diff
@@ -10,80 +10,128 @@ tutorial-next: pattern-matching
...
+// You got an email from special someone!
+// You got an SMS from special someone!
+
 ```
 
-In our example, the function `printTerm` is expressed as a pattern matching statement starting with the `match` keyword and consisting of sequences of `case Pattern => Body` clauses.
-The program above also defines a function `isIdentityFun` which checks if a given term corresponds to a simple identity function. This example uses deep patterns and guards. After matching a pattern with a given value, the guard (defined after the keyword `if`) is evaluated. If it returns `true`, the match succeeds; otherwise, it fails and the next pattern will be tried.
+_Case Classes_ is one of the biggest reasons why Scala code can be both concise and readable. With immutability and structural equality, working with 
+case classes are just like working with primative values (Int, Double, etc), which in turn makes writing programs in a functional style much simpler
+and intuitive.
```
heathermiller: Said this way, a much stronger point emerges. Care to try rewording?

jtcwang: Done. Let me know what you think!

SethTisue:
This looks ready to merge to me except for two things:

* Case classes are immutable by default, true, but you can also declare a case class's parameters as `var`. So the claim "With case classes, you cannot mutate their fields directly" is too strong. I don't think you need to give the `var` possibility much space, but you can't give the impression it doesn't exist, either.
* Only proper nouns should be capitalized, so e.g. it's just "pattern matching" not "Pattern Matching", "case classes" not "Case Classes", and so forth.

tutorials/tour/case-classes.md
jtcwang: This has been removed

jtcwang:
@SethTisue I've addressed your comments. Is it looking better now?

jtcwang:
Copied from my comment at #547:

I didn't end up showing the equivalent Java code as I mentioned that I will. I'm not sure whether it'll make the post too long-winded. If anyone thinks it'll be helpful, I can add it in.

SethTisue:
LGTM — further changes could always follow later. Thank you!!

heathermiller:
🎉 👏

jtcwang:
Cheers. Thanks for the thorough and prompt review :)

PR #600, opened 2016-10-18
Upgraded project docs to include incremental switch

charafau:
After discussion with @SethTisue (here: https://github.com/scala/scala-lang/issues/493 ) about compilation speed I have upgraded `jekyll` to `3.3.0` 
and upgraded docs to include `-I` incremental compilation flag. 

At first it took ~5 minutes on my machine to build the whole thing but after that only few seconds: 
```
Incremental build: enabled
      Generating...
                    done in 2.065 seconds.
```

Hope it helps with further updates. 

Here's the list of used gems:

```
➜  scala.github.com git:(feat/jekyll_incremental_upgrade) bundle show
Gems included by the bundle:
  * addressable (2.4.0)
  * bundler (1.13.5)
  * colorator (1.1.0)
  * ffi (1.9.14)
  * forwardable-extended (2.6.0)
  * jekyll (3.3.0)
  * jekyll-redirect-from (0.11.0)
  * jekyll-sass-converter (1.4.0)
  * jekyll-watch (1.5.0)
  * kramdown (1.12.0)
  * liquid (3.0.6)
  * listen (3.0.8)
  * mercenary (0.3.6)
  * pathutil (0.14.0)
  * rb-fsevent (0.9.7)
  * rb-inotify (0.9.7)
  * rouge (1.11.1)
  * safe_yaml (1.0.4)
  * sass (3.4.22)
```

For more speed improvements we would probably need to move to hugo or some other static generator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,29 +19,15 @@ This site uses a Jekyll, a Ruby framework. You'll need Ruby and Bundler installe
 
 cd into the directory where you cloned this repository, then install the required gems with `bundle install`. This will automatically put the gems into `./vendor/bundle`.
 
-Start the server in the context of the bundle:
+Start the server in the context of the bundle (use `-I` for incremental builds):
 
-    bundle exec jekyll serve
+    bundle exec jekyll serve -I
+
+`It might take around 5 minutes at first but incremental compilations will be fast.`
 
 The generated site is available at `http://localhost:4000`
 
-If you add `--watch` at the end of the command line above, Jekyll will automatically watch for changes on the filesystem, and regenerate the site. It can take a few minutes for your changes to appear. Watch Jekyll's output. When you start up you'll see this:
-
-     $ bundle exec jekyll serve --watch
-     Configuration file: /Users/ben/src/scala.github.com/_config.yml
-                 Source: /Users/ben/src/scala.github.com
-            Destination: /Users/ben/src/scala.github.com/_site
-          Generating...
-                     done.
-    Auto-regeneration: enabled for '/Users/ben/src/scala.github.com'
-
-When you change a file, this output will tell you that jekyll is regenerating. It's not done until it says `done.`
-
-    Server running... press ctrl-c to stop.
-      Regenerating: 1 files at 2014-11-29 09:19:04 ...done.
-      Regenerating: 3 files at 2014-11-29 09:21:39 ...done.
-      Regenerating: 2 files at 2014-11-29 09:25:10 ...done.
-      Regenerating: 2 files at 2014-11-29 09:27:49
+If you add `--watch` at the end of the command line above, Jekyll will automatically watch for changes on the filesystem, and regenerate the site.
 
 If you get `incompatible encoding` errors when generating the site under Windows, then ensure that the
 console in which you are running jekyll can work with UTF-8 characters. As described in the blog
@@ -61,4 +47,3 @@ There's a free markdown editor for OSX called [Mou](http://25.io/mou/). It's qui
 ## License ##
 
 All documentation contained in this repository is licensed by EPFL under a Creative Commons Attribution-Share Alike 3.0 Unported license ("CC-BY-SA"), unless otherwise noted. By submitting a "pull request," or otherwise contributing to this repository, you implicitly agree to license your contribution under the above CC-BY-SA license. The source code of this website is licensed to EPFL under the [Scala License](http://www.scala-lang.org/node/146), unless otherwise noted.
-


Gemfile
```diff
@@ -1,2 +1,8 @@
 source 'https://rubygems.org'
-gem 'github-pages', group: :jekyll_plugins
```
jarrodwb: Can we keep the dependency on github-pages. This makes it easy to assure ourselves that the site will look the same in production as it does on our personal computers.

jarrodwb:
Incremental compilation worded just fine for me with the prior Gemfile. Is there another reason for the upgrade?

charafau:
I thought that there is no specified version of Jekyll which is used. Some people were probably using 3.x (that's why incremental compilation worked) but majority of people don't know what version is used in production

jarrodwb:
Take a look at the Gamefile.lock file. Jekyll 3.0.3 is being used.

I should take a look at our documentation and see if it describes how to get local development setup properly. I am away from my computer at the moment.

jarrodwb:
Are you using `bundle exec jekyll ...`? The dependencies referenced in the README links to the Jekyll site. This should be change to the bundler site. `jekyll ...` commands should not be used without Bundler.

If Bundler is used then everyone will be using the Jekyll version listed in the Gemfile.lock file.

charafau:
Hello @jarrodwb  I'm using `bundle exec jekyll serve -I --watch` command. We can stay with jekyll 3.0.3 but I would like to add it to `Gemfile` and upgrade docs for incremental compilation.  What do you think ?

soc:
As suggested, posting this publicly...
I'm not sure it's doing the right thing. The gh-pages gem is used because Github promises that this gets you exactly the versions they use when building the site.
So I would suggest not getting rid of it, even if newer versions fix bugs, it will be more confusing because the bugs will be gone locally, but still exist on the published website until GitHub also updates their stack.

charafau:
Thank you. I understood issue. I will revert gem changes and add only `-I` switch into readme.

jarrodwb:
I suggest keeping the changes to the README but being rid of the other changes. The current Gemfile does exactly what we need.

If we want to update the site dependences we can do that in a different PR. All that needs to be done is a 'bundle update' then commit the new Gemfile.lock file. This should update Jekyll to about 3.2.x something.

I really think that gh pages gem should control the version of Jekyll. Of course if we move the site then we can do something different.

Thanks for the changes to the README. 😀

On another note. Nice to see you Simon!

SethTisue:
according to https://pages.github.com/versions/, GitHub Pages is on Jekyll 3.2.1. so isn't that the version our Gemfile should have as well...?

jarrodwb:
See my last comment. If we want to update site dependences then we should use bundle update.

https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/#keeping-your-site-up-to-date-with-the-github-pages-gem

Hard coding the version number is s mistake. If you want to update Jekyll then we can summit another PR. It really does not have much to do with this thread.

SethTisue:
> All that needs to be done is a 'bundle update' then commit the new Gemfile.lock file. This should update Jekyll to about 3.2.x something.

sorry, I was skimming and missed this.

I tried it just now and `bundle install` succeeded, but Gemfile.lock didn't change.

I think we need to make this change in Gemfile:

```difff
-    github-pages (46)
+    github-pages (98)
```

and then run `bundle install`. which I tried but I got errors I didn't understand (`Gem::Ext::BuildError: ERROR: Failed to build gem native extension`)

jarrodwb:
I don't think the gemfile should be changed at all.

If you want to use bundle install then you need to delete the gemfile.lock file first. If you use the bundle update then you don't need to delete it. It should be changed by the update command.

The whole point of the lock file is that all the developers use the same versions of all gem dependences.

The bundle install command will install all the versions in the lock file. Also the lock file should not be edited by hand.

Your error, I think, comes from not having system dependencies. Those dependencies would be managed elsewhere, outside of the gems. Maybe some like a C library.

SethTisue:
>I don't think the gemfile should be changed at all

Why is it good to have an old version of github-pages listed in the Gemfile, instead of the current version that GitHub Pages actually uses?

jarrodwb:
I will try to explain this again when I get home in about two hours.

jarrodwb:
Okay @SethTisue I am back. I will make a pull request to update the dependencies and I will go through all of the steps to explain how it is done.

Feel free to find me in gitter if you want to chat before I get the other PR finished.

jarrodwb:
See #601 for a bit on updating dependencies.

@SethTisue I think your error you mentioned earlier is either from the `ffi` gem or the `nokogiri` gem. You are missing a system dependency somewhere. You should be able to figure out which one by looking at those two gems. I would start with `nokogiri`. It's name sounds familiar and I think I ran into a similar problem on Ubuntu a while ago.

README.md
```diff
@@ -19,29 +19,15 @@ This site uses a Jekyll, a Ruby framework. You'll need Ruby and Bundler installe
...
-If you add `--watch` at the end of the command line above, Jekyll will automatically watch for changes on the filesystem, and regenerate the site. It can take a few minutes for your changes to appear. Watch Jekyll's output. When you start up you'll see this:
-
-     $ bundle exec jekyll serve --watch
-     Configuration file: /Users/ben/src/scala.github.com/_config.yml
-                 Source: /Users/ben/src/scala.github.com
-            Destination: /Users/ben/src/scala.github.com/_site
-          Generating...
-                     done.
-    Auto-regeneration: enabled for '/Users/ben/src/scala.github.com'
+`It might take around 5 minutes at first but incremental compilations will be fast.`
```
SethTisue: Are the backquotes in this line supposed to be there?

SethTisue:
LGTM other than the one nit

README.md
charafau: I added it as indication for users, that first build might take a while. I can change it if you think it's not a good idea.

SethTisue: I don't have a strong feeling about it. I'll go ahead and merge.

The fact that it takes so long now has its own ticket: https://github.com/scala/scala.github.com/issues/607

SethTisue:
thanks @charafau and @jarrodwb!

PR #601, opened 2016-10-18
update using bundle update

jarrodwb:
Okay.

I started with a master branch that is up to date. That is I `git pull` etc.

I then ran;

```
bundle update
```

Only two files where modified. `Gemfile.lock` and `.bundle/conf`.

Then I checked the version of jekyll.

```
bundle exec jekyll --version
```
and I get `jekyll 3.2.1`.

You can also see that in the  `Gemfile.lock` file many other gems have been updated as well.

The point of this is that the actual `Gemfile`, opposed to the `Gemfile.lock` file, was not changed. It is working exactly has it should be. It keeps our site working with the same versions of all the gems as is used on gh-pages.

Also note that the `gh-pages` gem was automatically updated as well. 46 -> 98 :-)


SethTisue:
awesome

SethTisue:
> The point of this is that the actual Gemfile, opposed to the Gemfile.lock file

got it. understand now

PR #605, opened 2016-10-20
avoid postfix syntax in operator tour

SethTisue:
since it's no longer a recommended part of the language

reported by Sebastian Nickel at
http://docs.scala-lang.org/tutorials/tour/operators.html#comment-2954987627

### DIFF
--- a/tutorials/tour/operators.md
+++ b/tutorials/tour/operators.md
@@ -10,28 +10,28 @@ tutorial-next: automatic-closures
 tutorial-previous: local-type-inference
 ---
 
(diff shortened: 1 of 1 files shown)


SethTisue:
@jarrodwb idea, can we enable `-Xfatal-warnings` with `tut`? then tut would have caught the usage of postfix syntax

jarrodwb:
Maybe. I don't know off the top of my head but that seems like something that should work. I will take a look at it and let you know.

SethTisue:
I made a separate ticket on the `-Xfatal-warnings` issue (#606)

SethTisue:
merging this on the fast-track, to facilitate #606. I'll respond promptly to any post-merge feedback

PR #611, opened 2016-10-27
Remove array from immutables tree

Nydhal:
Remove array from immutables tree
As suggested in the comments here  : http://docs.scala-lang.org/overviews/collections/overview.html


adriaanm:
LGTM, thank you!

PR #612, opened 2016-10-30
Link to an associated tutorial entry on singletons

cevantes:


tutorials/tour/classes.md
```diff
@@ -43,7 +43,7 @@ object Classes {
 }
 ```
 
-The program defines an executable application Classes in form of a top-level singleton object with a `main` method. The `main` method creates a new `Point` and stores it in value `pt`. Note that values defined with the `val` construct are different from variables defined with the `var` construct (see class `Point` above) in that they do not allow updates; i.e. the value is constant.
+The program defines an executable application Classes in form of a top-level [singleton object](singleton-objects.html) with a `main` method. The `main` method creates a new `Point` and stores it in value `pt`. Note that values defined with the `val` construct are different from variables defined with the `var` construct (see class `Point` above) in that they do not allow updates; i.e. the value is constant.
```
jarrodwb: I prefer links without the extention. It gives a cleaner experience. 

```
[singleton object](singleton-objects)
```

cevantes: One can probably do it in one pass with sed

jarrodwb: For sure. It would still need different PR. :+1:

heathermiller:
LGTM, thanks guys!

PR #614, opened 2016-11-03
Update fr translation to last en cheatsheet content and fix some errors

njouanin:


jarrodwb:
Thanks. I don't read french but it looks good. :+1:

fr/cheatsheets/index.md
```diff
@@ -55,12 +55,12 @@ language: fr
...
+|  `for (x <- xs if x%2 == 0) yield x*10` _est équivalent à_ <br>`xs.filter(_%2 == 0).map(_*10)`                    |  compréhension avec for: filter/map |
+|  `for ((x,y) <- xs zip ys) yield x*y` _est équivalent à_ <br>`(xs zip ys) map { case (x,y) => x*y }`              |  compréhension avec for : liaison déstructurée |
+|  `for (x <- xs; y <- ys) yield x*y` _est équivalent à_ <br>`xs flatMap {x => ys map {y => x*y}}`                  |  compréhension avec for : produit cartésien. |
+|  `for (x <- xs; y <- ys) {`<br>    `println("%d/%d = %.1f".format(x,y, x*y))`<br>`}`                     |  compréhension avec for : à la manière impérative <br>[sprintf-style](http://java.sun.com/javase/6/docs/api/java/util/Formatter.html#syntax) |
+|  `for (i <- 1 to 5) {`<br>    `println(i)`<br>`}`                                                        |  compréhension avec for : itère jusqu'à la borne supérieure comprise. |
+|  `for (i <- 1 until 5) {`<br>    `println(i)`<br>`}`                                                     |  compréhension avec for : itère jusqu'à la borne supérieure non comprise. |
```
sjrd: I wonder whether "for en compréhension" would be a more exact translation of "for comprehension". After all, this what we use in math: we can define "un ensemble en compréhension" (the syntax { x € Z | x > -10 }).

But that's French pedantry. TBH when I speak French I say "for comprehension" in English anyway. I would stick to the English term for this.

njouanin:
Hi,
Native French speaker too. I also asked myself how to translate ´for comprehension' in french so it sounds correct without loosing the meaning... I didn't know the mathematic expression you mentioned. I also agree on keeping the english term for this kind of expressions. I'll correct and add some emphasis on the english words.
I also think that translators should share a common glossary for common expressions, so they are always translated with the same terms.

PR #623, opened 2016-11-12
bump Scala version to 2.12.0

SethTisue:
ha, wow, embarrassing this got missed for the entire 2.11.x cycle

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,7 +15,7 @@ keywords:
 - Document
 - Guide
 
-scala-version: 2.10.0
+scala-version: 2.12.0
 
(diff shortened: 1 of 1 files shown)


SethTisue:
if you visit for example http://docs.scala-lang.org/overviews/collections/arrays.html, the link to "Array" at the top still points to the 2.10.0 Scaladoc. this fixes it to point to the 2.12.0 Scaladoc

jarrodwb:
I noticed it but figured you guys knew what you were doing. ;-) :-P

Actually I don't remember why I did not fix it myself. Lesson learned.

PR #626, opened 2016-11-16
Provide better example of upper type bound usage.

tashoyan:
As discussed in comments [here](http://docs.scala-lang.org/tutorials/tour/upper-type-bounds)


SethTisue:
thanks!

PR #628, opened 2016-11-17
Added an example on how to use anonymous functions in the "Anonymous …

jejo86:
…Function Syntax" page (tutorials/tour/anonymous-function-syntax.md).


jejo86:
I didn't know how to compile the code, that is why I asked about it on this page:

https://getsatisfaction.com/scaladocs/topics/how-do-i-check-if-contribution-is-free-of-errors

I noticed in your log you are running a

./scripts/run-tut.sh

script on linux with ruby installed. Since I don't have all this and it's too much effort for the small change I wanted to commit, we can just leave it out.

Thanks anyways for trying to merge my changes.

SethTisue:
the Travis log shows:

```
[tut] compiling: ./tutorials/tour/anonymous-function-syntax.md
[tut] *** Error reported at /home/travis/build/scala/scala.github.com/tutorials/tour/anonymous-function-syntax.md:58
<console>:2: error: illegal start of definition
package tour
^
```

SethTisue:
I guess tut doesn't like package declarations (`package tour`). but actually, I see no need to include that. if I remove it locally, tut succeeds. so if you make that change, we can merge this

tutorials/tour/anonymous-function-syntax.md
```diff
@@ -51,3 +51,33 @@ Function1[Int, Int]
...
+  var anonymousIncrementFunction = (x: Int) => x + 1
+
+  /**
+   * Main method
+   * @param args application arguments
+   */
+  def main(args: Array[String]) {
+
+    // Create an integer to test the anonymous function with
+    var myInteger: Int = 0;
```
SethTisue: I would suggest omitting `: Int`. (It's idiomatic in Scala to omit type declarations when the type is obvious, as it is here.)

You can omit the semicolon too.

tutorials/tour/anonymous-function-syntax.md
```diff
@@ -51,3 +51,33 @@ Function1[Int, Int]
 Function2[Int, Int, String]
 Function0[String]
 ```
+
+The following example shows how to use anonymous function of the beginning of this page
+
+```tut
+package tour
```
SethTisue: The `tut` error you're getting will go away if you simply omit this line, which isn't necessary anyway.

tutorials/tour/anonymous-function-syntax.md
```diff
@@ -51,3 +51,33 @@ Function1[Int, Int]
...
+   * @param args application arguments
+   */
+  def main(args: Array[String]) {
+
+    // Create an integer to test the anonymous function with
+    var myInteger: Int = 0;
+
+    println(myInteger) // Prints: 0
+
+    myInteger = anonymousIncrementFunction(myInteger)
```
SethTisue: I would suggest that the example avoid using `var` at all. In idiomatic Scala code, it's not normal for `var` to appear unless it's truly needed.

tutorials/tour/anonymous-function-syntax.md
jejo86: I updated that part.

tutorials/tour/anonymous-function-syntax.md
jejo86: I couldn't ommit the var in the two places it appeared, else the code did not compile. I guess it is 'truly needed' in those cases.

SethTisue:
Thanks, this is now closer to being mergeable.

But — and I'm sorry, I don't mean to be difficult — but there's really no reason to use `var` here, and it's really not a good idea for introductory Scala materials to send a signal to Scala newcomers that unnecessary use of `var` is normal, when it really isn't.

jejo86:
hey, thats all right. thanks for the feedback, but i really don't feel comfortable applying your requested change, since the code does not compile on my machine. 

i am using eclipse mars with scala ide v4.4.1, i tried both scala library 2.10.6 and 2.11.8 and just to make sure, i additionally tried to compile the code directly in my CMD running the scala interpreter. this is the output of the latter:

`Welcome to Scala version 2.11.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_20).
Type in expressions to have them evaluated.
Type :help for more information.

scala> object AnonymousFunction {
     |
     |   /**
     |    * Method to increment an integer by one.
     |    */
     |   anonymousIncrementFunction = ( x : Int ) => x + 1
     |
     |   /**
     |    * Main method
     |    * @param args application arguments
     |    */
     |   def main( args : Array[String] ) {
     |
     |     // Create an integer to test the anonymous function with
     |     myInteger = 0
     |
     |     println( myInteger ) // Prints: 0
     |
     |     myInteger = anonymousIncrementFunction( myInteger )
     |
     |     println( myInteger ) // Prints: 1
     |   }
     | }
<console>:12: error: not found: value anonymousIncrementFunction
         anonymousIncrementFunction = ( x : Int ) => x + 1
         ^
<console>:21: error: not found: value myInteger
           myInteger = 0
           ^
<console>:23: error: not found: value myInteger
           println( myInteger ) // Prints: 0
                    ^
<console>:25: error: not found: value myInteger
           myInteger = anonymousIncrementFunction( myInteger )
           ^
<console>:27: error: not found: value myInteger
           println( myInteger ) // Prints: 1
                    ^
`

are you telling me that this code compiles and runs on your machine without errors, or am i getting something wrong here???

i can change `var anonymousIncrementFunction` to `val anonymousIncrementFunction`, but ommiting it completely does not work. i cannot change `var myInteger = 0` to `val myInteger = 0`, since i get a 'reassignment to val' error in the line `myInteger = anonymousIncrementFunction( myInteger )`.

tutorials/tour/anonymous-function-syntax.md
conniec: You could do something like this to show that a previously incremented val will be incremented again:

```val myInteger = 0    

println(myInteger) // Prints: 0

val myIncrementedInteger = anonymousIncrementFunction(myInteger)

println(myIncrementedInteger) // Prints: 1

println(anonymousIncrementFunction(myIncrementedInteger)) // Prints: 2

println(anonymousIncrementFunction(anonymousIncrementFunction(myInteger)) // Prints: 2
```

It is more idiomatic to assign a result to a new val rather than reusing it.

SethTisue:
if we want to keep the exact structure of your original code, then we would need two `val`s, each with a different name.

but what about dodging the issue entirely and just doing:

```scala
val plusOne = (x: Int) => x + 1
println(plusOne(0))  // Prints: 1
```

this seems to me like the core of what the example is trying to convey, I doubt including any additional code besides this adds any value.

SethTisue:
oh by the way about getsatisfaction.com, sorry about that, see https://github.com/scala/scala.github.com/issues/645

SethTisue:
thank you!

PR #625, opened 2016-11-16
SIP-27: Trailing Commas, revised document

dwijnand:

PR #631, opened 2016-11-21
SIP-27: Trailing Commas, revised and moved document

dwijnand:
Builds on #625, adding the file move.

As a separate PR, as moving the file makes the diff near pointless.


dwijnand:
@jvican mergey mergey

PR #639, opened 2016-12-08
add a missing word to tour-of-scala.md

elmiko:
Looks like a "the" is missing from the first paragraph of the
"Scala is extensible" section of the tour-of-scala.md doc.


SethTisue:
thanks!

PR #641, opened 2016-12-09
add missing println statements to case-classes.md

elmiko:
The last example in the case-classes.md file shows what the example
output will be but the program snippet does not show any println
statements with the usage of the newly introduced case class function.
This change adds the println calls that use the new function and match
the output.


SethTisue:
thanks!

PR #650, opened 2017-01-09
`Queue` supports constant time `prepend`

Atry:


Atry:
Evidence: https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/collection/immutable/Queue.scala#L100

SethTisue:
shouldn't at least one of `prepend` and `append` be `eC`, not `C`, because of the occasional need to pivot?

SethTisue:
also, can you change the commit message to refer to `Queue` instead of `Stack`?

Atry:
Both `prepend` and `append` for `Queue` are `C`, not `eC`. See https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/collection/immutable/Queue.scala#L100 and https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/collection/immutable/Queue.scala#L114

SethTisue:
doh, you're right, it only pivots when you dequeue, an operation that isn't represented in the chart. sorry

PR #653, opened 2017-01-11
SIP-NN - Allow referring to other arguments in default parameters

pathikrit:


soronpo:
Quick note - give the SIP a proper title (leave it as SIP-NN until it receives an official number)

jvican:
Thanks for submitting this proposal @pathikrit! Will review in a bit. We're reviewing this at the end of this month, I'm preparing the SIP meeting in the next few days and will let you know the exact meeting date soon.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+```scala
+def substring(s: String, start: Int = 0)(end: Int = s.length): String
+```
+
+However, the above workaround is not always suitable in certain situations.
+
+The other more verbose alternative is by overloading:
+
+```scala
+def substring(s: String, start: Int = 0, end: Int = s.length): String
```
jvican: This is not working for me, I get `<console>:12: error: not found: value s` in the console.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+
+## History
+
+| Date          | Version          |
+|---------------|------------------|
+| Jan 11th 2017 | Initial Draft    |
+| Jan 12th 2017 | Initial Feedback |
+
+## Introduction
+Currently there is no way to refer to other arguments in the default parameters list:
```
jvican: Could you please add a motivation section that showcases why this would be useful?

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+```scala
+def substring(s: String, start: Int = 0, end: Int = s.length): String
+```
+
+The workaround to achieve this is by using a curried-function:
+```scala
+def substring(s: String, start: Int = 0)(end: Int = s.length): String
+```
+
+However, the above workaround is not always suitable in certain situations.
```
jvican: Why? Any examples or cases in which the previous transformation cannot be done? I get what you mean here, but explaining it a little bit more could be good for the upcoming review :smile:.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+#### Multiple Implicit Parameters
+[Multiple implicit parameters](https://github.com/scala/scala.github.com/pull/520) should also be allowed to refer to one another (left to right):
+```scala
+def codec[A](data: A)(implicit encoder: Encoder[A])(implicit decoder: Decoder[A] = encoder.reverse) // Legal
+```
+
+#### Referring to type members:
+The default parameters should be able to refer to type members of other arguments e.g.:
```
jvican: This section seems orthogonal to the target of this PR (this issue was raised by @olafurpg). I think that the good approach here would be to create another SIP that keeps track of this feature, since I would expect the implementation of this to be different to the one requested.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+
+substring(_, start = 0, end = 5)      // Legal
+substring(_, end = 5)                 // Legal (start = 0)
+substring(_, start = 5)               // Legal (same as s => substring(s, start = 5, end = s.length)
+```
+
+#### Multiple Implicit Parameters
+[Multiple implicit parameters](https://github.com/scala/scala.github.com/pull/520) should also be allowed to refer to one another (left to right):
+```scala
+def codec[A](data: A)(implicit encoder: Encoder[A])(implicit decoder: Decoder[A] = encoder.reverse) // Legal
```
xeno-by: Am I missing something, or this is not legal Scala syntax (multiple implicit parameter lists)?

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
+---
+layout: sip
+disqus: true
+title: SIP-NN - Allow referring to other arguments in default parameters
```
xeno-by: "Other arguments" or "previous arguments"?

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
```diff
@@ -0,0 +1,100 @@
...
+### Other languages
+The following languages allow referring to previously declared arguments in the function signature:
+* [CoffeeScript](http://coffeescript.org/)
+* [Kotlin](http://kotlinlang.org)
+* [TypeScript](https://www.typescriptlang.org/)
+
+AFAIK, there are no major languages where referring to parameters declared to the ***right*** is allowed.
+
+### Discussions
+[Scala Lang Forum](https://contributors.scala-lang.org/t/refer-to-previous-argument-in-default-argument-list/215/6)
```
xeno-by: I think it would be useful to have an implementation sketch to accompany this SIP. In my opinion, information about why this proposal hasn't been implemented yet and why it is feasible would help the potential discussion during one of the upcoming SIP meetings.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
dwijnand: Even if that worked you still can't have two methods with the same type signature.. This needs attention.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
jvican: I've been looking into this today and will submit an implementation sketch very soon /cc @pathikrit

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
jvican: Indeed. The trick here is scalac's generation of synthetic methods for the default args.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
jvican: I think that what @pathikrit means is that this change should also be valid for implicit parameter lists. That would marry well with this SIP.

jvican:
@pathikrit I've done some research into this proposal to shed some light into the way scalac encodes default args and how we can improve it according to what you suggest in this proposal. After some experimentation and an [implementation prototype](), I've decided to document my findings and explore different ways in which this feature can be implemented.

# Current implementation

Let's see a simple example of a method with defaults args:
```scala
def foo(a: Int, b: Int = 2, c: Int = 3): Int = a + b + c
```

The way scalac encodes this feature is by generating synthetic methods for every
default argument (see `Namers.scala`). Namers will find the method definition
`foo` and convert it into the equivalent of:

```scala
def foo(a: Int, b: Int = 2, c: Int = 3): Int = a + b + c
def foo$default$2() = 2
def foo$default$3() = 3
```

Therefore, when it finds a call site like `foo(1)`, `NamesDefaults.scala` will transform it into:
```scala
foo(1, foo$default$2(), foo$default$3())
```

As you see, the synthetic methods created don't take parameters. This is what
currently constraints dependencies between parameters defined in the same
parameter list.

What happens if we define the dependencies between curried parameters lists though?

```scala
def foo(a: Int)(b: Int = a, c: Int = a): Int = a + b + c
def foo$default$2(a: Int) = 2
def foo$default$3(a: Int) = 3
```

is transformed into:

```scala
foo(1, foo$default$2(1), foo$default$2(1))
```

In this case, scalac generates synthetic methods that take `a` as parameter, as it's the only parameter in previous parameter lists. This is done even if the default implementation of `b` and `c` does not depend on `a`. Otherwise, changing the default arguments' implementation will change the signature of the synthetic methods and break binary compatibility.

Note that these are simplified examples of default args. The current implementation takes care of more details, for example it copies type parameters of `foo` over the synthetic methods. This usually happens for both higher-kinded functions and constructors.
  
How can we implement the changes proposed in this proposal?

# Exploring other encodings
  
## First attempt
Our goal with the new encoding is to allow dependencies between parameters defined in the same parameter list, both on the right and on the left. The following method would be allowed:
  
```scala
// definition
def foo(a: Int, b: Int = c, c: Int = a): Int = a + b + c
// invocation
foo(a)
```
  
Let's see if we can find a way to implement it. First, we start with the same encoding used by dependencies in curried parameter lists and modify it to be more flexible. Let's create synthetic methods that define all the parameters that a given default argument could depend on. We'll get the following synthetic methods:
```scala
def foo$default$2(a: Int, c: Int) = c // default implementation of b
def foo$default$3(a: Int, b: Int) = a // default implementation of c
```
  
As `b` can theoretically depend on `a` and `c`, we need to express that potential dependency in the synthetic method definition, and the same for the synthetic method for `c`. However, this encoding produces a recursive dependency between `b` and `c`. To compute the default value of `b`, we need to invoke `foo$default$2` with an `a` and `c`, but `c` needs to be computed as well and it recursively depends on `b`. Unfortunately, there is no way we can generate code for the call `foo(a)`.
  
## Second attempt
The previous encoding didn't work because of recursive dependencies in the default args. Why are we generating synthetic methods that could express all the dependencies of an argument? Let's just do it for the parameters that the default arguments actually depend on.
```scala
def foo$default$2(c: Int) = c // default implementation of b
def foo$default$3(a: Int) = a // default implementation of c
```
  
Now that we have removed the circular dependency, we can generate code for the call site:
```scala
foo(a, foo$default$2(foo$default$3(a)), foo$default$3(a))
```
  
However, this encoding is worse than we thought. We are encoding the parameter dependency in the signature of the synthetic methods, which means that if we change the implementation of `b` or `c`, we'll also change the binary code. This will break users' binaries because the types for the default invocations won't match. This encoding, while feasible, it's too strict for users to use.
  
Can we do better?
  
# Third and last attempt
At this point, we can only constrain the scope of this proposal to make the default args implementation feasible. Another way to remove the circular dependency that we created in the first encoding attempt is to constrain the parameter dependencies. Instead of allowing any parameter dependency, we will only allow dependencies for:
1. parameters of previous parameter lists
2. parameters defined on the left of a default arg
  
Therefore, the method we started with is not valid anymore -- `b` cannot depend on `c`. Let's redefine `foo`:
```scala
// definition
def foo(a: Int, b: Int = a, c: Int = a + b): Int = a + b + c
// invocation
foo(a)
```
With our new limitation, these are the generated synthetic methods:
```scala
def foo$default$2(a: Int) = a // default implementation of b
def foo$default$3(a: Int, b: Int) = a + b // default implementation of c
```
A similar transformation applies to curried methods `def foo(a: Int)(b: Int = a, c: Int = a +b)`.
  
Fortunately, our new scheme allow us to:
1. Keep binary compatibility no matter what the implementation of `b` or `c` is.
2. Allow implementations of default args to depend on parameters defined on the left.
  
I think that binary compatibility is utterly important for Scala. I suggest that we don't allow default implementations to depend on arguments defined both on the left and the right, only those on the left. This is closer to what we have right now.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: other

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: See the link to the PR

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: Working on it...

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: Done.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: Oops, a typo. Fixed.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: I don't know much about scalac internals to comment. Maybe it is very similar (then keep 1 SIP) and maybe it isn't (2 SIPs). Can someone more familiar with scalac comment?

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
pathikrit: Maybe previous I guess if other is not feasible (circle dependencies)?

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
jvican: I've been looking into the implementation details of this and I can confirm that they are orthogonal issues. Concretely, referring to type members is tricky. I think it can be done but the feature interaction is scary. Could someone confirm? Maybe @xeno-by || @sjrd?

dwijnand: I think they're similar in spirit, but I'm sure they're distinct in the implementation.

To avoid a situation where the committee pushes you back a month to go look into and document this, I would keep it in. At the very least they can ask you to split it out - you can then drop it or pursue it as a separate SIP.

sips/pending/_posts/2017-01-11-refer-other-arguments-in-args.md
jvican: I would actually write "Allow references to left arguments of the same list in default parameters" if @pathikrit agrees with my previous analysis that doing it for all the arguments is unfeasible. :smile:

jvican:
FYI @pathikrit: @sjrd will be the reviewer for this proposal. You can talk to him or me for any doubt you have along the process. He will be the one giving you feedback and leading the discussions in our next meeting.

soronpo:
@pathikrit I suggest you checkout how this proposal [looks on the site](http://docs.scala-lang.org/sips/pending/refer-other-arguments-in-args.html), and fix the annotations. I recommend that you build the site locally and view the HTML version of your proposal. The MD to HTML translation is different than what is viewed by an MD editor. See #689

gvoicesammy-max (commented):
SIP-NN

Issue #607, opened 2016-10-21
Jekyll/Liquid generation is dead-slow because of tutorial ToC


dragos:
All the time is spent in several nested loops computing the table of contents for tutorials. Unfortunately Liquid is a pretty poor language so there aren't any obvious workarounds, but it's definitely worth investigating. Once I commented those loops times went down to around 1s. 

Maybe a solution using categories or tags would reduce the number of iterations (350 pages, but only 33 are part of the Scala tour).

```
$ bundle exec jekyll serve --profile --incremental
Configuration file: /Users/dragos/workspace/git/scala.github.com/_config.yml
            Source: /Users/dragos/workspace/git/scala.github.com
       Destination: /Users/dragos/workspace/git/scala.github.com/_site
 Incremental build: enabled
      Generating... 
    Liquid Warning: Liquid syntax error (line 164): Unexpected character { in "{{{` and `}}" in overviews/scaladoc/for-library-authors.md

Filename                                                           | Count |     Bytes |    Time
-------------------------------------------------------------------+-------+-----------+--------
_layouts/tutorial.html                                             |   132 | 20395.38K | 202.110
_includes/tutorial-tour-list.txt                                   |   133 | 18740.42K | 197.912
_includes/tutorial-toc.txt                                         |   132 | 18618.52K | 196.613
_layouts/overview-large.html                                       |   164 | 24041.61K | 151.205
_includes/toc-large.txt                                            |   164 | 20639.72K | 144.796
tutorials/index.md                                                 |     1 |   184.98K |   1.759
_layouts/overview.html                                             |    36 |  1177.23K |   1.444
es/overviews/index.md                                              |     1 |    41.87K |   0.729
_includes/localized-overview-index.txt                             |     1 |    41.71K |   0.728
_includes/header.txt                                               |   369 |  1777.55K |   0.456
```

jarrodwb:
I am not sure if speed is really something that need to optimized for. It is just an annoyance.

What is your motivation for working on it?

SethTisue:
@jarrdwb getting the site build down time from minutes to seconds seems like a pretty sweet improvement to me!

jarrodwb:
Okay you convinced me. ;)

SethTisue:
at https://github.com/scala/scala-lang/issues/493#issuecomment-254475807 Simon seemed to be saying this was fixed in his fork, so it might be worth looking there and trying to find what the relevant change was

heathermiller:
[...]

PR #654, opened 2017-01-13
`@stableABI` SIP.

DarkDimius:
TLDR:
This is intended to be @tailrec - like annotation that will fail compilation in case emitted methods or their signatures are different from those written by users. In order for @stableABI compatible class to succeed compilation there should be: no methods synthesized and no signatures changed.

The intended use-case is providing binary compatibility across major Scala versions by ensuring that a class or a trait `is compiled as written by developer`.

UPD: changed name of the annotation and the SIP.

### DIFF
--- a/sips/pending/_posts/2017-01-13-binary-compatibility.md
+++ b/sips/pending/_posts/2017-01-13-binary-compatibility.md
@@ -1,0 +1,152 @@
+---
+layout: sip
+title: SIP XX - Improving binary compatibility with @binaryCompatible
+disqus: true
+---
+
+__Dmitry Petrashko__
+
+__first submitted 13 February 2017__
+
+## Introduction ##
+
+Scala is a language which evolves fast and thus made a decision to only promise binary compatibility across minor releases.
+At the same time, there is a demand to develop APIs that live longer than a major release cycle of Scala.
+This SIP introduces an annotation `@binaryCompatible` that checks that `what you write is what you get`.
+It will fail compilation in case emitted methods or their signatures 
+are different from those written by users. 
+As long as signatures of methods in source is not changed, `@binaryCompatible` annotated class 
+will be compatible across major version of Scala. 
+
+## Use Cases
+Dotty currently uses java defined interfaces as public API for SBT in order to ensure binary compatibility. 
+These interfaces can be replaced by `@binaryCompatible` annotated traits to reach the same goal.  
+
+## Design Guidelines
+`@binaryCompatible` is a feature which is supposed to be used by a small subset of the ecosystem to be binary compatible across major versions of Scala.
+Thus this is designed as an advanced feature that is used rarely and thus is intentionally verbose. 
+It's designed to provide strong guarantees, in some cases sacrificing ease of use.
+ 
+The limitations enforced by `@binaryCompatible` are designed to be an overapproximation: 
+instead of permitting a list of features known to be compatible, `@binaryCompatible` enforces a stronger 
+check which is sufficient to promise binary compatibility. 
+
+## Overview ##
+In order for a class or a trait to succeed compilation with the `@binaryCompatible` annotation it has to be:
+  - defined on the top level;
+  - use a subset of Scala that during compilation does not require changes to public API of the class, including
+     - synthesizing new members, either concrete or abstract;
+     - changing binary signatures of existing members, either concrete or abstract;
+
+`@binaryCompatible` does not change the compilation scheme of a class:
+ compiling a class previously annotated with the `@binaryCompatible`, will produce the same bytecode with or without `@binaryCompatible` annotation. 
+
+Below are several examples of classes and traits that succeed compilation with `@binaryCompatible`
+```scala
+{% highlight scala %}
+@binaryCompatible
+trait AbstractFile {
+  def name(): String
+
+  def path(): String
+
+  def jfile(): Optional[File]
+}
+
+@binaryCompatible
+trait SourceFile extends AbstractFile {
+  /** @return The content of this file as seen by the compiler. */
+  def content(): Array[Char]
+}
+
+@binaryCompatible
+trait Diagnostic {
+  def message(): String
+
+  def level(): Int
+
+  def position(): Optional[SourcePosition]
+}
+
+@binaryCompatible
+object Diagnostic {
+  @static final val ERROR: Int = 2
+  @static final val WARNING: Int = 1
+  @static final val INFO: Int = 0
+}
+
+{% endhighlight %}
+```
+
+## Features that will fail compilation with `@binaryCompatible`
+The features listed below have complex encodings that may change in future versions. We prefer not to compromise on them.
+Most of those features can be simulated in a binary compatible way by writing a verbose re-impelemtation 
+which won't rely on desugaring performed inside compiler.
+
+  - public fields. Can be simulated by explicitly defining public getters and setters that access a private field;
+  - lazy vals. Can be simulated by explicitly writing an implementation in source;
+  - case classes. Can be simulated by explicitly defining getters and other members synthesized for a case class(`copy`, `productArity`, `apply`, `unApply`, `unapply`).
+
+The features listed below cannot be easily re-implemented in a class or trait annotated with `@binaryCompatible`.
+  - default arguments;
+  - default methods. See Addendum;
+  - constant types(both explicit and inferred);
+  - inline.
+    
+## `@binaryCompatible` and Scala.js
+
+Allowing to write API-defining classes in Scala instead of Java will allow them to compile with Scala.js, 
+which would have benefit of sharing the same source for two ecosystems.
+
+Scala.js currently is binary compatible as long as original bytecode compiled by Scala JVM is binary compatible.
+Providing stronger binary compatibility guarantees for JVM will automatically provide stronger guarantees for Scala.js.
+ 
+
+## Comparison with MiMa ##
+The Migration Manager for Scala (MiMa in short) is a tool for diagnosing binary incompatibilities for Scala libraries.
+MiMa allows to compare binary APIs of two already compiled classfiles and reports errors if APIs do not match perfectly.
+
+MiMa and `@binaryCompatible` complement each other, as `@binaryCompatible` helps to develop APIs that stay compatible 
+across major versions, while MiMa checks that previously published artifacts indeed have the same API.
+
+`@binaryCompatible` does not compare the currently compiled class or trait against previous version, 
+so introduction of new members won't be prohibited. This is a use-case for MiMa.
+  
+MiMa does not indicate how hard, if possible, would it be to maintain compatibility of a class across future versions of Scala.
+Multiple features of Scala, most notably lazy vals and traits, has been compiled diffently by different Scala versions
+making porting existing compiled bytecode across versions very hard. 
+MiMa will complain retroactively that the new version is incompatible with the old one. 
+`@binaryCompatible` will instead indicate at compile time that the old version had used features whose encoding is prone to change.
+This provides early guidance and warning when designing long-living APIs before they are publicly released.  
+
+## Compilation scheme ##
+No modification of typer or any existing phase is planned. The current proposed scheme introduces a late phase that runs before the very bytecode emission that checks that:
+ - classes and traits annotated as  `@binaryCompatible` are on the top level; 
+ - no non-private members where introduced inside classes and traits annotated as  `@binaryCompatible` by compiler using phase travel;
+ - no non-private members inside classes and traits annotated as  `@binaryCompatible` has changed their signature from the one written by developer.
+
+The current prototype is implemented for Dotty and supports everything descibed in this SIP. 
+The implementation is simple with less than 50 lines of non-boilerplate code. 
+The current implementation has a scope for improvement of error messages that will report domain specific details for disallowed features, but it already prohibits them.
+
+## Addendum: Default methods ##
+By `default methods` we mean non-abstract methods defined and implemented by a trait.
+
+The way how those methods are implemented by compiler has changed substantially over years.
+At the same time, `invokeinterface` has always been a reliable way to invoke such a method, 
+independently from how it was implemented under the hood.
+
+One might reason that, as there has been a reliable way to call methods on the binary level, 
+it should be allowed to use them in binary compatible APIs.
+ 
+At the same time, the mixin composition protocol that is followed when a class inherits those traits has also 
+changed substantially. 
+The classes which have been correctly inheriting those traits compiled by previous versions of Scala
+may need recompilation if trait has been recompiled with a new major version of Scala. 
+ 
+Thus, the authors of this SIP has decided not to allow default methods in the 
+`@binaryCompatible` traits.
+
+## See Also ##
+ * [dotty#1900](https://github.com/lampepfl/dotty/pull/1900) is an implementation for Dotty
+ * [MiMa](https://github.com/typesafehub/migration-manager) 
--- a/sips/pending/_posts/2017-01-13-binary-compatibility.md
+++ b/sips/pending/_posts/2017-01-13-binary-compatibility.md
@@ -6,7 +6,7 @@ disqus: true
 
 __Dmitry Petrashko__
 
-__first submitted 13 February 2017__
+__first submitted 13 January 2017__
 
 ## Introduction ##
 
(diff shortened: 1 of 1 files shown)


sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,152 @@
+---
+layout: sip
+title: SIP XX - Improving binary compatibility with @binaryCompatible
+disqus: true
+---
+
+__Dmitry Petrashko__
+
+__first submitted 13 February 2017__
```
heathermiller: It's currently January, not February yet ;-)

densh:
This seems to be a replacement for java-defined interfaces, why not name it `@interface` to make crystal clear what it means? Binary compatibility is quite a broad term.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,152 @@
...
+Scala is a language which evolves fast and thus made a decision to only promise binary compatibility across minor releases.
+At the same time, there is a demand to develop APIs that live longer than a major release cycle of Scala.
+This SIP introduces an annotation `@binaryCompatible` that checks that `what you write is what you get`.
+It will fail compilation in case emitted methods or their signatures 
+are different from those written by users. 
+As long as signatures of methods in source is not changed, `@binaryCompatible` annotated class 
+will be compatible across major version of Scala. 
+
+## Use Cases
+Dotty currently uses java defined interfaces as public API for SBT in order to ensure binary compatibility.
```
smarter: Actually `dotty-interfaces` is not used by sbt at all. `dotty-sbt-bridge` depends on dotty itself, and sbt will automatically compile it from source against the current version of `dotty` (this is nothing new, the same thing happens with the scalac sbt bridge).
However, `dotty-interfaces` is used by IntelliJ IDEA.

DarkDimius:
@densh, it does not only apply to interfaces. It also applies to classes.

Given that there's an assumption that this feature is rarely used and only used by advanced users, I'd prefer not to give it a common name such as `@interface`.

DarkDimius:
@milessabin thanks for your comment.
>FWIW, it seems that the criteria for @binaryCompatible more or less correspond to those for a direct unencoded mapping to Java language artefacts, so maybe @java would be a better name?

Inside the bodies of methods one can use all the features of Scala, as they won't be seen in the binary API observed from outside. I think I should be explicit about it and include examples that illustrate this. Thank you for pointing this out.

travisbrown:
:+1:, but also :+1: to reconsidering the name. In the cases where I've personally wanted something like this, the goal has been ease of Java interop in a multi-language codebase, not primarily _binary compatibility_ as I think most people would understand it.

dragos:
I think this would be a very useful, but I agree with previous commenters that the name should be more specific. Something that suggests that only platform signatures are being used.

aalleexxeeii:
Do you mean absolutely any synthetic construct? For instance, should bit-fields for lazy values be considered as those? Or stuff like `MODULE$`?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+disqus: true
+---
+
+__Dmitry Petrashko__
+
+__first submitted 13 January 2017__
+
+## Introduction ##
+
+Scala is a language which evolves fast and thus made a decision to only promise binary compatibility across minor releases.
```
xeno-by: I think it would also be useful to define what binary compatibility is.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+At the same time, there is a demand to develop APIs that live longer than a major release cycle of Scala.
+This SIP introduces an annotation `@binaryCompatible` that checks that `what you write is what you get`.
```
xeno-by: I agree with other commenters that a different name would be beneficial. To me, saying that something is binary compatible, immediately raises the question "with what?". This annotation doesn't answer that question.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+will be compatible across major version of Scala. 
+
+## Use Cases
+In case there's a need to develop an API that will be used by clients compiled using different major versions of Scala, 
+the current approach is to either develop them in Java or to use best guess to restrict what Scala features should be used.
+There's also a different approach which is used by SBT: instead of publishing a binary `compiler-interface`, sources are published instead 
+that would be locally compiled.
+
+There's also a use-case of defining a class which is supposed to be also used from Java. 
+`@binaryCompatible` will ensure that there are no not-expected methods that would show up in members of a class or an interface.
```
xeno-by: s/not-expected/unexpected/

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+
+Scala is a language which evolves fast and thus made a decision to only promise binary compatibility across minor releases.
+At the same time, there is a demand to develop APIs that live longer than a major release cycle of Scala.
+This SIP introduces an annotation `@binaryCompatible` that checks that `what you write is what you get`.
+It will fail compilation in case emitted methods or their signatures 
+are different from those written by users. 
+As long as signatures of methods in source is not changed, `@binaryCompatible` annotated class 
+will be compatible across major version of Scala. 
+
+## Use Cases
```
xeno-by: Can we number these use cases, i.e. have this section look as follows: 

```
1) Use case 1 
<explanation>

2) Use case 2
<explanation>

etc
```

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+Thus this is designed as an advanced feature that is used rarely and thus is intentionally verbose. 
+It's designed to provide strong guarantees, in some cases sacrificing ease of use.
+ 
+The limitations enforced by `@binaryCompatible` are designed to be an overapproximation: 
+instead of permitting a list of features known to be compatible, `@binaryCompatible` enforces a stronger 
+check which is sufficient to promise binary compatibility. 
+
+## Overview ##
+In order for a class or a trait to succeed compilation with the `@binaryCompatible` annotation it has to be:
+  - defined on the top level;
```
xeno-by: Why?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+  - use a subset of Scala that during compilation does not require changes to public API of the class, including
+     - synthesizing new members, either concrete or abstract;
+     - changing binary signatures of existing members, either concrete or abstract;
```
xeno-by: What about public API of class's companion?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+`@binaryCompatible` is a feature which is supposed to be used by a small subset of the ecosystem to be binary compatible across major versions of Scala.
+Thus this is designed as an advanced feature that is used rarely and thus is intentionally verbose. 
+It's designed to provide strong guarantees, in some cases sacrificing ease of use.
+ 
+The limitations enforced by `@binaryCompatible` are designed to be an overapproximation: 
+instead of permitting a list of features known to be compatible, `@binaryCompatible` enforces a stronger 
+check which is sufficient to promise binary compatibility. 
+
+## Overview ##
+In order for a class or a trait to succeed compilation with the `@binaryCompatible` annotation it has to be:
```
xeno-by: What about objects?

sips/pending/_posts/2017-01-13-binary-compatibility.md
xeno-by: What's a binary signature?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+  - defined on the top level;
+  - use a subset of Scala that during compilation does not require changes to public API of the class, including
```
xeno-by: What's a public API?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+    lazy val result = 0 // while lazy vals are prohibited in the class, they are allowed in the bodies of methods
+    result
+  }
+}
+{% endhighlight %}
+```
+
+## Features that will fail compilation with `@binaryCompatible`
+The features listed below have complex encodings that may change in future versions. We prefer not to compromise on them.
+Most of those features can be simulated in a binary compatible way by writing a verbose re-impelemtation
```
xeno-by: Typo.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+which won't rely on desugaring performed inside compiler.
+Note that while those features are prohibited in the public API, they can be safely used inside bodies of the methods.
+
+  - public fields. Can be simulated by explicitly defining public getters and setters that access a private field;
+  - lazy vals. Can be simulated by explicitly writing an implementation in source;
+  - case classes. Can be simulated by explicitly defining getters and other members synthesized for a case class(`copy`, `productArity`, `apply`, `unApply`, `unapply`).
+
+The features listed below cannot be easily re-implemented in a class or trait annotated with `@binaryCompatible`.
+  - default arguments;
+  - default methods. See Addendum;
```
xeno-by: What's a default method? There's no such definition in SLS.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+  - constant types(both explicit and inferred);
+  - inline.
```
xeno-by: Why exclude inline?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+    
+## `@binaryCompatible` and Scala.js
+
+Allowing to write API-defining classes in Scala instead of Java will allow them to compile with Scala.js, 
+which would have benefit of sharing the same source for two ecosystems.
+
+Scala.js currently is binary compatible as long as original bytecode compiled by Scala JVM is binary compatible.
```
xeno-by: Binary compatible with what?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+Multiple features of Scala, most notably lazy vals and traits, has been compiled diffently by different Scala versions
+making porting existing compiled bytecode across versions very hard. 
+MiMa will complain retroactively that the new version is incompatible with the old one. 
+`@binaryCompatible` will instead indicate at compile time that the old version had used features whose encoding is prone to change.
+This provides early guidance and warning when designing long-living APIs before they are publicly released.  
+
+## Compilation scheme ##
+No modification of typer or any existing phase is planned. The current proposed scheme introduces a late phase that runs before the very bytecode emission that checks that:
+ - classes and traits annotated as  `@binaryCompatible` are on the top level; 
+ - no non-private members where introduced inside classes and traits annotated as  `@binaryCompatible` by compiler using phase travel;
```
xeno-by: What's phase travel?

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+ - no non-private members inside classes and traits annotated as  `@binaryCompatible` has changed their signature from the one written by developer.
```
xeno-by: What does this mean? Does erasure of type arguments count as a change, for instance?

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: I'll add a link defining both terms. Thank you.

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: While I agree that the `@binaryCompatible` name isn't perfect, I didn't yet find a better one. 
As this annotation is intended to be rarely used and only by advanced users, I don't think we should concentrate that much on the name, but I welcome name suggestions.

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+
+__first submitted 13 January 2017__
+
+## Introduction ##
+
+Scala is a language which evolves fast and thus made a decision to only promise binary compatibility across minor releases.
+At the same time, there is a demand to develop APIs that live longer than a major release cycle of Scala.
+This SIP introduces an annotation `@binaryCompatible` that checks that `what you write is what you get`.
+It will fail compilation in case emitted methods or their signatures 
+are different from those written by users.
```
xeno-by: s/users/authors/

sips/pending/_posts/2017-01-13-binary-compatibility.md
```diff
@@ -0,0 +1,168 @@
...
+layout: sip
+title: SIP XX - Improving binary compatibility with @binaryCompatible
+disqus: true
+---
+
+__Dmitry Petrashko__
+
+__first submitted 13 January 2017__
+
+## Introduction ##
```
xeno-by: I think that the introduction would benefit from the definitions of the following terms used later in the document:
1) Binary compatibility (both the intransitive and transitive variations used below)
2) Public API
3) Binary signature
4) The notion of signatures being different from those written by their authors.

I'd suggest having a small example that introduces and illustrates these notions.

DarkDimius:
@xeno-by, thank you for your through review. 

As most questions are asking about definitions of terms, I think I'll introduce a new section that defines terms used in this SIP. It's indeed harder to follow than most sips as current writing assumes knowledge of both compiler internals and JVM spec.

DarkDimius:
Just had a Dotty meeting, which was also joined by @retronym.
It seems that everyone present on the meeting has converged on the name `@stableABI`.

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: > Why limit this to just top-level classes? 

1. Allowing to define `@stableABI` inner classes will most likely fail compilation due to second reason, as they will get successors synthesized to access the outer class;
2. there are several subtleties in how outer class can affect compilation of inner class, eg by converting private method to public;
3. use-case of an inner class can be easily handled by defining a top-level supper class and making inner class extend it;

In order to have a more user-friendly behavior than 1 & 2, and given presence of a simple workaround, I've decided to prohibit it.

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: added.

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: The idea is that `@stableABI` classes don't need `@scalaSignature` or `@tasty` to compile against. In order for inline to work, we'll need to be able to read the trees from those sections and make assumptions about those trees, that can be dependent on compiler version. 

I propose to go conservative route here, we'll have an opportunity to allow more features later, but we won't have an opportunity to remove them.

sips/pending/_posts/2017-01-13-binary-compatibility.md
DarkDimius: I don't think `Compilation scheme` section needs explanation of common terms used inside both compilers.

olafurpg:
FWIW. With `@stableABI`, I imagine it might become popular to release small libraries without cross compilation. A hypothetic example, if Function and Tuple were to get stable ABIs (don't think they're included in the current draft of the SIP), libraries such as [geny](https://github.com/lihaoyi/geny/blob/master/geny/shared/src/main/scala/geny/Generator.scala) could take advantage of this.

DarkDimius:
The CI seems to fail, without reporting any errors.
@heathermiller @jvican, could you please help with getting this merged?
Afaik it should be discussed tomorrow on the SIP meeting, so it would be nice to get it in.

heathermiller:
It won't be displayed even if the CI succeeds. The GitHub pages builds are also timing out and failing. So until we can move this site to chara/alaska, or fix #607, this page can't be displayed on the site.

DarkDimius:
@heathermiller, ouch. Thanks for having a look.

jvican:
Hey @DarkDimius, a fix is on the way https://github.com/scala/scala.github.com/pull/684. When we merge it, can you rebase your changes so that Travis doesn't time out?

jvican:

DarkDimius:
@jvican thank you for the ping. Rebased.

xeno-by:
I'm a bit confused about the current name for the annotation. Is it `@stableABI` or `@binaryCompatible`? From what I can see, the PR title has been updated, but the SIP text hasn't.

DarkDimius:
@xeno-by thanks for pinging. Apperantly when fixing the repo we've rebased and merged the wrong version. https://github.com/scala/scala.github.com/pull/691 is the up-to-date one and the one that was reviewed in this PR.

PR #655, opened 2017-01-13
Enabled pt-br lang for Scala Tour

fforbeck:
- Added the pt-br in the languages list 
- Moved the lang list from `abstract-types.md` to `tour-of-scala.md`

@heathermiller I put the lang list in the first page of Scala Tour tutorial. I believe it is easier to update there (for future translations). What do you think?


heathermiller:
Yep, I agree that it's easier there. Thank you @fforbeck!!

PR #658, opened 2017-01-19
SIP25: answer feedback & add agreements with Scala.js

DarkDimius:
Feedback was: 
 - answer if @static could this be @tailrec-like annotation;
 - clarify if and how @static affects initialization order;
 - describe if @static affects binary compatibility.

Addition to the SIP was an update of situation with Scala.js

### DIFF
--- a/sips/pending/_posts/2016-01-11-static-members.md
+++ b/sips/pending/_posts/2016-01-11-static-members.md
@@ -115,6 +115,8 @@ The following rules ensure that methods can be correctly compiled into static me
 
 6. Only `@static` methods and vals are supported in companions of traits. Java8 supports those, but not vars, and JavaScript does not have interfaces at all.
 
+Note that because of platform requirements for JavaScript interop, rules `3` and `4` would be lifted for objects that have a companion class that inherits `js.Any`.
+
 ## Compilation scheme ##
 
 No modification of the typer is planned. The current proposed scheme piggybacks on already existing scoping restrictions in the typer, thus requiring `@static` methods to be defined in `object`s.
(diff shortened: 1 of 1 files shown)


lrytz:
Not directly related to the changes in this PR, but the proposal in general. Normally, superclass fields are initialized before the fields of a subclass

```
scala> class C { val x = {println("x"); 1 } }
defined class C

scala> object O extends C { val y = { println("y"); 2 } }
defined object O

scala> O
x
y
res0: O.type = O$@49293b43
```

If we make `y` static in the above, it looks like that would change initialization order (`y` before `x`), no?

DarkDimius:
@lrytz, thanks for a good question. I indeed need to clarify that the rules enforce that there's no way to observe changes in initialization _within the same class_, which was somehow lost while editing and proofreading.

Indeed, static fields can be initialized earlier than they used to be initialized, but not later and by defining them first in the same object, we guarantee that you can't observe the changes in initialization withing the same object.

lrytz:
OK, it would be good if the proposal would explicitly show / discuss that semantic change, so that reviewers are aware.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -132,12 +134,53 @@ Java allows classes to define static methods with the same name and signature as
 This is required because in Java calling a `static` method on a class instance is supported.
 This proposal does not need to introduce this notion as we do not support such calls.
 
+## ScalaJS and @JSstatic ##
+As ScalaJS needs this feature fast, a decision has been made to ship it under a name of `@JSStatic` before waiting for this SIP to be accepted and implemented in `scalac`. When this SIP is accepted and implemented in `scalac` the `@JSStatic` would become a deprecated type alias to `scala.static`.
```
sjrd: ScalaJS -> Scala.js
`@JSstatic` -> `@JSStatic`

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -132,12 +134,53 @@ Java allows classes to define static methods with the same name and signature as
...
+
+## Scala.Native ##
+
```
sjrd: Is this empty section intended?

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -134,10 +134,8 @@ Java allows classes to define static methods with the same name and signature as
 This is required because in Java calling a `static` method on a class instance is supported.
 This proposal does not need to introduce this notion as we do not support such calls.
 
-## ScalaJS and @JSstatic ##
-As ScalaJS needs this feature fast, a decision has been made to ship it under a name of `@JSStatic` before waiting for this SIP to be accepted and implemented in `scalac`. When this SIP is accepted and implemented in `scalac` the `@JSStatic` would become a deprecated type alias to `scala.static`.
-
-## Scala.Native ##
+## Scala.js and @JSStatic ##
+As Scala.js needs this feature fast, a decision has been made to ship it under a name of `@JSStatic` before waiting for this SIP to be accepted and implemented in `scalac`. When this SIP is accepted and implemented in `scalac` the `@JSStatic` would become a deprecated type alias to `scala.static`.
```
jvican: :+1:

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -149,16 +147,47 @@ It also does not address the question of `@static` members in inner objects and
...
+  @static val y = { println("y"); 2 }
+   // prints:
+   // y
+   // x
+}
+{% endhighlight %}
+```
 
-The `@static` proposal is similar to `@tailrec` in a sense that it fails compilation in the case where the user did not write code that follows the aforementioned rules. These rules exist to enforce the unlikelyhood of an observable difference in semantics if `@static` annotations are dropped; without resorting to code which either uses `Unsafe` or exhibits undefined behaviour under the JVM.
+Static fields can be initialized earlier than they used to be initialized while being non-static, but never later.
```
jvican: I think this is a good trade-off.

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -132,12 +134,82 @@ Java allows classes to define static methods with the same name and signature as
...
+}
+{% endhighlight %}
+```
+Let's consider possible options:
+ - if the field `c` is emitted as `static` on the bytecode level, it will be initialized before the `c` in superclass is initialized, reordering side-effects in initializers;
+ - if the field `c` is emitted as `static` on the bytecode level it will use a different storage that the memory layout defined by `Super`, leading to potential bugs and wasted memory;
+ - if the field `c` is _not_ emitted as `static` but the field `d` is, then the order of initialization would also be affected, reordering side-effects.
+ 
+Based on the previous study done in preparation for this SIP, the authors believe that the only reasonable way to maintain current sematics would be to say that such alternative would require these rules:
+ - only the fields which were not declared by parents of the object can be emitted as static;
```
jvican: I think you need an space here for Jekyll to show it as an enumeration, the same for line 203 of this diff, in "let's consider possible options".

jvican:
Actually, one last thing: can you change the status of this proposal (in the jekyll header, line 6) from "under revision" to "under review"?

DarkDimius:
@jvican done

sips/pending/_posts/2016-01-11-static-members.md
```diff
@@ -132,12 +134,84 @@ Java allows classes to define static methods with the same name and signature as
...
+object Object extends Super {
+ override val c = {println(2); 2}
+ val d = {println(3); 2}
+}
+{% endhighlight %}
+```
+Let's consider possible options:
+
+ - if the field `c` is emitted as `static` on the bytecode level, it will be initialized before the `c` in superclass is initialized, reordering side-effects in initializers;
+ - if the field `c` is emitted as `static` on the bytecode level it will use a different storage that the memory layout defined by `Super`, leading to potential bugs and wasted memory;
```
sjrd: I think this bullet is irrelevant. If `c` is not emitted as `static` on the bytecode level, it gets a difference storage *anyway*: there is one field in `Super` initialized with `{println(1); 1}` (then never read), and one field in `Object` initialized with `{println(2): 2}` (and accessed through the generated getter `c()`).

DarkDimius: Thanks for spotting this. I forgot that we're not in the linker here...
Indeed, the bullet should be removed.

DarkDimius:
> The CI seems to fail, without reporting any errors.
> @heathermiller @jvican, could you please help with getting this merged?
> Afaik it should be discussed tomorrow on the SIP meeting, so it would be nice to get it in.
@heathermiller @jvican, could you please help with getting this merged?

jvican:

DarkDimius:
@jvican thank you for the ping. Rebased.

PR #660, opened 2017-01-22
replace some words with more specific works

openmaya:
콘크리트로 번역된 부분들이 사실은 "특정", "구체적인", "지정" 등의 의미로 번역되면 좀더 자연스러워 고쳤습니다.


heathermiller:
Thank you @openmaya! I will just trust that these fixes are correct. :-) Thanks again.

openmaya:
Thanks. I wrote a discussion message below the document page but no one answered. 
I think Korean version uses old code samples, I will update all samples.

PR #661, opened 2017-01-22
update recent version example In Korean version, add more Korean translation comment

openmaya:
Because it didn' make sense for me, I modified some word.


openmaya:
update recent version example, add more Korean translation comment

heathermiller:
👏 🎉 Thank you!

PR #666, opened 2017-01-26
refined some expressions more natually

openmaya:


heathermiller:
👏👍 Thank you @openmaya!!

PR #670, opened 2017-02-01
Make unified-types.md more intuitive to new comers

shogowada:
- It simplifies wordings by removing unnecessary and redundant information.
- It says `scala.AnyVal` is non-nullable, which I think is a critical information to communicate.
- It removes `scala.ScalaObject`.
    - I still don't know exactly what it is, but there is [a comment](http://docs.scala-lang.org/tutorials/tour/unified-types#comment-2685040354) saying it is deprecated. I thought it's no harm to take it out of the picture, but correct me if it's still important.
- In the example, it uses `List[Any]` instead of `scala.collection.mutable.LinkedHashSet[Any]`.
    - I think more people are familiar with list than linked hash set.
- In the example, it uses anonymous function instead of partially applied main function.
    - I think using partially applied function here throws out new comers, while I expect many of them are already familiar with anonymous functions.
- In the example, it uses `foreach` instead of `iterator`.
- It removes use of `object` or `App` because they are unrelated.

### DIFF
--- a/tutorials/tour/unified-types.md
+++ b/tutorials/tour/unified-types.md
@@ -36,7 +36,7 @@ val list: List[Any] = List(
 list.foreach(element => println(element))
 ````
 
-The application defines a variable `list` of type `List[Any]`. The list is initialized with elements of various types, but they all are instance of `scala.Any`, so you can add them to the list.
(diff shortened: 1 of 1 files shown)


shogowada:
This is approved but not merged. What's the next step?

SethTisue:
since there hasn't been any other feedback, I'll hit merge. thanks for improving this!

heathermiller:
🎉  Thank you @shogowada! 🙂

PR #671, opened 2017-02-02
Add Basics tutorial page

shogowada:
I believe majority of people coming to tutorials are more interested in quick examples than detailed explanations.

The current first real tutorial is Unified Types. While it's helpful to know that information, it doesn't give you much insight into how Scala code would look like. I think it can lose interest of readers.

Also, while many pages are using features like `object`, `App`, `var`, `val`, etc, what they are isn't necessarily clearly communicated before it's introduced. I think it can throw out readers.

I created a Basics page to address those concerns. The page is supposed to provide just the most important information out of the most important topics.

Personally, for this kind of tutorial, I think it makes more sense to use REPL because it will be easier for writers to make sure what they provide is completely reproducible on reader's machine, and it will be more interesting to readers because it will be easier to interact and play with the code. While I didn't do it because it will be a bigger change if you consider consistency, I think it's worth considering.

Another option is to give them link to the [ScalaFiddle](https://scalafiddle.io) so that we can achieve the same goal without readers installing anything on their machine.

I tried to test it locally, but I am having trouble setting up the build environment (`bundle install` fails because `gem install json -v '1.8.3'` fails, tried multiple solutions found online, and eventually gave up), so I haven't tested it yet, but it looks fine on GitHub.

### DIFF
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -24,18 +24,24 @@ It is a perfect way for anybody to experiment a piece of Scala code anywhere, an
 
 ## Variables
 
-You can define variables with `var` keyword.
-
-```
-var x = 1 + 1
-x += 1
-```
-
-Often, you want your variables to be immutable. You can use `val` keyword in that case.
+You can define variables with the `val` keyword.
 
 ```
 val x = 1 + 1
-x += 1 // This does not compile because you declared the variable with "val" keyword
+```
+
+Variables defined by `val` cannot be re-assigned and are immutable in that sense.
+
+```
+val x = 1 + 1
+x += 2 // This does not compile because it is re-assigning the variable.
+```
+
+If you need a mutable variable, you can use the `var` keyword instead.
+
+```
+var y = 1 + 1
+y += 2 // This compiles because "y" is declared with "var" keyword.
 ```
 
 Type of variables can be inferred, but you can also explicitly state type like below.
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -22,32 +22,33 @@ You can run Scala in your browser by using [ScalaFiddle](https://scalafiddle.io)
 
 It is a perfect way for anybody to experiment with a piece of Scala code anywhere, anytime!
 
-## Variables
+## Values and Variables
 
-You can define variables with the `val` keyword.
-
-```
-val x = 1 + 1
-```
-
-Variables defined by `val` cannot be re-assigned and are immutable in that sense.
+You can define values (a.k.a. immutable variables) with the `val` keyword.
 
 ```
 val x = 1 + 1
-x += 2 // This does not compile because it is re-assigning the variable.
 ```
 
-If you need to mutate variables, you can use the `var` keyword instead.
+Values cannot be re-assigned and are immutable in that sense.
+
+```
+val x = 1 + 1
+x = 3 // This does not compile because it is re-assigning the value.
+```
+
+If you need to re-assign, you must declare them as variables using the `var` keyword.
 
 ```
 var y = 1 + 1
-y += 2 // This compiles because "y" is declared with "var" keyword.
+y = 3 // This compiles because "y" is declared with "var" keyword.
 ```
 
-Type of variables can be inferred, but you can also explicitly state type like below.
+Types of values and variables can be inferred, but you can also explicitly state types like below.
 
 ```
 val x: Int = 1 + 1
+var y: Int = 1 + 1
 ```
 
 ## Functions
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -75,9 +75,83 @@ println(point) // (1, 2)
 
 We will cover classes in depth [later](classes.md).
 
+## Case Classes
+
+By default, classes are compared by reference. If you want to compare them by values, you need to override `equals` and `hashCode` methods.
+
+```
+class Point(val x: Int, val y: Int) {
+  override def equals(o: Any) = o match {
+    case that: Point => x == that.x && y == that.y
+    case _ => false
+  }
+  override def hashCode = (x.hashCode << 16) + y.hashCode
+}
+```
+
+While it works, it's tedious to define those methods for all classes you want to compare by value.
+
+Thankfully, Scala has another type of class called `case class` that's compared by value by default.
+
+```
+case class Point(x: Int, y: Int)
+```
+
+You can instantiate case classes without `new` keyword.
+
+```
+val point = Point(1, 2) // no "new" here
+val anotherPoint = Point(1, 2)
+val yetAnotherPoint = Point(2, 2)
+println(point == anotherPoint) // true
+println(point == yetAnotherPoint) // false
+```
+
+Case classes are immutable by default, but it also provides `copy` method so that you can easily create another instance of the class while reusing the values from exiting instances.
+
+Using `copy` method, you can also write the above code like below.
+
+```
+val point = Point(1, 2)
+val anotherPoint = point.copy()
+val yetAnotherPoint = point.copy(x = 2)
+println(point == anotherPoint) // true
+println(point == yetAnotherPoint) // false
+```
+
+Because of case classes, you almost never see classes overriding `equals` and `hashCode` methods in Scala.
+
+There are many other features you get out-of-the-box by using case classes. We will cover them in depth [later](case-classes.md).
+
+## Singleton Objects
+
+You can define singleton objects with `object` keyword.
+
+```
+object IdFactory {
+  private var counter = 0
+
+  def create(): Int = {
+    counter += 1
+    counter
+  }
+}
+```
+
+You can access singleton objects just by referring its name.
+
+```
+val newId: Int = IdFactory.create()
+println(newId) // 1
+val newerId: Int = IdFactory.create()
+println(newerId) // 2
+```
+
+We will cover singleton objects in depth [later](singleton-objects.md).
+
 ## Traits
 
-Traits define types as signature of fields and methods.
+Traits define specification of types as signature of fields and methods.
 
 You can define traits with `trait` keyword.
 
@@ -111,80 +185,6 @@ class MyPoint(var x: Int, var y: Int) extends Point {
 ```
 
 We will cover traits in depth [later](traits.md).
-
-## Case Classes
-
-By default, classes are compared by reference. If you want to compare them by values, you need to override `equals` and `hashCode` methods.
-
-```
-class Point(x: Int, y: Int) {
-  override def equals(o: Any) = o match {
-    case that: Point => x == that.x && y == that.y
-    case _ => false
-  }
-  override def hashCode = (x.hashCode << 16) + y.hashCode
-}
-```
-
-It works, but it's tedious to define those methods for all classes you want to compare by value.
-
-Thankfully, Scala has another type of class called `case class` that's compared by value by default.
-
-```
-case class Point(x: Int, y: Int)
-```
-
-Because of case classes, you almost never see Scala classes overriding `equals` and `hashCode` methods.
-
-You can instantiate case classes without `new` keyword.
-
-```
-val point = Point(1, 2) // no "new" here
-val anotherPoint = Point(1, 2)
-val yetAnotherPoint = Point(2, 2)
-println(point == anotherPoint) // true
-println(point == yetAnotherPoint) // false
-```
-
-Case classes are immutable by default, but it also provides `copy` method so that you can easily create another instance of the class while reusing the values from exiting instances.
-
-Using `copy` method, you can also write the above code like below.
-
-```
-val point = Point(1, 2)
-val anotherPoint = point.copy()
-val yetAnotherPoint = point.copy(x = 2)
-println(point == anotherPoint) // true
-println(point == yetAnotherPoint) // false
-```
-
-There are many other features you get out-of-the-box by using case classes. We will cover them in depth [later](case-classes.md).
-
-## Singleton Objects
-
-You can define singleton objects with `object` keyword.
-
-```
-object IdFactory {
-  private var counter = 0
-
-  def create(): Int = {
-    counter += 1
-    counter
-  }
-}
-```
-
-You can access singleton objects just by referring its name.
-
-```
-val newId: Int = IdFactory.create()
-println(newId) // 1
-val newerId: Int = IdFactory.create()
-println(newerId) // 2
-```
-
-We will cover singleton objects in depth [later](singleton-objects.md).
 
 ## Main Method
 
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -22,48 +22,79 @@ You can run Scala in your browser by using [ScalaFiddle](https://scalafiddle.io)
 
 It is a perfect way for anybody to experiment with a piece of Scala code anywhere, anytime!
 
-## Values and Variables
+## Expressions
 
-You can define values (a.k.a. immutable variables) with the `val` keyword.
+Expressions are computable statements.
+
+You can output results of expressions using the `println`.
+
+```
+println(1) // 1
+println(1 + 1) // 2
+println("Hello!") // Hello!
+println("Hello," + " world!") // Hello, world!
+```
+
+You can name the result of an expression by using the `val` keyword.
 
 ```
 val x = 1 + 1
+println(x) // 2
 ```
 
-Values cannot be re-assigned and are immutable in that sense.
+Named results are called values. For example, in above case, `x` is called a value. Values only hold results of expressions, so they will not re-compute expressions when referenced.
+
+Values cannot be re-assigned.
 
 ```
 val x = 1 + 1
-x = 3 // This does not compile because it is re-assigning the value.
+x = 3 // This does not compile.
 ```
 
-If you need to re-assign, you must declare them as variables using the `var` keyword.
-
-```
-var y = 1 + 1
-y = 3 // This compiles because "y" is declared with "var" keyword.
-```
-
-Types of values and variables can be inferred, but you can also explicitly state types like below.
+Types of values can be inferred, but you can also explicitly state types like below.
 
 ```
 val x: Int = 1 + 1
-var y: Int = 1 + 1
+```
+
+You can create a single expression out of multiple statements by wrapping them with `{}`. When statements are wrapped with `{}`, the result of the last statement will be the result of the overall expression.
+
+```
+println({
+  val x = 1 + 1
+  x + 1
+}) // 3
+```
+
+## Variables
+
+Variables are similar to values except you can re-assign results they hold. You can define variables with the `var` keyword.
+
+```
+var x = 1 + 1
+x = 3 // This compiles because "x" is declared with the "var" keyword.
+println(x) // 3
+```
+
+Just like values, you can also explicitly state types of variables like below.
+
+```
+var x: Int = 1 + 1
 ```
 
 ## Functions
 
-You can define functions that returns a given integer + 1 like below.
+Functions are expressions with parameters.
+
+You can define a function that returns a given integer + 1 like below.
 
 ```
 (x: Int) => x + 1
 ```
 
-Left hand side of `=>` is parameter(s) and right hand side of `=>` is body.
+Left hand side of `=>` is a list of parameters and right hand side of `=>` is an expression taking the parameters.
 
-Notice that the function has no `return` keyword. This is because, in Scala, the last expression of the function is automatically returned.
-
-You can also assign functions to variables.
+You can also name functions.
 
 ```
 val addOne = (x: Int) => x + 1
@@ -84,36 +115,25 @@ val getTheAnswer = () => 42
 println(getTheAnswer()) // 42
 ```
 
-If your function body spans across multiple lines, you must wrap them with `{}`.
-
-```
-val greet = (name: String) => {
-  println("Hello, " + name + "!")
-  println("How are you doing today?")
-}
-greet("Scala developer")
-// Hello, Scala developer!
-// How are you doing today?
-```
-
 We will cover functions in depth [later](anonymous-function-syntax.md).
 
 ## Methods
 
 Methods look and behave very similar to functions, but there are a few key differences between them.
 
-Methods are defined like below with the `def` keyword followed by its name, parameter(s), return type, and body.
+Methods are defined like below with the `def` keyword followed by its name, a list of parameter groups, return type, and an expression.
 
 ```
 def add(x: Int, y: Int): Int = x + y
 println(add(1, 2)) // 3
 ```
 
-Methods cannot be assigned to variables.
+Methods cannot be named with the `val` or `var` keywords.
 
 ```
 def add(x: Int, y: Int): Int = x + y
 val add2 = add // This does not compile.
+var add3 = add // This does not compile either.
 ```
 
 Methods can take multiple parameter groups.
@@ -123,7 +143,7 @@ def addThenMultiply(x: Int, y: Int)(multiplier: Int): Int = (x + y) * multiplier
 println(addThenMultiply(1, 2)(3)) // 9
 ```
 
-Or no parameter group at all.
+Or no parameter groups at all.
 
 ```
 def name: String = System.getProperty("name")
@@ -165,7 +185,11 @@ You can instantiate case classes without `new` keyword.
 val point = Point(1, 2)
 val anotherPoint = Point(1, 2)
 val yetAnotherPoint = Point(2, 2)
+```
 
+And they are compared by values.
+
+```
 if (point == anotherPoint) {
   println(point + " and " + anotherPoint + " are the same.")
 } else {
@@ -223,7 +247,7 @@ trait Greeter {
 }
 ```
 
-Traits can also have default implementation.
+Traits can also have default implementations.
 
 ```
 trait Greeter {
@@ -231,7 +255,7 @@ trait Greeter {
 }
 ```
 
-You can extend traits with the `extends` keyword and override their implementation with the `override` keyword.
+You can extend traits with the `extends` keyword and override their implementations with the `override` keyword.
 
 ```
 class DefaultGreeter extends Greeter
@@ -252,6 +276,8 @@ customGreeter.greet("Scala developer") // How are you, Scala developer?
 We will cover traits in depth [later](traits.md).
 
 ## Main Method
+
+Main method is an entry point of a program.
 
 Using objects, you can define main methods like below.
 
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -52,18 +52,59 @@ val x: Int = 1 + 1
 
 ## Functions
 
-You can define functions like below.
+You can define functions that returns a given integer + 1 like below.
 
 ```
-(name: String) => println("Hello, " + name + "!")
+(x: Int) => x + 1
 ```
+
+Left hand side of `=>` is parameter(s) and right hand side of it is body.
+
+Notice that the function has no `return` keyword. This is because, in Scala, the last expression of the function is automatically returned.
 
 You can also assign functions to variables.
 
 ```
-val greet = (name: String) => println("Hello, " + name + "!")
-greet("Scala") // Hello, Scala!
+val addOne = (x: Int) => x + 1
+println(addOne(1)) // 2
 ```
+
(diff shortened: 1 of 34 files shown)


tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+object Main {
+  def main(args: Array[String]): Unit = {
+    println("Hello, world!")
+  }
+}
+```
+
+Because every runnable application has a main method, Scala also provides a more convenient way of defining your main method by extending `App` trait.
```
MasseGuillaume: This trait is deprecated. (App -> DelayedInit)

http://www.scala-lang.org/api/current/scala/DelayedInit.html

(Since version 2.11.0) DelayedInit semantics can be surprising. Support for App will continue. See the release notes for more details: https://github.com/scala/scala/releases/tag/v2.11.0-RC1

MasseGuillaume:
That's an awesome start!

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+## Variables
+
+You can define variables with `var` keyword.
+
+```
+var x = 1 + 1
+x += 1
+```
+
+Often, you want your variables to be immutable. You can use `val` keyword in that case.
```
jvican: What about "Mutable variables are not idiomatic Scala and therefore discouraged. Most of the times, you can use `val` definitions."?

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Functions
+
+You can define functions like below.
+
+```
+(name: String) => println("Hello, " + name + "!")
+```
+
+You can also assign functions to variables.
```
jvican: Can you explain the different syntactic parts of a function? Signature and body? This is good :smile:, but I think those parts should be brought to the attention of the reader.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+You can define classes with `class` keyword.
+
+```
+class Point(var x: Int, var y: Int) {
+  def move(dx: Int, dy: Int): Unit = {
+    x = x + dx
+    y = y + dy
+  }
+
+  override def toString: String = "(" + x + ", " + y + ")"
```
jvican: I would not write this in the class example, since case classes are already doing this for you. Do you think it's a good idea to remove it and add the following to the case class section?
> What if we want to `println` a case class? The Scala compiler automatically writes it for you, so it's not necessary that you override `toString`.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+```
+val point = new Point(1, 2)
+println(point) // (1, 2)
+```
+
+We will cover classes in depth [later](classes.md).
+
+## Case Classes
+
+By default, classes are compared by reference. If you want to compare them by values, you need to override `equals` and `hashCode` methods.
```
jvican: I would enumerate more clearly the benefits of a case class over a class, either here or at the end of this section. I think it's good to show them need by need, but they should be summarized somewhere :smile:.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+val point = Point(1, 2) // no "new" here
+val anotherPoint = Point(1, 2)
+val yetAnotherPoint = Point(2, 2)
+println(point == anotherPoint) // true
+println(point == yetAnotherPoint) // false
+```
+
+Case classes are immutable by default, but it also provides `copy` method so that you can easily create another instance of the class while reusing the values from exiting instances.
```
jvican: "Case classes are immutable by default. They provide a `copy` method that allows you to modify any of the values defined in a case class."

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+
+```
+val point = Point(1, 2)
+val anotherPoint = point.copy()
+val yetAnotherPoint = point.copy(x = 2)
+println(point == anotherPoint) // true
+println(point == yetAnotherPoint) // false
+```
+
+Because of case classes, you almost never see classes overriding `equals` and `hashCode` methods in Scala.
```
jvican: Case classes generate their own `equals` and `hashCode` methods so that you don't need to define them yourselves. If you want custom implementations, you can always override them.

tutorials/tour/basics.md
jvican: This trait is not deprecated. It will be maintained both in future Scala and Dotty versions.

MasseGuillaume: You don't want to start an application inside a constructor.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+```
+val x: Int = 1 + 1
+```
+
+## Functions
+
+You can define functions like below.
+
+```
+(name: String) => println("Hello, " + name + "!")
```
travis032654: This example introduces a few things that might not be obvious coming from certain languages:

- Anonymous functions
- Type declaration after identifier
- Return type inference 
- Lack of "return" keyword

I think it's important to explain all of those or at least declare the return type and not make the function anonymous. Otherwise, it could be a bit cryptic.

travis032654:
This tutorial makes Scala a lot less intimidating for an absolute beginner.

tutorials/tour/basics.md
a1russell: @jvican That wouldn't tell the newcomer that `val` defines an immutable variable. I also don't think we should conflate the goals of a style guide (to explain "idiomatic" Scala) with what should simply be the introduction to the language features.

heathermiller: @a1russell, I don't necessarily agree with this sentiment. Because you can write code that is essentially Java with a `.scala` extension and pass it through the Scala compiler, I think it's extremely important to focus on the basics of what Scala really is in this article. When it comes to immutable vs mutable references (`val` vs `var), Scala is all about `val`s. It's rare to have to use a `var` anywhere in a normal Scala application. Therefore, I think introducing `val`s should probably come first, and `var`s should probably be introduced as that thing from Java that you can use if you absolutely have to, but is generally avoided.

heathermiller:
This is great @shogowada!! Would you be up for addressing some of the reviewer feedback?

heathermiller:
PS, I wish something like this existed when I was trying to learn Scala years ago. This is perfect for someone who already knows another language and who just wants to quickly be introduced to the "main Scala things one's gotta know about". This is great because it's so succinct and so concise; in literally 10 minutes, a Java developer could probably write a cute little Scala application. 

I love this ❤️

tutorials/tour/basics.md
a1russell: @heathermiller I agree that `val` should probably be introduced first. My issue was with @jvican's proposed wording is all.

jvican: It's not only that it should be introduced first, it should be emphasized as the de-facto alternative when declaring variables, @heathermiller conveys very well why. @a1russell

tpolecat: `var` is an advanced feature that should not be mentioned at all in a beginner document, or should be mentioned only in passing.

tutorials/tour/basics.md
tpolecat: I would start with `case class` … non-`case` classes are relatively rare in Scala code. `new` and reference equality are advanced feature that should be de-emphasized.

tutorials/tour/basics.md
tpolecat: Agree, `App` is problematic due to `DelayedInit`. I suggest not mentioning it.

tutorials/tour/basics.md
tpolecat: You are using `def` and `override` which you have not yet introduced, as well as the constructor syntax which is unusual when compared with other OO languages. I would explain these.

tutorials/tour/basics.md
tpolecat: I would leave the `equality/hashCode` business until a chapter on classes. It's introducing a lot of new and problematic ideas that probably don't belong on page 1.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+println(point == yetAnotherPoint) // false
+```
+
+Because of case classes, you almost never see classes overriding `equals` and `hashCode` methods in Scala.
+
+There are many other features you get out-of-the-box by using case classes. We will cover them in depth [later](case-classes.md).
+
+## Singleton Objects
+
+You can define singleton objects with `object` keyword.
```
tpolecat: We should be careful about "singleton" because it alludes to the Go4 pattern which isn't quite right. I would say "`object` defines a module of definitions" or something like that.

Introducing global mutable state in an `object` is not a good example. Maybe just declare some values and methods?

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Traits can also have implementation.
+
+```
+trait Point {
+  var x: Int
+  var y: Int
+  
+  override def toString: String = "(" + x + ", " + y + ")"
```
tpolecat: Can we provide an implementation that's not an override, since we haven't talked about that yet? How about `def distanceFromOrigin: Double = math.hypot(x, y)`?

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+}
+```
+
+You can extend traits with `extends` keyword.
+
+```
+class MyPoint(var x: Int, var y: Int) extends Point {
```
tpolecat: I recommend a case class here and `offset` that returns a new instance.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,207 @@
...
+
+We will cover traits in depth [later](traits.md).
+
+## Main Method
+
+Using singleton objects, you can define main methods like below.
+
+```
+object Main {
+  def main(args: Array[String]): Unit = {
```
Sciss: How about removing the unnecessary braces? That also shows why regular syntax without procedure variant is useful:

```scala
  def main(args: Array[String]): Unit =
    println("Hello, world!")
```

stephenjudkins:
I think we should move `var`, non-case classes, custom `toString`, and `override` out of the intro sections of this. None of these are really idiomatic Scala at this level. I'm OK with having them somewhere in the early sections, with a brief caveat that good Scala style is to avoid them in favor of immutable values and data types. I just don't think they should be presented so early, especially not before things like `val` and `case class`.

shogowada:
Thank you all for the feedbacks! It's really helpful. I'll revise the documentation based on the feedbacks.

I realized that I might not necessarily agree with what the "Basics" should be with everybody in the thread. So please let me explain where I'm coming from.

What I want to achieve here is to grow the size of Scala community by welcoming developers from diverse backgrounds and beliefs. Although the diversity might introduce some conflicts here and there, I believe those conflicts ultimately benefit the community as a whole by letting them think and innovate. After all, if two people are having the same thought, one of them is redundant.

I believe Scala has a great potential of becoming such a community because, IMO, the language itself is not so opinionated towards one paradigm or another; it supports both OOP and FP, both mutable and immutable objects/collections, etc without developers fighting with the language.

To welcome developers from various background, one thing I want to make sure in this Basics tutorial is not to intimidate them. To achieve that, I think we need to sacrifice some of the "Scala-ism" especially in this Basics page. In other words, it's not my intention to make this page "Scala FP Basics", "Scala OOP Basics", or even "Scala Best Practices". My intention is rather to make this page just "Scala Basics".

For example, there are many popular languages where `var` equivalent is default instead of `val` equivalent. While I agree that we should introduce `val` first (and I'll fix that!), I think not mentioning `var` at all confuses many readers. Worst case, it might mislead them Scala is too limited and opinionated for them, and I don't want that to happen. I'd like to introduce non-case classes in this page too for the same reason.

Given that, I hope you could understand where I am coming from and why I am trying to be make this page like how it is.

But I agree this page is currently introducing some unnecessarily things such as `override` or `toString`. I'll try to simplify them.

Thanks again for all the feedbacks!

tutorials/tour/basics.md
shogowada: > We should be careful about "singleton" because it alludes to the Go4 pattern which isn't quite right.

Can you elaborate? Do you mean it is not a singleton because the instance is not retrieved via method? I think "singleton" is more friendly than "a module of definitions" to most readers, and it's not wrong either. Do you agree?

> Introducing global mutable state in an object is not a good example.

I agree, but I couldn't think of better way of showing this object actually is a singleton without having states. Ideas?

tutorials/tour/basics.md
shogowada: Sure. Maybe my assumption of "most readers are already familiar with anonymous function" is wrong. Even if it was right, I agree it's better be explained. Let me think of a simpler way of introducing functions.

tutorials/tour/basics.md
shogowada: It sounds better, but I think "modify any of the values defined in a case class" also sounds like it will mutate values, which is not what's happening. I cannot think of a simple way of wording what's actually happening, though. I started to think maybe `copy` is not worth mentioning in this page.

tutorials/tour/basics.md
shogowada: Good point. Let me address those concerns.

tutorials/tour/basics.md
a1russell: I was thinking of something more along the lines of:

> You can define values with the `val` keyword.
>
> ```
> val x = 1 + 1
> ```
>
> Values cannot be re-assigned and are immutable in that sense. If you do require a mutable variable, however, you can use the `var` keyword. However, `val` should be used by default since `var` generally indicates that it *will* be mutated down the line.
>
> ```
> var y = 1 + 2
> y += 4
> ```

Feel free to further clean up the wording, but really all I meant was that I think we should avoid hand-wavy words like "idiomatic", and when we introduce something, we should briefly describe what it is.

tutorials/tour/basics.md
a1russell: "Module of definitions" sounds more like a trait to me, especially when reading it from the perspective of a Ruby developer.

tutorials/tour/basics.md
a1russell: To play Devil's advocate, much like the [1TBS](https://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS_.28OTBS.29) vs [traditional K&R](https://en.wikipedia.org/wiki/Indent_style#K.26R_style) religious war, the argument could be made that consistency is preferable, and that braces should stay no matter whether the method is one or multiple lines.

I realize this is against the mainstream Scala community consensus and the [Scala style guide](http://docs.scala-lang.org/style/declarations#methods), however.

tutorials/tour/basics.md
tpolecat: An `object` is a module. A namespace. Like a package, but unlike a package it's a value. When you say "singleton", as a Java programmer I think "singleton of what?" ... is `object Foo` an instance of `class Foo`? I realize "singleton" is the common usage but to me it's misleading.

tutorials/tour/basics.md
tpolecat: Yes, +1. Remove unnecessary braces.

tutorials/tour/basics.md
shogowada: @jvican 

> I would not write this in the class example, since case classes are already doing this for you.

I agree. Let me change the example to something more suitable for classes.

@tpolecat 

> I would start with case class … non-case classes are relatively rare in Scala code.

We need to make this tutorial easy to understand for non-Scala developers. My concern is that introducing case classes first confuses them because case class is very specific to Scala.

> You are using def and override which you have not yet introduced, as well as the constructor syntax

Good point. I will add section for "Methods" to cover the `def` keyword and avoid using `override` here at all. I'll also add some explanation around constructor.

tutorials/tour/basics.md
shogowada: @jvican 

> I would enumerate more clearly the benefits of a case class over a class

While I agree there is a lot more to case class, I am also concerned mentioning things like pattern matching or even `copy` method (that I'm thinking to remove) can intimidate readers.

It is not my goal to give readers a complete picture of Scala. My goal here is to nudge readers into writing their first Scala code because I am convinced that once they start writing it, majority of them will have a positive experience.

Scala is a relatively complex language with many features. One common dilemma I see is that its own feature richness is intimidating newcomers, while the language is actually designed to make your code as simple as possible.

In the end of this section, I have the following sentence.

"There are many other features you get out-of-the-box by using case classes. We will cover them in depth [later](http://docs.scala-lang.org/tutorials/tour/case-classes)."

I might change the wording so that it is clearer that there is a lot more to case classes, but do you think having a link to the complete [Case Classes page](http://docs.scala-lang.org/tutorials/tour/case-classes) is good enough?

@tpolecat 

Agreed. Let me remove them.

tutorials/tour/basics.md
shogowada: I see. Let me name the section "Objects" instead of "Singleton Objects", then explain what it really is in details in the section.

tutorials/tour/basics.md
ivanopagano: I'm not sure that _immutable_ should be used at all. I (maybe incorrectly) expect that newcomers will be later confused by the whole immutable reference vs. immutable object argument, since it's not immediately intuitive that immutability at the variable definition level doesn't guarantee immutability of _what_ is referenced by that variable name.

@a1russell your definition seems good to me in spirit, but I'd rephrase the second part as

> `val`s cannot be re-assigned and they will always reference the same value. If you need to re-assign a variable, you can use the `var` keyword. It is considered good practice to reduce the use of `var`s to the bare minimum.
  ```
  var y = 1 + 2
  y += 4
  //y now is equal to 7
  ```

shogowada:
I think I've addressed most of the concerns. I'd appreciate if you could review again.

I also added ["Running Scala on Browser" section](https://github.com/scala/scala.github.com/pull/671/files#diff-c2f160f90dcc962112c59cf07bb40f00R15) at the top so that it will be super easy for readers to start writing their first Scala code.

Sciss:
On 03/02/17 03:42, Adam Russell wrote:
> To play Devil's advocate, much like the 1TBS
> <https://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS_.28OTBS.29> vs
> traditional K&R <https://en.wikipedia.org/wiki/Indent_style#K.26R_style>
> religious war, the argument could be made that consistency is
> preferable, and that braces should stay no matter whether the method is
> one or multiple lines.
> 
> I realize this is against the mainstream Scala community consensus and
> the Scala style guide
> <http://docs.scala-lang.org/style/declarations#methods>, however.
> 

Well, the style guide is some authority ;)
I think Scala is about minimising ceremony, for me that clearly includes
dropping unnecessary braces and other visual clutter...

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+
+tutorial: scala-tour
+num: 2
+next-page: unified-types
+previous-page: tour-of-scala
+---
+
+In this page, we will cover basics of Scala.
+
+## Running Scala on Browser
```
travis032654: Maybe "Trying Scala in the Browser" to be clear it's not a feature of the language but rather a way to try it out.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+
+You can run Scala on browser by using [ScalaFiddle](https://scalafiddle.io).
```
travis032654: "on browser" => "in your browser"

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+
+1. Go to https://scalafiddle.io.
+2. Copy and paste `println("Hello, world!")` to the left pane.
+3. Hit "Run" button and see it prints "Hello, world!" on the right pane.
+
+It is a perfect way for anybody to experiment a piece of Scala code anywhere, anytime!
```
travis032654: "experiment a piece of Scala code" => "experiment _with_ a piece of Scala code"

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Variables
+
+You can define variables with the `val` keyword.
```
travis032654: Is there another way to explain vals without calling them variables? Of course in many languages, there are no vals and they _are_ quite similar to variables so I see why you chose this approach. Perhaps you could call this section "Values and Variables".

DarkDimius: Would something along the lines of
 "you can give a name to a computed expression using a keyword `val`"
be better?

tpolecat: The `val` keyword binds an identifier to a computed value.

DarkDimius: @tpolecat, that's precisely the version I started with in my mind, but then I changed to "gives name" because I think that binding is a more formal\scientific term, while "give name" would be more accessible.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+```
+val x = 1 + 1
+x += 2 // This does not compile because it is re-assigning the variable.
+```
+
+If you need to mutate variables, you can use the `var` keyword instead.
+
+```
+var y = 1 + 1
+y += 2 // This compiles because "y" is declared with "var" keyword.
```
tpolecat: It might be better here and in the case above to use simple assignment like `y = 42` rather than the compound operation `+=`. Or at least desugar it to `y = y + 2`.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+Or it can take multiple parameters.
+
+```
+val add = (x: Int, y: Int) => x + y
+println(add(1, 2)) // 3
+```
+
+Or it can take no parameters.
```
tpolecat: To avoid confusion with no-arg `def`s we might want to say "an empty argument list".

tutorials/tour/basics.md
```diff
@@ -0,0 +1,261 @@
...
+
+```
+val getTheAnswer = () => 42
+println(getTheAnswer()) // 42
+```
+
+If your function body spans across multiple lines, you can wrap them with `{}`.
```
tpolecat: s/can/must/

tutorials/tour/basics.md
shogowada: Oh, I just didn't know the official term:) Thanks for pointing this out.

Maybe something like "You can define values (a.k.a. immutable variables) with the `val` keyword." will be correct and understandable? I see a lot of website referring `val` as immutable variables.

heathermiller: I'd side with @DarkDimius and @shogowada on this one; it's just friendlier to leave the sciencespeak out here. While "blah keyword binds an identifier to a computed expression" is the terminology you'd use in a PL fundamentals course or in a paper, I think most developers would find this terminology alienating. 

So yeah, how about this average between @shogowada's and @DarkDimius's suggestions:

"You can give names a computed expression using the `val` keyword. We call these _values_. Importantly, values are immutable; that is, once a computed expression is named with the `val` keyword, it cannot be reassigned to refer to a different computed expression."

(PS, we shouldn't use "immutable variable" as it's not exactly correct terminology)

WDYT?

tutorials/tour/basics.md
shogowada: Good point; it can be confusing.

I think methods taking multiple parameter groups is more advanced than function taking parameters. Maybe we should instead be explicit on methods section, explicitly stating difference between parameters and parameter groups. Then, readers can understand difference between "no parameters" vs. "no parameter groups" without confusing.

tutorials/tour/basics.md
shogowada: Sorry, I prematurely modified the line so now this comment is outdated. Let me start a new comment on the latest line.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. Go to https://scalafiddle.io.
+2. Copy and paste `println("Hello, world!")` to the left pane.
+3. Hit "Run" button and see it prints "Hello, world!" on the right pane.
+
+It is a perfect way for anybody to experiment with a piece of Scala code anywhere, anytime!
+
+## Values and Variables
+
+You can define values (a.k.a. immutable variables) with the `val` keyword.
```
shogowada: Continued from https://github.com/scala/scala.github.com/pull/671#discussion_r99347365

It's tough:) So,

- we don't want to use the term "immutable variable" because it's technically incorrect.
- I personally think "computed expression" can be intimidating although it is correct.

Maybe we should have "Expressions" section before "Values and Variables", so that terms like "computed expression" or "result of the expression" won't be as intimidating. There, we can also state things like "almost everything in Scala is an expression" or "multiple expressions must be grouped by `{}`", and we will be able to use the term "expression" without intimidating readers for the rest of the tutorial. In the section, we can use `println` as a tool to show the result of each expression because we are not using REPL.

How do you guys think?

tutorials/tour/basics.md
shogowada: Moved to https://github.com/scala/scala.github.com/pull/671/files#r99466445.

tutorials/tour/basics.md
ivanopagano: Why not simply?
> You can assign a name to the result of an expression, with the `val` keyword

tutorials/tour/basics.md
```diff
@@ -84,36 +115,25 @@ val getTheAnswer = () => 42
...
-```
-
 We will cover functions in depth [later](anonymous-function-syntax.md).
 
 ## Methods
 
 Methods look and behave very similar to functions, but there are a few key differences between them.
 
-Methods are defined like below with the `def` keyword followed by its name, parameter(s), return type, and body.
+Methods are defined like below with the `def` keyword followed by its name, a list of parameter groups, return type, and an expression.
```
tpolecat: It might be worth mentioning that (unlike functions) methods can also have *type* parameters.

shogowada: I felt like type parameters should not be included in this Basics page because then we probably need to add it for classes, case classes, and traits section too, and we will have hard time explaining what it is especially for people coming from dynamically typed language backgrounds.

I think we probably should create a whole new page dedicated for type parameters. Right now the topic seems to be scattered across multiple pages.

shogowada:
Any other concerns? Let me know if I missed some, too.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,294 @@
...
+2. Copy and paste `println("Hello, world!")` to the left pane.
+3. Hit "Run" button and see it prints "Hello, world!" on the right pane.
+
+It is a perfect way for anybody to experiment with a piece of Scala code anywhere, anytime!
+
+## Expressions
+
+Expressions are computable statements.
+
+You can output results of expressions using the `println`.
```
a1russell: Probably don't need "the":
> You can output results of expressions using `println`.

tutorials/tour/basics.md
```diff
@@ -0,0 +1,294 @@
...
+
+```
+println(1) // 1
```
a1russell: Do you think it's obvious enough that `//` indicates a comment?

heathermiller: Syntax highlighting should make it pretty obvious. So I think we don't have to worry about this 🙂

shogowada: Ah, I didn't even think of that. But I think you are right and we'd better explicitly state that because not all language use `//` for comments.

heathermiller:
🎉 Thank you so much for your contribution @shogowada!! This is really huge! We **really** appreciate it ❤️ 

Merging now. Any further feedback, we can deal with in a separate PR 🙂 

(I'll address @a1russell's comment about removing "the" in a follow-up commit)

tutorials/tour/basics.md
```diff
@@ -0,0 +1,294 @@
...
+
+if (point == yetAnotherPoint) {
+  println(point + " and " + yetAnotherPoint + " are the same.")
+} else {
+  println(point + " and " + yetAnotherPoint + " are different.")
+}
+// Point(1,2) and Point(2,2) are different.
+```
+
+There is a lot more to case classes that we'd like to introduce, and we are convinced you will fall in love with it! We will cover them in depth [later](case-classes.md).
```
a1russell: Fall in love with *them*, I think.

heathermiller: Thanks, I'll address this too 🙂

tutorials/tour/basics.md
shogowada: > Syntax highlighting should make it pretty obvious.

That's a good point too.

PR #674, opened 2017-02-07
SIP-NN - Match infix & prefix types to meet expression rules

soronpo:


sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Backward Compatibility
+Changing infix type associativity and precedence affects code that uses type operations and conforms to the current specification.
+(Author's note: I don't know if providing a flag to select the precedence is good or not. IMHO, it is better to create a tool that adds brackets to convert code to the old associativity.)    
+
+---
+
+### Extended proposal alternative
+It is possible to extend this proposal and allow the developer to annotate the expected associativity and precedence per operation. 
+(Author's note: I personally don't like this, but if such a solution is better for the community, then I will gladly modify this SIP to reflect that.)
```
jvican: I don't think this is a good idea as it adds a significant layer of complexity on the language and forces Scala developers to inspect the library-defined operations to know what's their precedence and how they can be used.
  
I would propose this is not included in this document and, if someone wants this feature, that a new proposal is submitted!

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+val fails : 1 + 2 * 3 + 4 = 11 //left associative:(((1+2)*3)+4))) = 13
+```
+
+#### Developer issues
+The following stackoverflow question demonstrate developers are 'surprised' by the difference in infix precedence.
+http://stackoverflow.com/questions/23333882/scala-infix-type-aliasing-for-2-type-parameters
+
+
+
+---
```
jvican: Are these dashes intended?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+    type Result_Surprise = N1 + N2 * N3 + N4
+    //Result_Surprise expands to Plus[Prod[Plus[N1,N2],N3],N4]
+    type Result_Expected = N1 + (N2 * N3) + N4
+    //Result_Expected expands to Plus[Plus[N1,Prod[N2,N3]],N4]
+}
+```
+
+* **Prefix operators bracketless unary use**: While expressions have prefix unary operators, there are none for types. See the [Prefix Operations](http://scala-lang.org/files/archive/spec/2.12/06-expressions.html#prefix-operations) section of the Scala specification. 
+This is a lacking feature of the type language Scala offers. See also interactions of this feature with other Scala features, further down this text. 
+(Author's note: Not crucial as infix precedence, but good for completeness)
```
jvican: Can you remove author's notes? :smile:. The document will be exposed to all the community and add a little bit of overhead to the document.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+---
+## Introduction
+Currently scala allows symbol operators (`-`, `*`, `~~>`, etc.) for both type names and definition names.
+Unfortunately, there is a 'surprise' element since the two differ in behaviour:
+
+* **Infix operator precedence and associativity**: 
+Infix types are 'mostly' left-associative, 
+while the expression operations are more intuitive with different precedence weights.
+Please see [Infix Types](http://scala-lang.org/files/archive/spec/2.12/03-types.html#infix-types) and [Infix Operations](http://scala-lang.org/files/archive/spec/2.12/06-expressions.html#infix-operations) sections of the Scala specifications for more details.
```
jvican: Can you roughly explain what the current rules for infix operations are? A little bit of context will help the reader understand what you're proposing.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+| Date          | Version          |
+|---------------|------------------|
+| Feb 7th 2017  | Initial Draft    |
+
+---
+## Introduction
+Currently scala allows symbol operators (`-`, `*`, `~~>`, etc.) for both type names and definition names.
+Unfortunately, there is a 'surprise' element since the two differ in behaviour:
+
+* **Infix operator precedence and associativity**:
```
jvican: Should we make these subtitles instead of bold text?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+  type N1 = -N //Not working
+  type N2 = ~N //Not working
+  type N3 = !N //Not working
+  type N4 = +N //Not working
+}
+```
+
+---
+## Proposal
+The proposal is split into two; type infix precedence, and prefix unary types. Note to the SIP committee: It might be better to vote on the two parts separately.
```
jvican: It looks like the two features that you propose are related. However, I wonder if it would be better that you submit two different proposals. As the SIP Committee Lead, I would suggest that we do it this way because it's conceptually simpler -- if one is accepted and the other one is not, we'll be able to mark one of the documents in our official website as approved. This way, it wouldn't be neither approved nor rejected and it would be confusing to keep track of it.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+(Author's note: I can copy-paste the specification and modify it, if it so required)
+### Proposal, Part 2: Prefix unary types
+Add prefix types, exactly as specified for prefix expression. 
+(Author's note: I can copy-paste the specification and modify it, if it so required)
+
+---
+
+## Motivating examples
+#### Dotty infix type similarity
+Dotty infix type associativity and precedence seem to be the same as expressions (Author's note: I have seen no documentation of this, but checked the implementation for a simple example `implicitly[(N1 + (N2 / N3) + N4) =:= (N1 + N2 / N3 + N4)]`).
```
jvican: Can you try if the code snippet used as an example at the beginning works in Dotty?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+  implicit def apply[L](implicit check : Require[L > 0]) : MyVec[L] = new MyVec[L]()
+}
+val myVec : MyVec[10] = MyVec[4 + 1].doubleSize
+val myBadVec = MyVec[-1] //fails compilation, as required
+```  
+We currently loose some of the intuitive appeal due to the precedence issue:
+```scala
+val works : 1 + (2 * 3) + 4 = 11
+val fails : 1 + 2 * 3 + 4 = 11 //left associative:(((1+2)*3)+4))) = 13
+```
```
jvican: Can you expand more on this motivation? Also, can you rename the title `Singleton ops library` and use instead Motivation? I think this use case is valid, but I'm a little bit worried that it can be read as "we need this functionality only for this library". I think there's a more general, overarching philosophical purpose to the change you propose, since essentially you're trying to make terms and types behave equally regarding operation precedence. For a functional language like Scala, that argument would read better :smile:.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Developer issues
+The following stackoverflow question demonstrate developers are 'surprised' by the difference in infix precedence.
```
jvican: Can you elaborate a little bit on what they are surprised? A little bit of context here would be super useful.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+E.g.
+```scala
+trait Negate[A]
+trait Positive[A]
+type unary_-[A] = Negate[A] 
+type unary_+[A] = Positive[A] 
+trait Contravariant[B, -A <: -B] //contravariant A subtype upper-bounded by Negate[B]
+trait Covariant[B, +A <: +B] //covariant A subtype upper-bounded by Positive[B]
+```
+(Author's note: it seem very unlikely that such feature interaction will occur)
```
jvican: Have you explored this feature interaction? Perhaps you can have a look at the parser grammar and "prove" that it's not possible for it to happen?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+Note: It is not possible to annotate a positive literal type in Scala (checked both in TLS and Dotty):
+```scala
+val a : 42 = +42 //works
+val b : -42 = -42 //works
+val c : +42 = 42 //error: ';' expected but integer literal found 
+```
+This means that if unary prefix types are added, then `+42` will be a type expansion of `unary_+[42]`.
+
+#### Scala meta
+Open question how this SIP affects `scala-meta`.
```
jvican: The parser would need to be changed, but I think this section can go away. The trailing commas SIP doesn't mention Scala Meta (IIRC) and it's a detail not relevant to discuss the proposal in depth (any change to parser rules needs changes in tooling).

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+---
+
+## Backward Compatibility
+Changing infix type associativity and precedence affects code that uses type operations and conforms to the current specification.
```
jvican: FTR, it would be very interesting to see how often type alias precedence is used. If you really care about this (and want to enrich the proposal with high-quality data that helps the Committee decide), you can explore it using Scala Meta. Run it it common Scala projects and give us some stats (note that this is optional :wink).

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+---
+
+### Extended proposal alternative
+It is possible to extend this proposal and allow the developer to annotate the expected associativity and precedence per operation. 
+(Author's note: I personally don't like this, but if such a solution is better for the community, then I will gladly modify this SIP to reflect that.)
+See the following [Typelevel Scala issue](https://github.com/typelevel/scala/issues/69) for the suggestion.
+
+### Other languages
+Would love some help to complete what happens in different programming languages.
```
jvican: Can we put this in the description instead of the document? :smile:

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Discussions
```
jvican: Can we make more explicit that any comment on this proposal should go to the Discourse thread? Something like: "If you're interested in discussing this proposal, head over to [this]() Scala Contributors thread and let me know what you think".

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,197 @@
...
+[Scala Contributors](https://contributors.scala-lang.org/t/sip-nn-make-infix-type-alias-precedence-like-expression-operator-precedence/471)
+
```
jvican: Can we put this into "Bibliography"?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Will gladly do so. What do you mean by 'subtitles'? I'm not that familiar with MD.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Is it possible to do the following? Not to separate them, but if the committee approves one but not the other then I will split the SIP into two to allow marking them separately.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Yes I have. I verified it works. Maybe it's not clear from the context. I had to modify to example to a `/` from `*` because dotty currently has a bug https://github.com/lampepfl/dotty/issues/1961

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: This feature interaction does not collide. It may be confusing for a developer to see `-` and `+` prefix symbols in different context.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: OK. Will remove the 'Scala meta' subsection.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Moving notes to commit description.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: OK. Gave a short description and provided an example. I can place a screenshot of the entire section.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: >  can you rename the title Singleton ops library and use instead Motivation? 

It's a subsection within the Motivation section. Is that not enough?

> I think there's a more general, overarching philosophical purpose to the change you propose, since essentially you're trying to make terms and types behave equally regarding operation precedence. For a functional language like Scala, that argument would read better

OK, will modify text to reflect that.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Done

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Added at the top.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Done

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,187 @@
+---
+layout: sip
+disqus: true
+title: SIP-NN - Make types behave like expressions
```
DarkDimius: I think the current title is too general, it's not clear what is this sip about if you only read the title.
May I suggest:
`Types have precedence and associativity rules as expressions`

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
jvican: Something like '###' would be enough :smile:.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: The proposal is more than infix types.
`Match infix & prefix types to meet expression rules` is better?

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
jvican: Let's do it this way, yes. We'll see how this experiment goes, we haven't actually had this situation happened yet.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
jvican: Yes, good enough.

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
jvican: I agree with @DarkDimius that a better name is desirable -- this one makes you wonder: "Behave like expression in regard to what?". I think Dmitry's suggestion is good, I would just add 'Make' at the beginning of the sentence: `Make types have precedence and associativity rules as expressions`. I'm not sure whether 'like' would be more appropriate in that case, maybe you can choose @soronpo, my non-native English is not that good.

jvican: Yeah, I think that's way better @soronpo, let's go with that name!

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Well you did approve meta while requesting it to be split into two SIP. In that case, though, the condition was that they both were still coupled together. In this SIP I want both parts to be approved, but I understand that changing infix precedence is much more likely to be approved than adding prefix types.

jvican:

jvican:
Hi @soronpo, are you sure you have rebased from latest master? Can you try again? Drone has worked for @DarkDimius PRs, but something is off here because Drone hasn't even recognised your latest commit (and this happens because `.drone.yml` is not there).

soronpo:
@jvican I screwed something in the template, because this looks awful at http://docs.scala-lang.org/sips/pending/make-types-behave-like-expressions.html 
Any suggestion of what I should fix? I copied mine from http://docs.scala-lang.org/sips/pending/refer-other-arguments-in-args.html which looks the same.

jvican:
It probably has to do with spaces, can you check with the template of the `binaryCompatible` SIP?

soronpo:
Perhaps #687 will resolve this. I compared with `binaryCompatible` and it had an extra annotation for code segments

jvican:
I think it's the good way to fix it , thanks! 👍

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
```diff
@@ -0,0 +1,189 @@
...
+  type +[N1, N2] = Plus[N1, N2]
+  type /[N1, N2] = Div[N1, N2]
+  trait N1
+  trait N2
+  trait N3
+  trait N4
+  //Error!
+  //Left  expands to Plus[Plus[N1,Div[N2,N3]],N4] (Surprising)
+  //Right expands to Plus[Div[Plus[N1,N2],N3],N4]
+  implicitly[(N1 + N2 / N3 + N4) =:= (N1 + (N2 / N3) + N4)]
```
TomasMikula: I think "Left" and "Right" are mixed up here.

TomasMikula:
👍  for this SIP.

Here's perhaps a more concise demonstration of the problem:

```scala
type +[A, B] = Either[A, B]
type *[A, B] = (A, B)
type O = Nothing
type I = Unit

            1 + 1 * 0  ==  (1 + (1 * 0))  // true
implicitly[(I + I * O) =:= (I + (I * O))] // error: Cannot prove that I + I * O =:= I + (I * O).
```

sips/pending/_posts/2017-02-07-make-types-behave-like-expressions.md
soronpo: Thanks. Fixed that in https://github.com/scala/scala.github.com/pull/855

soronpo:
Updated PR in https://github.com/scala/scala.github.com/pull/855

PR #675, opened 2017-02-07
many small tweaks to the Basics page

SethTisue:
all of these changes are small, don't involve changing ordering,
and are hopefully not controversial.

### DIFF
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -14,9 +14,9 @@ In this page, we will cover basics of Scala.
 
 ## Trying Scala in the Browser
 
-You can run Scala in your browser with [ScalaFiddle](https://scalafiddle.io).
+You can run Scala in your browser with ScalaFiddle.
 
-1. Go to https://scalafiddle.io.
+1. Go to [https://scalafiddle.io](https://scalafiddle.io).
 2. Paste `println("Hello, world!")` in the left pane.
 3. Hit "Run" button. Output appears in the right pane.
 
@@ -41,7 +41,7 @@ You can name results of expressions with the `val` keyword.
 
 ```tut
 val x = 1 + 1
-println(x * x) // 4
+println(x) // 2
 ```
 
 Named results, such as `x` here, are called values. Referencing
@@ -91,7 +91,7 @@ println({
 
 ## Functions
 
-Functions are expressions that take parameters and compute a result.
+Functions are expressions that take parameters.
 
 You can define a function that returns a given integer plus one:
 
--- a/tutorials/tour/basics.md
+++ b/tutorials/tour/basics.md
@@ -14,13 +14,13 @@ In this page, we will cover basics of Scala.
 
 ## Trying Scala in the Browser
 
-You can run Scala in your browser by using [ScalaFiddle](https://scalafiddle.io).
+You can run Scala in your browser with [ScalaFiddle](https://scalafiddle.io).
 
 1. Go to https://scalafiddle.io.
-2. Copy and paste `println("Hello, world!")` to the left pane.
-3. Hit "Run" button and see it prints "Hello, world!" on the right pane.
+2. Paste `println("Hello, world!")` in the left pane.
+3. Hit "Run" button. Output appears in the right pane.
 
-It is a perfect way for anybody to experiment with a piece of Scala code anywhere, anytime!
+This is an easy, zero-setup way to experiment with pieces of Scala code.
 
 ## Expressions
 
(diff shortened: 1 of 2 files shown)


SethTisue:
review by @shogowada and community

SethTisue:
there are a few `tut` code blocks that don't compile. @tpolecat I wanted to use `tut:failed` for those, but it didn't work because the code blocks included valid entries as well as invalid entries, so tut complained that _everything_ didn't fail. is that working-as-designed...?

tpolecat:
@SethTisue "correct as implemented" … I think the surprisingly-named `:nofail` will do what you want.

SethTisue:
> I think the surprisingly-named :nofail will do what you want

yeah but it's deprecated. let's discuss further at https://github.com/tpolecat/tut/issues/141

tutorials/tour/basics.md
```diff
@@ -37,51 +37,52 @@ println("Hello," + " world!") // Hello, world!
...
 ### Values
 
-You can name results of expressions by using the `val` keyword.
+You can name results of expressions with the `val` keyword.
 
-```
+```tut
 val x = 1 + 1
-println(x) // 2
+println(x * x) // 4
```
shogowada: Personally, it looks like it's adding unnecessary complication only to show that values hold results of expressions.

What's the reason for this change?

tutorials/tour/basics.md
```diff
@@ -90,33 +91,33 @@ println({
 
 ## Functions
 
-Functions are expressions with parameters.
+Functions are expressions that take parameters and compute a result.
```
shogowada: Does "compute a result" redundant with "Functions are expressions"?

shogowada:
Just for my knowledge, what does `tut` do? I couldn't find anything on the internet or on contributing page.

tutorials/tour/basics.md
SethTisue: I wanted the example to have a _reason_ to use `val`, so we are illustrating not only the syntax, but also the use. With my change, introducing `x` now avoids repeating a computation (namely, the adding of 1 and 1).

Maybe I'm overthinking this. I don't have a strong feeling about it.

tutorials/tour/basics.md
SethTisue: Ummm... "expressions with parameters" just sounds funny to me. I would never say/write that. Does it mean "functions are values", or does it mean that the body of a function is an expression, or what?

I'm open to other suggestions than the one I've made, but I do think the original sentence is unclear and should be changed _somehow_.

SethTisue:
> what does tut do?

see #591 and #593 

oops, the readme should definitely explain this! I opened https://github.com/scala/scala.github.com/issues/677 on it

SethTisue:
Does anyone have any idea why Travis is failing? I just hit "restart" on the build a second time

tutorials/tour/basics.md
shogowada: I see.

How about leaving this line as is and adding another example, maybe something like below?

> You can use values in other expressions.
> ```
> val x = 1 + 1
> println(x * 2) // 4
> ```

tutorials/tour/basics.md
shogowada: Oh, I think rephrasing it to "Functions are expressions that take parameters" is an improvement.

By saying "Functions are expressions with parameters", I tried to communicate that functions are, in essence, just expressions with parameters added in its scope. But I think "that take parameters" communicates it clearer.

I just thought, maybe, "and compute a result" is redundant because expressions always compute results.

tutorials/tour/basics.md
```diff
@@ -14,21 +14,21 @@ In this page, we will cover basics of Scala.
 
 ## Trying Scala in the Browser
 
-You can run Scala in your browser by using [ScalaFiddle](https://scalafiddle.io).
+You can run Scala in your browser with [ScalaFiddle](https://scalafiddle.io).
 
 1. Go to https://scalafiddle.io.
```
shogowada: Can you please also link the URL? Automatic link isn't working on the website.
```
1. Go to [https://scalafiddle.io](https://scalafiddle.io).
```

tutorials/tour/basics.md
SethTisue: I'll just restore the original.

tutorials/tour/basics.md
SethTisue: I'll use your suggested form.

tutorials/tour/basics.md
SethTisue: done

tutorials/tour/basics.md
tpolecat: Functions *have* parameters and *take* arguments. It probably doesn't matter, but if you're keeping a running list of things that bug me there you go.

SethTisue:
I've responded to all reviewer feedback. not sure what the Travis flakiness is, but so far it always eventually passes if you hit rebuild a few times, so I'm going to go ahead and merge, since this page is getting attention right now since Heather tweeted (and scala_lang retweeted) about it https://twitter.com/heathercmiller/status/829653981089714176

tutorials/tour/basics.md
som-snytt: In a function application, are arguments applied to a function or the function to the arguments? Params and args can also be values or types. Functions also reject args when an application fails to typecheck.

som-snytt: For now, I will just think of methods as something like functions.

PR #681, opened 2017-02-12
Higher

travis032654:


tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
 num: 8
 next-page: nested-functions
 previous-page: anonymous-function-syntax
+assumed-knowledge: sequence-comprehensions
 ---
 
-Scala allows the definition of higher-order functions. These are functions that _take other functions as parameters_, or whose _result is a function_. Here is a function `apply` which takes another function `f` and a value `v` and applies function `f` to `v`:
+Higher order functions take other functions as parameters or return a function as
+a result. This is possible because functions are first-class objects in Scala.
```
julienrf: I would say “first-class values” instead of “objects”.

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
 
 ```tut
-def apply(f: Int => String, v: Int) = f(v)
+val salaries = Seq(20000, 70000, 40000)
+val newSalaries = salaries.map(_ * 2)
 ```
+The Scala compiler already knows the type of the parameters (a single Int) that function argument
+for `map` needs. Therefore you just need to provide the right side of the function `doubleSalary`. The only
+caveat is that you need to use `_` in place of a parameter name (it was `x` in
+the previous example).
```
julienrf: Well, you don’t *have to* use `_`. You could also write `salaries.map(x => x * 2)`.

Actually, I would first use `salaries.map(x => x * 2)`, and then later explain that there is a shorter notation with `_ * 2`.

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
-object FunTest extends App {
-  def apply(f: Int => String, v: Int) = f(v)
-  val decorator = new Decorator("[", "]")
-  println(apply(decorator.layout, 7))
+  def getForecastInFahrenheit() = temperatures.map(convertCtoF) // <-- passing the method convertCtoF
 }
 ```
- 
-Execution yields the output:
+Here the method `convertCtoF` is passed to getForecastInFahrenheit  This is possible because the compiler coerces `convertCtoF` to the function `_ => convertCtoF(_)`.
```
julienrf: `_` has a lot of meanings in Scala, whether it is placed on the left or on the right of an `=>` (or `=`), which might lead to confusions… Your example would not work because the first `_` defines a function with one ignored parameter, and the second one means “a function that takes one parameter and applies it to `convertCtoF`”.

But, again, I would not use `_` here and go with a simpler version: `x => convertCtoF(x)`. (Where `x` is actually a generated name which is guaranteed to be unique and to not clash with names that are already in scope)

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
+One reason to use higher-order functions is to reduce redundant code. Let's say you wanted functions  that could either search for files by directory, by regular expression, or by a substring. Without creating a higher-order function,
+it might look something like this:
 
+```tut
+  object FileFinder {
+    private def filesHere = (new java.io.File(".")).listFiles
+
+    def filesEndingWith(query: String) =
+      for (file <- filesHere; if file.getName.endsWith(query))
+        yield
```
julienrf: `yield file`

julienrf: Also, this example makes use of the `for yield` syntax, which may not be familiar to the reader.

What about staying in the context of salaries and define the following methods?

~~~ scala
def smallPromotion(salaries: List[Double]): List[Double] =
  salaries.map(salary => salary * 1.1)

def greatPromotion(salaries: List[Double]): List[Double] =
  salaries.map(salary => salary * math.log(salary))
~~~

Which can then be generalized as follows:

~~~ scala
def promotion(salaries: List[Double], promotionFunction: Double => Double): List[Double] =
  salaries.map(promotionFunction)
~~~

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
-class Decorator(left: String, right: String) {
-  def layout[A](x: A) = left + x.toString() + right
-}
+  private def convertCtoF(temp: Double) = temp * 1.8 + 32
 
-object FunTest extends App {
-  def apply(f: Int => String, v: Int) = f(v)
-  val decorator = new Decorator("[", "]")
-  println(apply(decorator.layout, 7))
+  def getForecastInFahrenheit() = temperatures.map(convertCtoF) // <-- passing the method convertCtoF
```
julienrf: I think you should always add the return type of your method definitions.

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
+a result. This is possible because functions are first-class objects in Scala.
+One of the most common examples is the higher-order
+function `map` which is available for collections in Scala.
+```tut
+val salaries = Seq(20000, 70000, 40000)
+val doubleSalary = (x: Int) => x * 2
+val newSalaries = salaries.map(doubleSalary) // List(40000, 140000, 80000)
+```
+In this case, the function `doubleSalary` gets applied to each element in the
+list of salaries. A more idiomatic way to write the same piece of code would be
```
julienrf: Maybe you should explain a little bit more what the `x => …` syntax means?

tutorials/tour/higher-order-functions.md
```diff
@@ -8,34 +8,112 @@ tutorial: scala-tour
...
+​
+val endpoint = "users"
+val query = "id=1"
+val url = getURL(endpoint, query) // "https://www.example.com/users?id=1": String
+​
 ```
 
-In this example, the method `decorator.layout` is coerced automatically to a value of type `Int => String` as required by method `apply`. Please note that method `decorator.layout` is a _polymorphic method_ (i.e. it abstracts over some of its signature types) and the Scala compiler has to instantiate its method type first appropriately.
+Notice the return type of urlBuilder `(String, String) => String`. This means that
+the returned anonymous function takes two Strings and returns a String.
```
julienrf: Maybe you should emphasize that the syntax at the type level is homogeneous with the syntax at the value level.

tutorials/tour/higher-order-functions.md
julienrf: Also, a convention in Scala is to put parens to methods that take no parameter only if this method actually performs side-effects. Since `getForecastInFahrenheit` is pure, you should remove the parens to its definition. Also, I would omit the `get` prefix, which is not common in Scala: `def forecastInFahrenheit: Double`.

jvican:

mpociecha:
FYI you don't need to close PRs and create new ones. It's possible to push once again with -f flag. HTH

travis032654:
Unfortunately I messed up the branching and the pull requests were including stuff from other branches so I started from scratch to clean it up.

travis032654:
I'll use the push -f if i need to update a pr. Thanks for the tip.

PR #685, opened 2017-02-13
Add drone.yml and remove travis

jvican:
Travis, goodbye. Your last attempt of killing our beloved tut setup is
not nice. I'm sorry, this is how life is. We just change partner.

See urho3d/Urho3D#651 for related discussions
on memory leaks in Travis and Travis killing other processes that are
memory intensive as tut.


heathermiller:
I'm confused by the order of operations here. Why is there drone output about failed CI before this PR has been merged?

jvican:
Travis hasn't kicked out because it doesn't find a `travis.yml` file, and Drone has already this repository enabled (because I did enable it in the admin panel). If it finds `drone.yml` it will try to build it. I'm now choosing the correct image that has both java and the correct ruby version, and this will be good to go.

heathermiller:
There's no Travis output in this PR. Only broken drone output, claiming that coursier can't be found.

jvican:
Yes, there's no Travis output because Travis doesn't kick out any if there's no `travis.yml` file, and this commit removes it.
  
I'm working on fixing the drone output.

jvican:
@heathermiller This is finally ready for review, everything passes in Drone. I took the liberty to remove the coursier installation from the drone script and add coursier directly to the image we depend on.
  
As a data point, Drone builds everything in ~5 minutes where Travis did it in around ~10 (this includes the optimization that we merged some hours ago).

heathermiller:
LGTM, thanks @jvican!! 🎉

PR #698, opened 2017-02-15
new classes tour

travissarles:


tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
   override def toString: String =
-    "(" + x + ", " + y + ")"
+    s"($x, $y)"
 }
 ```
 
-Classes in Scala are parameterized with constructor arguments. The code above defines two constructor arguments, `x` and `y`; they are both visible in the whole body of the class.
+This `Point` class has four members: the variables `x` and `y` and the methods `move` and
+`toString`. Unlike many other languages, the primary constructor is in the class signature `(var x: Int, var y: Int)`. The `move` method takes two integer arguments but does not return a value (the return type `Unit` corresponds to `void` in Java-like languages). `toString`, on the other hand, does not take any parameters but returns a `String` value. Since `toString` overrides the inherited `toString` method, it is tagged with the `override` keyword.
+
```
julienrf: Before jumping to secondary constructors I would put an example of instantiation and member calls.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
-object Classes {
-  def main(args: Array[String]) {
-    val pt = new Point(1, 2)
-    println(pt)
-    pt.move(10, 10)
-    println(pt)
+class Point {
+  def this(x: Int, y: Int) {
+    this()
+    if (x > 100 || y > 100) println("Warning: Out of bounds")
```
julienrf: This example is confusing since the `x` and `y` parameters are eventually discarded.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
   }
 }
 ```
+There are few requirements for auxiliary constructors:
+* They must start with `def this`
+* The parameters must _not_ be declared `val` or `var`
```
julienrf: I would say “can _not_”

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
+```tut
+import scala.math._
+
+class Point(var x: Double, var y: Double) {
+  private def distanceFromOrigin = sqrt(x * x + y * y)
+
+  override def toString: String =
+    s"Point: ($x, $y), Distance from origin: $distanceFromOrigin"
+  }
+```
```
julienrf: Minor remark: usually private members are used to hide state from the outside. In your example that’s the opposite situation: private members are pure, while public members `x` and `y` are vars.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
+
+You can also simply nest private methods inside of the public ones
+that use them:
```
julienrf: I think this is a general point related to scopes more than classes. I’m not sure it’s best place is in this page.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,105 @@ categories: tour
...
+```
+
+Primary constructor parameters with `val` and `var` are public. However, because `val`s are immutable, you can't write the following.
+```
+class Point(val x: Int, val y: Int)
+val point = new Point(1, 2)
+val.x = 3  // <-- does not compile
+```
 
+Parameters without `val` or `var` are private values, visible only within the class and its subclasses.
```
julienrf: Actually they are not visible from subclasses.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
 next-page: traits
 previous-page: unified-types
+topics: classes
+assumed-knowledge: no-return-keyword, type-declaration-syntax, string-interpolation, procedures
 ---
 
-Classes in Scala are static templates that can be instantiated into many objects at runtime.
-Here is a class definition which defines a class `Point`:
+Classes in Scala are blueprints for creating objects. They can contain methods,
+values, and variables, which are collectively called _members_.
```
tpolecat: They can also contain types, objects, traits, and classes which are also members.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
+
+## Defining a class
+A minimal class definition is simply the keyword `class` and
+an identifier:
```
tpolecat: It's important to note that class names should be capitalized because it's a universal convention and it has a downstream effect on pattern-matching.

*A minimal class definition is simply the keyword class and an identifier. Class names should be capitalized.*

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
+```tut
+class User
+
+val user1 = new User
+```
+The keyword `new` is used to create an instance of the class. `User` has an implicit constructor which takes no arguments because no constructor was defined. However, you'll often want a constructor and class body. Here is an example class definition for a point:
```
tpolecat: I would say "default" rather than "implicit" here to avoid confusion with `implicit` which is an unrelated concept.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
 }
+
+val point1 = new Point(2, 3)
+point1.x  // 2
+println(point1)  // prints (x, y)
 ```
 
-Classes in Scala are parameterized with constructor arguments. The code above defines two constructor arguments, `x` and `y`; they are both visible in the whole body of the class.
+This `Point` class has four members: the variables `x` and `y` and the methods `move` and
+`toString`. Unlike many other languages, the primary constructor is in the class signature `(var x: Int, var y: Int)`. The `move` method takes two integer arguments but does not return a value (the return type `Unit` corresponds to `void` in Java-like languages). `toString`, on the other hand, does not take any parameters but returns a `String` value. Since `toString` overrides the inherited `toString` method, it is tagged with the `override` keyword.
```
tpolecat: For "but does not return a value (the return type `Unit` corresponds to `void` in Java-like languages)" I suggest

*and returns the Unit value `()`, which carries no information. This corresponds roughly with `void` in Java-like languages.*

tpolecat: > toString overrides the inherited toString method

Inherited from what? We haven't talked about subclassing. Do we need to do this here?

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
+In this version of the `Point` class, `x` and `y` have the default value `0` so no arguments are required. However, because the constructor reads arguments left to right, if you just wanted to pass in a `y` value, you would need to name the parameter.
+```
+class Point(var x: Int = 0, var y: Int = 0)
+val point2 = new Point(y=2)
+println(point2.y)  // prints 2
+```
+
+This is also a good practice to enhance clarity.
+
+## Member scope
```
tpolecat: I suggest renaming this section something like **Private Members and Getter/Setter Syntax**.

tutorials/tour/_posts/2017-02-13-classes.md
travis032654: Perhaps we should sprinkle naming conventions throughout the tour. It's easier than looking at a big naming-convention document, though both are important.

tutorials/tour/_posts/2017-02-13-classes.md
travis032654: I was going to add this later in a "nesting" tutorial but to be accurate, we need to mention it now and say "more on that later".

tutorials/tour/_posts/2017-02-13-classes.md
julienrf: I can maybe just add a link to the “unified types” page.

tutorials/tour/_posts/2017-02-13-classes.md
```diff
@@ -9,46 +9,103 @@ categories: tour
...
+point1.y = 101 // prints the warning
+```
+In this version of the `Point` class, the data is stored in private variables `_x` and `_y`. There are methods `def x` and `def y` for accessing the private data. `def x_=` and `def y_=` are for validating and setting the value of `_x` and `_y`. Notice the special syntax for the setters: the method has `_=` appended to the identifier of the getter and the parameters come after.
 
-Here is the output of the program:
+Primary constructor parameters with `val` and `var` are public. However, because `val`s are immutable, you can't write the following.
+```
+class Point(val x: Int, val y: Int)
+val point = new Point(1, 2)
+val.x = 3  // <-- does not compile
```
SethTisue: I think you mean `point.x`, not `val.x`. 

If it's not appropriate to introduce `copy` here, maybe at least add a note that links forward to where `copy` will be introduced?

travis032654:
@heathermiller ready to merge

PR #699, opened 2017-02-15
Rewrote higher-order-functions

travissarles:


tutorials/tour/_posts/2017-02-13-higher-order-functions.md
```diff
@@ -9,34 +9,112 @@ categories: tour
...
 next-page: nested-functions
 previous-page: anonymous-function-syntax
+assumed-knowledge: sequence-comprehensions, anonymous-function-syntax
 ---
 
-Scala allows the definition of higher-order functions. These are functions that _take other functions as parameters_, or whose _result is a function_. Here is a function `apply` which takes another function `f` and a value `v` and applies function `f` to `v`:
-
+Higher order functions take other functions as parameters or return a function as
+a result. This is possible because functions are first-class values in Scala.
+One of the most common examples is the higher-order
```
julienrf: example~s~ (without “s”).
Or, I would just say “A common example …”

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
```diff
@@ -9,34 +9,112 @@ categories: tour
...
 ```tut
-class Decorator(left: String, right: String) {
-  def layout[A](x: A) = left + x.toString() + right
+val salaries = Seq(20000, 70000, 40000)
+val newSalaries = salaries.map(_ * 2)
+```
+Since the Scala compiler already knows the type of the parameters (a single Int),
+ you just need to provide the right side of the function. The only
+caveat is that you need to use `_` in place of a parameter name (it was `x` in
+the previous example).
```
julienrf: Why is this a “caveat”.

Maybe a more intuitive way to describe the `_ * 2` syntax is to say that `_` means “a hole”, thus the expression `_ * 2` means “something (that has to be filled in), multiplied by two”

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
```diff
@@ -9,34 +9,112 @@ categories: tour
...
+    promotion(salaries, salary => salary * math.log(salary))
+
+  def hugePromotion(salaries: List[Double]): List[Double] =
+    promotion(salaries, salary => salary * salary)
+}
 ```
-[7]
+
+The new function, `promotion`, takes the salaries plus a function of type `Double => Double`
+(i.e. a function that takes a Double and returns a Double) and returns the product.
```
julienrf: What do you mean by “returns the product”?

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
```diff
@@ -9,34 +9,112 @@ categories: tour
...
- 
-Execution yields the output:
 
+Notice how each of the three methods vary only by the multiplication factor. To simplify,
+you can extract the repeated code into a higher-order function like so:
+
+```tut
+object SalaryRaiser {
+
+  private def promotion(salaries: List[Double], promotionFunction: Double => Double): List[Double] =
```
Ichoran: This is a poor example because the resulting code is more verbose and harder to follow.  Abstraction in examples should lead to a very clear win to help motivate the feature.

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
```diff
@@ -9,34 +9,112 @@ categories: tour
...
+The new function, `promotion`, takes the salaries plus a function of type `Double => Double`
+(i.e. a function that takes a Double and returns a Double) and returns the product.
+
+## Functions that return functions
+
+There are certain cases where you want to generate a function. Here's an example
+of a method that returns a function.
+
+```tut
+def urlBuilder(ssl: Boolean, domainName: String): (String, String) => String = {
```
Ichoran: This is not a good example because you would almost never see this.  Instead, you'd just write `urlBuilder(ssl: Boolean, domainName: String): String` and get the function above with `urlBuilder _`.

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
travis032654: I've heard the hole metaphor used before and I don't think it's used outside of Scala documentation.

travis032654: But yes, it should be reworded.

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
travis032654: it returns the product of the current salary multiplied by a factor. I guess it could be clearer to say it returns the new salary.

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
travis032654: @Ichoran I agree with that point. Can you think of a better example?

Ichoran: Maybe do something with `case class Point(x: Double, y: Double)` and `def closestBy(points: List[Point], target: Point, metric: (Point, Point) => Double`?  Then you can have `closestByDistance(points: List[Point], target: Point) = closestBy(points, target, (p, q) => (p.x - q.x)*(p.x - q.x) + (p.y - q.y)*(p.y - q.y)` and for `closestByBlock` `math.abs(p.x - q.x) + math.abs(p.y-q.y)` and so on.

Whatever you choose, the key should be that the work of the common function shouldn't be *completely* trivial (i.e. it should be more than just one other method call!).

julienrf: @Ichoran Yeah, nice example! @travis032654 Maybe we can also mention that this is an application of the `Strategy` design pattern, do you think that would help enterprise programmer to get it?

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
arigoldx: Hmm.. maybe I'm missing something but just so it's clear `One of the most common examples` *with* the `s` is the correct English.

That said, you can certainly say `A common example...`

tutorials/tour/_posts/2017-02-13-higher-order-functions.md
gdemelo: Just use "difference" instead of "caveat".
If you want to make it more clear, you could also add Julien's explanation but call it a "slot" that has to be filled.

SethTisue:
There was some good feedback on this that never got addressed, but it's still an improvement, as-is, so I figure hitting "merge" is better than just letting it languish forever. I invite anyone watching the repo to submit a followup PR with further improvements, whether suggested by reviewers here or otherwise.

PR #684, opened 2017-02-13
Fix #607: Improve build times drastically

jvican:
Per @dragos's suggestion, this commit implements a new approach to
generate table of contents that uses internally populated Jekyll data
structures to speed up listing and matching posts that are from the same
category.

As categories are only populated by posts and our current tutorials are
pages, this commits turns pages into posts and makes sure that their
permalinks correspond to the current ones.

Build times are now around two minutes, see the following profile
information from Jekyll:

```
jvican in /data/rw/code/scala/scala.github.com                                                                                                                 [11:02:40]
> $ bundle exec jekyll serve --profile                                                                                                           [±drone-integration ●●●]
Configuration file: /data/rw/code/scala/scala.github.com/_config.yml
Configuration file: /data/rw/code/scala/scala.github.com/_config.yml
            Source: /data/rw/code/scala/scala.github.com
       Destination: /data/rw/code/scala/scala.github.com/_site
 Incremental build: disabled. Enable with --incremental
      Generating...

Filename                                                           | Count |     Bytes |   Time
-------------------------------------------------------------------+-------+-----------+-------
_layouts/overview-large.html                                       |   164 | 26568.57K | 92.778
_includes/toc-large.txt                                            |   164 | 23156.00K | 88.934
_layouts/tutorial.html                                             |   199 |  3005.98K |  1.350
_layouts/overview.html                                             |    37 |  1190.86K |  0.953
_includes/tutorial-toc.txt                                         |   199 |   267.11K |  0.455
es/overviews/index.md                                              |     1 |    46.66K |  0.443
_includes/localized-overview-index.txt                             |     1 |    46.49K |  0.443
_includes/tutorial-tour-list.txt                                   |   200 |   244.89K |  0.432
_includes/header.txt                                               |   439 |  2114.25K |  0.347
_layouts/cheatsheet.html                                           |     6 |   225.08K |  0.217
_includes/cheatsheet-sidebar.txt                                   |     6 |    42.70K |  0.207
style/index.md                                                     |     1 |    68.00K |  0.184
_includes/footer.txt                                               |   449 |   855.91K |  0.175
tutorials/index.md                                                 |     1 |    54.34K |  0.168
_includes/footerbar.txt                                            |   458 |   767.51K |  0.116
_includes/topbar.txt                                               |   419 |   822.04K |  0.110
_includes/pager.txt                                                |   363 |    27.10K |  0.079
_layouts/sip.html                                                  |    29 |   626.83K |  0.079
[...]

PR #706, opened 2017-02-20
Fix pt-br Tour of Scala

adamwy:
Makes pt-br translation show up in Tour of Scala ToC which became broken in #684.


SethTisue:
thank you!

PR #707, opened 2017-02-22
Rewrote operators

travissarles:


tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
...
+}
+
+val vector1 = new Vector(1.0, 1.0)
+val vector2 = new Vector(2.0, 2.0)
+
+val vector3 = vector1 + vector2
+vector3.x  // 3.0
+vector3.y  // 3.0
+```
+The class vector has a method `+` which we used to add vector1 and vector2. When the method has two or more parameters, use a tuple:
```
Ichoran: This isn't actually a tuple, just a consequence of point-free notation: `a.b(c)` becomes `a b c`, but `a.b(c, d)` becomes `a b (c, d)`.

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -28,11 +60,4 @@ def not(x: MyBool) = x.negate
...
 
-This helps to make the definition of `xor` more readable.
-
-Here is the corresponding code in a more traditional object-oriented programming language syntax:
-
-```tut
-def not(x: MyBool) = x.negate
-def xor(x: MyBool, y: MyBool) = x.or(y).and(x.and(y).negate)
-```
+This helps to make the definition of `xor` more readable.If the method takes more than one argument, use a tuple:
```
Ichoran: You seem to have stray text on the end here.

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
...
+```
+
+However, it more closely resembles math if you write it as an infix operator:
+```
+10 + 1
+```
+
+You can also give `+` specific meaning in your class:
+```tut
+class Vector(val x: Double, val y: Double) {
```
Ichoran: Why don't you call this `Vec` to avoid collision with `scala.collection.immutable.Vector` (whose name is really unfortunate for anyone thinking about mathematical vectors)?

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
 next-page: automatic-closures
 previous-page: local-type-inference
 ---
+In Scala, operators are methods. For example, `+` can be called with dot-notation:
+```
+10.+(1)
+```
+
+However, it more closely resembles math if you write it as an infix operator:
```
Ichoran: I don't think "closely resembles math" is quite what you mean.  It *is* math.  It more closely resembles classical mathematical notation for addition, but that's quite a mouthful.

Maybe something like, "However, the expression is easier to recognize when written in a familiar form with `+` as an operator:".  (I don't think that wording is ideal either, actually.)

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
 next-page: automatic-closures
 previous-page: local-type-inference
 ---
+In Scala, operators are methods. For example, `+` can be called with dot-notation:
```
Ichoran: You never actually say that methods can be used as operators!  This is **the most important** piece of information in this section!  Please add it again!

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
...
+```
+10.+(1)
+```
+
+However, it more closely resembles math if you write it as an infix operator:
+```
+10 + 1
+```
+
+You can also give `+` specific meaning in your class:
```
julienrf: I would say “You can define your own operators”

tutorials/tour/_posts/2017-02-13-operators.md
julienrf: I don’t think it is worth mentioning this special case. In practice, infix notation is mostly used with methods that take one parameter.

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,40 @@ num: 30
...
+class Rectangle(val x: Double, val y: Double) {
+  def changeSize(dx: Int, dy: Int): Rectangle = new Rectangle(x + dx, y + dy)
+}
+
+val rect1 = new Rectangle(1, 2)
+val rect2 = rect1 changeSize (2, 2)  // equivalent to rect1.changeSize(2, 2)
+```
 
-Any method which takes a single parameter can be used as an *infix operator* in Scala. Here is the definition of class `MyBool` which includes methods `and` and `or`:
+Using parentheses, you can build up complex expressions with readable syntax. Here is the definition of class `MyBool` which includes methods `and` and `or`:
```
julienrf: You could maybe stay in the same `Vec` context and use `add` instead of `+` and say that all methods (not just those with symbolic names) can be used with the infix notation.

julienrf:
Also, what do you think of giving some words, at the end, about associativity and precedence resolution?

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,31 @@ num: 30
...
+
+However, it's easier to read as an infix operator:
+```
+10 + 1
+```
+
+## Defining and using operators
+You can use any legal identifier as an operator. This includes a name like `add` or a symbol(s) like `+`.
+```tut
+class Vec(val x: Double, val y: Double) {
```
MasseGuillaume: `case class`

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -10,8 +10,31 @@ num: 30
...
+  def +(that: Vec) = new Vec(this.x + that.x, this.y + that.y)
+}
+
+val vector1 = new Vec(1.0, 1.0)
```
MasseGuillaume: just `Vec`

tutorials/tour/_posts/2017-02-13-operators.md
```diff
@@ -30,9 +53,26 @@ def xor(x: MyBool, y: MyBool) = (x or y) and not(x and y)
...
+```
+This applies to functions you define. For example, the following expression:
+```
+a + b ^? c ?^ d less a ==> b | c
+```
+Is equivalent to
+```
+((a + b) ^? (c ?^ d)) less ((a ==> b) | c)
 ```
+`?^` has the highest precedence because it starts with the character `?`. `+` has the second highest precedence, followed by `^?`, `==>`, `|`, and `less`.
```
MasseGuillaume: we are missing the suffix `:`

for example:

```
case class Foo(v: Int) {
  def bar:(x: Int): = Foo(v + x)
}

1 bar: Foo(2)
```

is equivalent to `Foo(2).bar:(1)`

a notable example is `val a = 1 :: 2 :: 3 :: Nil`

PR #710, opened 2017-02-22
Corrects a mistake in the article on implicit conversions

codeZeilen:
This is a very minor correction to address the issue discussed on disqus. It makes the example easier to understand.


SethTisue:
thank you!

PR #711, opened 2017-02-22
Correct description note of Iterator sameElements

RemyHaemmerle:
Align the note in the method description with the corresponding note in
the API documentation see
http://www.scala-lang.org/api/2.11.8/index.html#scala.collection.Iterator@sameElements(that:Iterator[_]):Boolean.

The note is in the API since version 2.9.2


SethTisue:
thank you!

PR #713, opened 2017-02-24
Rewrote traits tour

travissarles:


tutorials/tour/_posts/2017-02-13-traits.md
```diff
@@ -9,46 +9,74 @@ categories: tour
 num: 5
 next-page: mixin-class-composition
 previous-page: classes
+assumed-knowledge: expressions, classes, generics, objects, companion-objects
 ---
 
-Similar to interfaces in Java, traits are used to define object types by specifying the signature of the supported methods. Like in Java 8, Scala allows traits to be partially implemented; i.e. it is possible to define default implementations for some methods. In contrast to classes, traits may not have constructor parameters.
-Here is an example:
- 
+Traits are used to share interfaces and fields between classes. They are similar Java 8's interfaces. Classes and objects can extends traits but traits cannot be instantiated.
```
SethTisue: similar -> similar to

can extends -> can extend

soronpo:
Thought (I have no solid opinion on this): Shouldn't examples in Scala Tour focus more on functional/immutable code, than imperative/mutable code? 

I think there should be an example that demonstrates inheritance from more than one trait (or else, why not just use an abstract class). 
I wonder if there should be a mention that traits cannot accept parameters. That of course changes in Dotty, and there currently is a SIP that may change that for Scalac as well.

SethTisue:
> I think there should be an example that demonstrates inheritance from more than one trait

the next section of the tour is http://docs.scala-lang.org/tutorials/tour/mixin-class-composition.html, but a PR adding a forward reference would be welcome. (I think it would even be reasonable to merge the two pages.)

I agree an immutable example would be preferable, but this PR doesn't move backwards on that, at least. a separate PR with a better example would be welcome

PR #716, opened 2017-02-27
style guide: revise sections on infix and postfix notation

SethTisue:
the postfix text is mostly the same, but now comes last and uses the
word "postfix" instead of "suffix".

the infix text is totally rewritten and now more strongly discourages
use of infix for alphabetic method names.  back when the style guide
was first written, I think there was more experimentation with, and
more acceptance of, such use of infix notation.  these days, not as
much.

I omitted the material about mixing dot and dotless notation -- the "a
b.(c) d" example.  I found that passage highly confusing and I've seen
others be confused by it as well.  I don't think it's important to
include.

review by @WadeWaldron and community

### DIFF
--- a/style/method-invocation.md
+++ b/style/method-invocation.md
@@ -54,19 +54,19 @@ readability and will make it much easier to understand at a glance the
 most basic operation of any given method. Resist the urge to omit
 parentheses simply to save two characters!
 
-### Suffix Notation
+## Postfix Notation
 
-Scala allows methods of arity-0 to be invoked using suffix notation:
+Scala allows methods that take no arguments to be invoked using postfix notation:
 
+    // recommended
     names.toList
 
-    // is the same as
-
-    names toList // Unsafe, don't use!
+    // discourage
+    names toList
 
 This style is unsafe, and should not be used.  Since semicolons are
 optional, the compiler will attempt to treat it as an infix method
-if it can, potentially taking a term from the next line.  
+if it can, potentially taking a term from the next line.
 
     names toList
     val answer = 42        // will not compile!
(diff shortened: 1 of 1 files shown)


SethTisue:
the diffs will be easier to review if you look at the two commits separately.

Ichoran:
Looks good!

dwijnand:
RIP. :-(

vpetro:
👍

SethTisue:
thanks, reviewers

SethTisue:
cc @djspiewak whose prose I think I recognize here :-)

som-snytt:
I thought the better-known names for `/:` and `:\` were `querulous` and `chagrin`, respectively.

djspiewak:
@SethTisue This is indeed my prose.  :-)  I agree that infix alphabetical function usage is now far less common than it used to be, but I still strongly believe it is the most consistent style.  Consider the following example:

```scala
xs map { a => a + a } filter { _ < 12 }

xs.map { a => a + a }.filter { _ < 12 }

xs.map(a => a + a).filter(_ < 12)
```

Let's just agree right now that the second form is absurd and shouldn't be encouraged.  If nothing else, "dotting" on the end of a lambda sometimes doesn't parse the way you would expect (e.g. the `pendingUntilFixed` method in specs2).

From a personal aesthetic standpoint, I strongly prefer the third option.  It's the most concise, feels the most "functional" (in that it treats lambdas as normal values), avoids the subjectively-ugly curly brace, scales well (you can just keep chaining functions without parentheses, regardless of arity), is amenable to explicit type parameters and currying, etc.  There's a ton of benefits!  Where it falls over though is if you swap the lambda passed to `map`:

```scala
xs map {
  case a => a + a
} filter { _ < 12 }

xs.map {
  case a => a + a
}.filter { _ < 12 }   // not even sure if this parses correctly

xs.map({
  case a => a + a
}).filter(_ < 12)
```

You see the problem.  I've never heard anyone give me a plausible solution to this issue.

There's a secondary problem here as well, which is what happens if you aren't switching to a partial function, but you've discovered that your lambda is now non-trivial and needs an intermediate value:

```scala
xs map { a => 
  val back = a + a
  back
} filter { _ < 12 }

xs.map({ a => 
  val back = a + a
  back
}).filter(_ < 12)
```

A more common pitfall here is to try to inject a `println(a);` prior to the `a + a`, which completely breaks if you use parentheses since the lambda will end at the `;`, rather than at the `)`.  Neither of these are issues for curly braces.

So every time I come back to this problem, I really *want* to use dot notation for alphabetical functions and parentheses for lambdas.  I *really* really want to.  But curly braces are unavoidable, both for partial functions and for multi-line lambdas, and they just don't play nicely with a paren- and dot-focused syntactic idiom.

For this reason, I still strongly advocate for an infix style on alphabetical functions.  At the very least alphabetical functions which may be used in conjunction with other alphabetical functions which take lambdas.

SethTisue:
> I thought the better-known names for /: and :\ were querulous and chagrin, respectively.

those were the eighth and ninth dwarves in one of Walt's early drafts, I think

@djspiewak your "not even sure if this parses correctly" does parse correctly and works fine (and reads fine too IMO), so I don't see the problem. dots and curlies are compatible, so wanting curlies doesn't mean losing dots

djspiewak:
> @djspiewak your "not even sure if this parses correctly" does parse correctly and works fine (and reads fine too IMO), so I don't see the problem. dots and curlies are compatible, so wanting curlies never means losing dots, afaict. and I've never seen a case where it was necessary or appropriate to resort to ({...})

Example from specs2:

```scala
"do the things" in {
  "stuff" must haveLength(5)
}.pendingUntilFixed
```

Play around with different associations of parentheses, dots, lack of dots, etc.  The target of the `pendingUntilFixed` dispatch is highly unintuitive.

SethTisue:
ah. well, ugh on these English-y DSLs in first place, to my personal taste

in a codebase where the use of the infix style for this DSL was considered standard, I would write your example as

```scala
("do the things" in {
  "stuff" must haveLength(5)
}).pendingUntilFixed
```

and now I understand the motivation behind the original text better. regardless, I think it's a niche enough issue that I'm comfortable omitting it.

som-snytt:
I don't know if [my understanding as a learner](http://stackoverflow.com/a/13873899/1296806) clarifies the earlier comment about braces, but I agree that `receiver op { body }.f` is hard to see.

To say more, `f` can take `(args)` or `{ stms ; res }`, and that made sense to me.

som-snytt:
In recent contributions, I have restored dots where I had to think twice about the scansion.

I started to think twice about it after the "Scala style" keynote from a couple of years ago, so it is actually a long process to change a way of seeing code. So leadership in setting the pace is appreciated.

I would expect backticked max is a tic, however.

Maybe backtick is easier on a Swiss keyboard. And in emacs. For vi, it is dangerously close to ESC, especially on a laptop with chiclet keys.

fommil:
I've encountered unexpected behaviour when using `{...}` curly notation when combined with `_` underscore. Is possible to introduce multi lines where the first line is evaluated only once, not on every evocation of the lambda (as one would expect). Yes, it was nasty horrible side-effecting actor code, but it was enough to scar me for life and I discourage any use of both together now. Underscore and parens, or use a curly!

Ichoran:
@djspiewak - All of my code is the second form, save for whitespace, and it's utterly readable and not the least bit absurd.

```scala
  myData.
    map{ x => someOperationOn(x) + someOtherOperationOn(x) }.
    filter{ y => foo(y) && !bar(y) }
```

The nice thing about this form is that it's infinitely indentable--just drop the parens down a line and keep going:

```scala
  myData.
    map {
      case x if small(x) => someOperationOn(x)
      case x => someOtherOperation(x)
    }.
    filter(_ < 12)
```

Whether or not this is recommended style (I don't care!), I don't instantly agree that it's absurd, or I wouldn't be doing it everywhere :P

non:
Yeah, a variant of the second style that we use frequently is something like this:

```scala
source
  .map { x =>
    ...
  }
  .filter { y =>
    ...
  }
  .grouped
```

Ichoran:
@non - I do a lot of development by-paste-in-REPL, and that form doesn't work there.

dwijnand:
@Ichoran You could wrap it in parens as you're pasting it in..

@fommil Good point. As an illustration:

```scala
scala> List(1, 2) map { println("hi"); _ + 1 }
hi
res6: List[Int] = List(2, 3)
```

vs

```scala
scala> List(1, 2) map { x => println("hi"); x + 1 }
hi
hi
res7: List[Int] = List(2, 3)
```

dwijnand:
My gripe with dot-notation is you can't use it for `match` or `yield`. And that it's more noisy that whitespace.

djspiewak:
@odersky 

> The one exception is that I always write method names infix when followed by {, because then they look like control structures. @djspiewak: It seems that takes care of all of your counter-examples?

Maybe I should just get used to what you're suggesting.  Technically, my current style is already inconsistent in the case of multiple parameter blocks or explicit type parameter instantiations (since those don't work with infix or postfix style).  What you're suggesting *is* inconsistent w.r.t. the dispatch "operator" (space vs dot), but perhaps it's less inconsistent overall.  And it does enable the more pleasing parenthetical style.

@fommil

> I've encountered unexpected behaviour when using {...} curly notation when combined with _ underscore. Is possible to introduce multi lines where the first line is evaluated only once, not on every evocation of the lambda (as one would expect). Yes, it was nasty horrible side-effecting actor code, but it was enough to scar me for life and I discourage any use of both together now. Underscore and parens, or use a curly with an explicit parameter!

I've had the opposite issue.  To wit:

```scala
xs.map(a => println(a); a + a)
```

I think I pointed this one out in the original comment.  Or, if you favor underscores:

```scala
xs.foldLeft(0)(println(_); _ + 1)
```

@Ichoran 

> All of my code is the second form, save for whitespace, and it's utterly readable and not the least bit absurd.

Maybe this is a taste thing, but I legitimately have a hard(er) time reading your snippet.  Especially with the unspaced open curlies and the trailing dots.

@non

> Yeah, a variant of the second style that we use frequently is something like this

I've used that style.  Sometimes you have to, because of the way newline inference special-cases leading dots.

@dwijnand 

> My gripe with dot-notation is you can't use it for match or yield. And that it's more noisy that whitespace.

I don't think of `yield` or `match` as being functions, so that works out pretty well for me.  :-)  So I would keep them infix for exactly the same reason that I would continue to put a space between the `if` and the `(` character.  Also, even if `match` *were* a function (scala trivia note: it used to be!), infix form would fall out of @odersky's proposed stylistic distinction.

PR #717, opened 2017-02-27
Rewrote pattern matching tour

travissarles:


tutorials/tour/_posts/2017-02-13-pattern-matching.md
```diff
@@ -10,38 +10,129 @@ num: 12
 
 next-page: singleton-objects
 previous-page: case-classes
+prerequisite-knowledge: case-classes, string-interpolation, subtyping
 ---
 
-Scala has a built-in general pattern matching mechanism. It allows to match on any sort of data with a first-match policy. 
-Here is a small example which shows how to match against an integer value:
+Pattern matching is a mechanism for checking a value for a pattern. It is a more powerful version of the `switch` statement in Java and it can likewise be used in place of a series of if/else statements.
```
SethTisue: suggested for the first sentence:

Pattern matching is a mechanism for checking a value against a pattern. A successful match can also deconstruct a value into its constituent parts.

tutorials/tour/_posts/2017-02-13-pattern-matching.md
```diff
@@ -10,38 +10,129 @@ num: 12
...
-object MatchTest1 extends App {
-  def matchTest(x: Int): String = x match {
-    case 1 => "one"
-    case 2 => "two"
-    case _ => "many"
-  }
-  println(matchTest(3))
+import scala.util.Random
+
+val x: Int = new Random().nextInt(10)
```
SethTisue: you don't need `new` here. `Random.nextInt(10)` works

tutorials/tour/_posts/2017-02-13-pattern-matching.md
```diff
@@ -10,38 +10,129 @@ num: 12
...
+println(showNotificationSpecial(someSms, SPECIAL_EMAIL, SPECIAL_NUMBER))
+println(showNotificationSpecial(someVoiceRecording, SPECIAL_EMAIL, SPECIAL_NUMBER))
+println(showNotificationSpecial(specialEmail, SPECIAL_EMAIL, SPECIAL_NUMBER))
+println(showNotificationSpecial(specialSms, SPECIAL_EMAIL, SPECIAL_NUMBER))
+```
+
+## Matching on type
+It is also possible to match on type.
+```tut
+def printType(x: Any): Unit = x match {
```
SethTisue: ten minutes til @tpolecat shows up and asks if we have to use an example based on `Any`.

I'm not really sure what the right way to handle this is. The feature exists, it definitely should be included in this section of the tour, and we don't want to say it's always bad, but we don't want to give the impression that matching on `Any` is normal, either. perhaps up above, before showing case class extraction, you could show something like:

```
notification match {
    case e: Email => "you've got mail: $e"
    case s: SMS => "someone texted you: $s"
    case v: VoiceRecording => "check your voicemail: $v"
}
```

and then quickly move on and say "but most of the time, you don't just want to know what type something is, you want to do something with the individual fields, too", and then show the unapply-based version?

leaving the existing progression alone but just coming up with a non-`Any` based example also seems like a possibility?

tpolecat: image

travis032654:
@heathermiller ready to merge

PR #718, opened 2017-03-01
Correct zh-cn 'future' section

ZhiruiLi:
Code examples of this section have some error, that code can’t
compiled, and can’t corresponding to text.


SethTisue:
thank you!

PR #719, opened 2017-03-02
Fix some mistakes in collections pages

Philippus:
this should fix some mistakes mentioned in:
http://docs.scala-lang.org/overviews/collections/sets
http://docs.scala-lang.org/overviews/collections/trait-iterable and
http://docs.scala-lang.org/overviews/collections/trait-iterable.html

(the 'Set' is not a 'PartialFunction' part has not been fixed in the jp/zh-cn versions


SethTisue:
thank you!

PR #723, opened 2017-03-04
variances-page: use T and S instead of A and B

Philippus:
like in the description and some styling fixes

mentioned in the comments of http://docs.scala-lang.org/tutorials/tour/variances.html


SethTisue:
someone was telling me recently that `A` and `B` are more idiomatic in Scala, and perhaps we should try to be consistent on this tour-wide. but regardless, this is an improvement. thank you!

som-snytt:
I also learned early to use `A` and `B` and find anything else jarring except `CC` in collections, or maybe `Elem`.

I haven't heard why that's idiomatic but not `f(x, y, z)`. With named and partial application of type params, the mores would change.

What the neutral `A` expresses is, as a `List[A]`, I don't care what I'm a list of. But if I'm a `Map[A, B]`, I might prefer `Map[K, V]` or similar.

PR #724, opened 2017-03-04
fixes a typo in the collections page

Philippus:
as mentioned here http://docs.scala-lang.org/overviews/collections/creating-collections-from-scratch.html


SethTisue:
sweet, thank you

PR #725, opened 2017-03-04
fix example on string-interpolation page

Philippus:
as mentioned here http://docs.scala-lang.org/overviews/core/string-interpolation.html


SethTisue:
thank you!

PR #727, opened 2017-03-09
Menu for korean is added.

zenos-lee:
The menu for korean is added to all documents in guides and overviews.
An overview index document is added.


SethTisue:
I don't think it makes sense to add the menu items unless or until the translations actually exist...?

zenos-lee:
Ok, I agree ~

2017-03-30 8:02 GMT+09:00 Seth Tisue <notifications@github.com>:

PR #730, opened 2017-03-14
Reworked default parameter tour

travissarles:


tutorials/tour/_posts/2017-02-13-default-parameter-values.md
```diff
@@ -9,65 +9,39 @@ categories: tour
...
-  public HashMap(int initialCapacity, float loadFactor);
-}
+log("System starting")  // prints INFO: System starting
+log("User not found", "WARNING")  // prints WARNING: User not found
 ```
 
-There's really only two constructors here; one that takes another map, and one that takes a capacity and load factor.  The third and fourth constructors are there to allow users of <code>HashMap</code> to create instances with the probably-good-for-most-cases defaults of both load factor and capacity.
-
-More problematic is that the values used for defaults are in both the Javadoc *and* the code.  Keeping this up to date is easily forgotten.  A typical pattern around this would be to add public constants whose values will show up in the Javadoc:
+The parameter `level` has a default value so it is optional. Where you might do overloaded methods in Java, you can use methods with optional parameters to achieve the same effect. However, ff the caller omits an argument, any following arguments must be named.
```
maasg: However, *if* the caller...

travis032654: Good catch.

travis032654:
I do combine default values with overrides in the second code example I think. Is that not what you're talking about?

PR #731, opened 2017-03-15
Final update to SIP-27 Trailing Commas

dwijnand:


dwijnand:
Part of the checklist for fixing scala/scala-dev#229.

dwijnand:
Thank you, Jorge.

PR #732, opened 2017-03-16
Rewrote currying tour

travissarles:


SethTisue:
I suggest renaming the whole section from "currying" to "multiple parameter lists", and never saying "curry" anywhere in the body text, either. Scala people don't normally talk about "currying".

I don't think you've chosen appropriate example code here — there is no good reason, that I can see anyway (and the text doesn't help me), to use multiple parameter lists for this function.

The old example code was unsatisfactory also.

I'd suggest looking for a commonly used method in the Scala standard library that has more than one parameter list. then show how use it, explain why it has multiple parameter lists, and then show how it could be defined. perhaps `foldLeft`?

jvican:
Maybe we can let this open so that a contributor can pick it up?

SethTisue:
Yes please, is there a volunteer who would like to get it over the finish line?

jvican:
As you took the other one, I'll take this one.

jvican:
As we got quite many volunteers, I'm gonna see if a volunteer dares to do this one! If there isn't one in two weeks, I'll take care of it :smile:.

SethTisue:
replaced by #986

PR #734, opened 2017-03-16
Rewrote anonymous function syntax tour

travissarles:


SethTisue:
Perhaps this section should be dropped from the tour. Is there a different section where you think the material could be incorporated?

travis032654:
Yeah, it's pretty light on content. Maybe it could be the intro for the higher-order functions tour because that's where they are most often used.

travis032654:
Actually I'll just put it in the basics tour.

Issue #733, opened 2017-03-16
Rewrite tour of Scala


travissarles:
I'm rewriting most of the tour of Scala. However some guides may not be as relevant like XML processing. Thoughts? @heathermiller 

[The tour](http://docs.scala-lang.org/tutorials/tour/tour-of-scala.html)
[More info on the rewrite](https://contributors.scala-lang.org/t/learning-scala-made-easy-ier/529)

**Current status:** 

- [x] **[Higher-order functions](https://github.com/scala/scala.github.com/pull/699)** ([old version](http://docs.scala-lang.org/tutorials/tour/higher-order-functions.html))
- [x] **[Singleton objects](https://github.com/scala/scala.github.com/issues/704)** ([old version](http://docs.scala-lang.org/tutorials/tour/singleton-objects.html))
- [x] **[Generic classes](https://github.com/scala/scala.github.com/issues/705)** ([old version](http://docs.scala-lang.org/tutorials/tour/generic-classes.html))
- [x] **[Operators](https://github.com/scala/scala.github.com/issues/707)** ([old version](http://docs.scala-lang.org/tutorials/tour/operators.html))
- [x] **[Unified Types](https://github.com/scala/scala.github.com/issues/708)** ([old version](http://docs.scala-lang.org/tutorials/tour/unified-types.html))
- [x] **[Traits](https://github.com/scala/scala.github.com/issues/713)** ([old version](http://docs.scala-lang.org/tutorials/tour/traits.html))
- [x] **[Currying](https://github.com/scala/scala.github.com/issues/732)** ([old version](http://docs.scala-lang.org/tutorials/tour/currying.html))
- [x] **[Classes](https://github.com/scala/scala.github.com/issues/698)** ([old version](http://docs.scala-lang.org/tutorials/tour/classes.html))
- [x] **[Packages and imports](https://github.com/scala/scala.github.com/issues/714)**
- [x] **[Case classes](https://github.com/scala/scala.github.com/issues/715)** ([old version](http://docs.scala-lang.org/tutorials/tour/case-classes.html))
- [x] **[Pattern matching](https://github.com/scala/scala.github.com/issues/717)** ([old version](http://docs.scala-lang.org/tutorials/tour/pattern-matching.html))
- [x] **[Named arguments](https://github.com/scala/scala.github.com/issues/728)** ([old version](http://docs.scala-lang.org/tutorials/tour/named-parameters.html))
[...]

PR #735, opened 2017-03-16
Rewrote nested functions tour

ZeonTrevor:
an easier to understand example of using nested methods through the computation of factorial.

### DIFF
--- a/tutorials/tour/_posts/2017-02-13-nested-functions.md
+++ b/tutorials/tour/_posts/2017-02-13-nested-functions.md
@@ -1,6 +1,6 @@
 ---
 layout: tutorial
-title: Nested Functions
+title: Nested Methods
 
 disqus: true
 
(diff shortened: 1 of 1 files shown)


tutorials/tour/_posts/2017-02-13-nested-functions.md
```diff
@@ -11,25 +11,27 @@ next-page: currying
...
 
 ```tut
-object FilterTest extends App {
-  def filter(xs: List[Int], threshold: Int) = {
-    def process(ys: List[Int]): List[Int] =
-      if (ys.isEmpty) ys
-      else if (ys.head < threshold) ys.head :: process(ys.tail)
-      else process(ys.tail)
-    process(xs)
+object FactorialTest extends App {
```
MasseGuillaume: I would use `def main` here

travis032654:
Hey @ZeonTrevor, huge thanks for your work. It's going to help a lot of people who use this as their primary reference. Check out the conversation here: #733 

I'm also tracking the work in the Project Page. You'll find the link on the right.

tutorials/tour/_posts/2017-02-13-nested-functions.md
travis032654: I think it's clearer if all of the examples are in worksheet mode and not inside classes.

ZeonTrevor: @travis032654 I have rewritten the example in worksheet mode

SethTisue:
nice, thanks!

SethTisue:
after merging I noticed an error that I fixed in 4becbaf7f2ecbe98a1dbb2f013dac1f865c28142

PR #737, opened 2017-03-17
Rewrote local type inference tour

travissarles:


SethTisue:
I think we could drop the word "local" from the title. The number of people who will come in expecting ML/Haskell style global type inference is a small proportion of the overall audience.

SethTisue:
I would suggest adding material that explains that:

* method parameter types are never inferred
* but parameter types of anonymous functions can sometimes be inferred from context

(hmm, well, unless this material exists elsewhere in the tour? I haven't checked. if it does, not sure whether we should move it here, or link to it with a note, or what.)

travis032654:
OK I made the changes you suggested.

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+disqus: true
+
+tutorial: scala-tour
+categories: tour
+num: 29
+next-page: operators
+previous-page: polymorphic-methods
+prerequisite-knowledge: unified-types, generic-classes
+---
+The Scala compiler can often infer the type of a value so you don't have to declare it explicitly. You'll often see this with variables and return types.
```
SethTisue: instead of "type of a value", suggest "type of an expression"

SethTisue: and I'm not sure "You'll often see this with variables and return types" really adds any meaning?

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Omitting the type
+```
+val businessName = "Montreux Jazz Café"
+```
+The compiler can detect that `businessName` is a String. It works similarly with methods:
+```
+def squareOf(x: Int) = x * x
+```
+The compiler can infer that the return type is an int so no return type is required.
```
SethTisue: s/an int/an `Int`/

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+case class MyPair[A, B](x: A, y: B);
+val p = MyPair(1, "scala") // type: MyPair[Int, String]
+
+def id[T](x: T) = x
+val q = id(1)              // type: Int
+```
+The compiler uses the types of the arguments of `MyPair` to figure out what type `A` and `B` are. Likewise for the type of `x`.
+
+##  Parameters
+The compiler can never infer method parameter types. However, in certain cases, it can infer anonymous function parameter types when the function is passed as argument.
```
SethTisue: s/can never infer/never infers/ — slightly different shade of meaning

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+val doubled = Seq(1, 3, 4).map(x => x * 2)  // List(2, 6, 8)
+```
+The parameter for map is `f: A => B`. Because we put integers in the Seq, the compiler knows that `A` is `Int` (i.e. that `x` is an int). Therefore, the compiler can infer from `x * 2` that `B` is type `Int`.
```
SethTisue: `Seq` is missing backquotes around it

s/int/Int/ — unless you are really just using it as a noun, in which case "integer" not "int"

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+
+## When _not_ to rely on type inference
+
+It is generally considered more readable to declare the type of members exposed in a public API. Also, in some situations it can be quite dangerous to rely on Scala's type inference mechanism as the following program shows:
```
SethTisue: I don't think it's necessary to call something "quite dangerous" unless it compiles, yet fails at runtime.

I'd suggest: "In some situations, the compiler may infer a type that wasn't the one you intended."

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+
+```tut:fail
+object InferenceTest4 {
+  var obj = null
+  obj = new Object()
```
SethTisue: `new AnyRef` is more idiomatic Scala than `new Object()`

tutorials/tour/_posts/2017-02-13-type-inference.md
```diff
@@ -0,0 +1,63 @@
...
+val doubled = Seq(1, 3, 4).map(x => x * 2)  // List(2, 6, 8)
+```
+The parameter for map is `f: A => B`. Because we put integers in the Seq, the compiler knows that `A` is `Int` (i.e. that `x` is an int). Therefore, the compiler can infer from `x * 2` that `B` is type `Int`.
+
+## When _not_ to rely on type inference
+
+It is generally considered more readable to declare the type of members exposed in a public API. Also, in some situations it can be quite dangerous to rely on Scala's type inference mechanism as the following program shows:
+
+```tut:fail
+object InferenceTest4 {
```
SethTisue: is the `object` wrapper needed here?

SethTisue:
I can take care of getting this over the finish line.

tutorials/tour/_posts/2017-02-13-type-inference.md
jvican: I would add to your suggestion: "Therefore, we recommended that you make the type explicit for any APIs that will be exposed to users of your code" or something similar.

SethTisue:
replaced by #985

PR #738, opened 2017-03-17
Rewrote Mixin-class composition tour

travissarles:


jarrodwb:
I will work on the review tomorrow morning Hamburg time. :-)

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
 num: 6
 next-page: anonymous-function-syntax
 previous-page: traits
+prerequisite-knowledge: inheritance, traits, abstract-classes, unified-types
 ---
+Mixin class composition means that a class can reuse the members of one or more mixins (i.e. traits and types).
```
jarrodwb: I think it would be better to define what we mean by 'mixins' instead of 'mixin class composition'. Also, as far I undertand, you can only mixin 'Traits' so "(i.e. traits and types)" does not make sense to me.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
...
+
+```tut
+abstract class A {
+  def message: String
+}
+class B extends A {
+  def message = "I'm an instance of class B"
+}
+trait C extends A {
+  def makeMessageLoud = message.toUpperCase()
```
jarrodwb: I recommend renaming 'makeMessageLoud' to 'loudMessage'.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
...
+}
+class D extends B with C
+
+val d = new D
```
jarrodwb: I recommend replacing
```
class D extends B with C

val d = new D
```
with
```
val mixedClassInstance = new B with C
```

This simplifies the presentation and points out that 'B with C' is a class.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
 num: 6
 next-page: anonymous-function-syntax
 previous-page: traits
+prerequisite-knowledge: inheritance, traits, abstract-classes, unified-types
 ---
+Mixin class composition means that a class can reuse the members of one or more mixins (i.e. traits and types).
+
+```tut
```
jarrodwb: We may be able to find better names for classes A and B and trait C.

jarrodwb: Another idea is to change 'def' to 'val'. I don't have a strong preference here but it simplifies things a bit.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
...
-trait RichIterator extends AbsIterator {
-  def foreach(f: T => Unit) { while (hasNext) f(next) }
-}
-```
- 
-Here is a concrete iterator class, which returns successive characters of a given string:
- 
+This abstract iterator class could be used as a parameter type where an iterator is needed.
+
+Next, we'll implement a concrete class (all abstract members `type`, `hasNext`, and `next` have implementations):
```
jarrodwb: This should read:

... abstract members `T`, `hasNext`, and `next` have ...

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
...
- 
-```tut
-trait RichIterator extends AbsIterator {
-  def foreach(f: T => Unit) { while (hasNext) f(next) }
-}
-```
- 
-Here is a concrete iterator class, which returns successive characters of a given string:
- 
+This abstract iterator class could be used as a parameter type where an iterator is needed.
```
jarrodwb: What do you mean by "used as a parameter type"?

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -9,28 +9,41 @@ categories: tour
...
+
+Next, we'll implement a concrete class (all abstract members `type`, `hasNext`, and `next` have implementations):
+
 ```tut
```
jarrodwb: Here are a few changes I suggest to the StringIterator class
```
class StringIterator(s: String) extends AbsIterator {
  type T = Char
  private var i = 0

  def hasNext = i < s.length

  def next: Char = { 
    val ch = s charAt i
    i += 1
    ch
  }
}
```

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -39,17 +52,28 @@ class StringIterator(s: String) extends AbsIterator {
   def next = { val ch = s charAt i; i += 1; ch }
 }
 ```
- 
-We would like to combine the functionality of `StringIterator` and `RichIterator` into a single class. With single inheritance and interfaces alone this is impossible, as both classes contain member implementations with code. Scala comes to help with its _mixin-class composition_. It allows the programmers to reuse the delta of a class definition, i.e., all new definitions that are not inherited. This mechanism makes it possible to combine `StringIterator` with `RichIterator`, as is done in the following test program which prints a column of all the characters of a given string.
- 
+`StringIterator` takes a `String` and can be used to iterate over a String (e.g. to see if a String contains a certain character). `type T` must be `Char` so we know the return type of `next`.
```
jarrodwb: remove "`type T` must be `Char` so we know the return type of `next`."

jarrodwb: Also "over a String" to "over the String".

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -39,17 +52,28 @@ class StringIterator(s: String) extends AbsIterator {
...
+
+Now let's create a mixin class which also extends `AbsIterator`.
```
jarrodwb: Now let's create a trait which also extends ...

Traits are not classes even if I want to mix them in later. This could be debated. I don't have a strong opinion here but I think the words 'mixin class' are not common usage.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -39,17 +52,28 @@ class StringIterator(s: String) extends AbsIterator {
...
+
+```tut
+trait RichIterator extends AbsIterator {
+  def foreach(f: T => Unit) { while (hasNext) f(next) }
```
jarrodwb: This is deprecated style. We need the '='.

```
def foreach(f: T => Unit): Unit = while (hasNext) f(next)
```

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -39,17 +52,28 @@ class StringIterator(s: String) extends AbsIterator {
...
 object StringIteratorTest {
   def main(args: Array[String]) {
-    class Iter extends StringIterator(args(0)) with RichIterator
-    val iter = new Iter
+    class RichStringIter extends StringIterator(args(0)) with RichIterator
+    val iter = new RichStringIter
     iter foreach println
   }
 }
 ```
```
jarrodwb: Here is an idea to improve this. I think we should encourage people to explore these examples in the repl and not as stand alone programs. Just copy and paste into ammonite or scastie.
```
object StringIteratorTest {
    val testString = "I am a test string"
    val iter = new StringIterator(testString) with RichIterator
    def runTest(): Unit = iter foreach println
  }
}
```
Then run:
```
StringIteratorTest.runTest()
```

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
```diff
@@ -39,17 +52,28 @@ class StringIterator(s: String) extends AbsIterator {
...
- 
-The `Iter` class in function `main` is constructed from a mixin composition of the parents `StringIterator` and `RichIterator` with the keyword `with`. The first parent is called the _superclass_ of `Iter`, whereas the second (and every other, if present) parent is called a _mixin_.
+The new class `RichStringIter` has `StringIterator` as a superclass and `RichIterator` as a mixin.
+
+With single inheritance and interfaces alone this is impossible, as both classes contain member implementations with code. Because of Scala's support for _mixin-class composition_, classes can use members from mixins that aren't defined in the superclass or previously used mixins.
```
jarrodwb: There are many things wrong with this paragraph and it is unclear to me how to fix it.

I suggest scrapping it and rewriting it. Maybe something as simple as this.

" With single inheritance we would not be able to achieve this level of flexibility."

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
som-snytt: Or just paste into REPL.

I'd suggest losing the `main` method, which also uses procedure syntax here, and use `App` when called for. I'm a bigger fan of just pasting snippets, using the REPL as intended to show values and avoiding println everywhere. My least favorite style is a custom entry point like `runTest` as shown. I'm not above `MyApp main null` in REPL.

jarrodwb: Sounds good to me. Could you post a code snippet?

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
som-snytt: Something like this needs to be said. Example confusion on SO just now: http://stackoverflow.com/questions/43948337/variable-visibility-in-unrelated-scala-classes   Here, "with single inheritance..." is anticipating reader confusion instead of just saying what mixin composition does for you. (It's not as confusing as previous line 43.) It allows the concrete class to inherit members from traits that are otherwise unrelated. (Coming up with these phrasings is difficult, for sure.)

Anticipating the Java reader's confusion, one could say they're like interfaces with default method implementations.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
som-snytt: I wonder if `private[this] var` is too pedantically pragmatic for examples.

jarrodwb: I am not sure. I am not apposed to it. Maybe a link to another part of the tour that explains the semantics of  `private[this] var...` would make it acceptable.

tutorials/tour/_posts/2017-02-13-mixin-class-composition.md
travissarles: Once Guillaume finishes Scastie embedded, I'm going to move all of these examples into Scastie so people can play with it directly on the docs page. (Or be able to click on the code and have it bring you to a Scastie window to play with it).

som-snytt: That's really neat.

PR #741, opened 2017-03-21
Rewrote variances tour

mhzajac:
Split different types of variances into separate sections,
and replaced the complicated `Stack` example with simpler
ones.

### DIFF
--- a/tutorials/tour/_posts/2017-02-13-variances.md
+++ b/tutorials/tour/_posts/2017-02-13-variances.md
@@ -10,31 +10,142 @@ next-page: upper-type-bounds
 previous-page: generic-classes
 ---
 
-Scala supports variance annotations of type parameters of [generic classes](generic-classes.html). In contrast to Java 5 (aka. [JDK 1.5](http://java.sun.com/j2se/1.5/)), variance annotations may be added when a class abstraction is defined, whereas in Java 5, variance annotations are given by clients when a class abstraction is used.
-
-In the page about [generic classes](generic-classes.html) an example for a mutable stack was given. We explained that the type defined by the class `Stack[T]` is subject to invariant subtyping regarding the type parameter. This can restrict the reuse of the class abstraction. We now derive a functional (i.e. immutable) implementation for stacks which does not have this restriction. Please note that this is an advanced example which combines the use of [polymorphic methods](polymorphic-methods.html), [lower type bounds](lower-type-bounds.html), and covariant type parameter annotations in a non-trivial fashion. Furthermore we make use of [inner classes](inner-classes.html) to chain the stack elements without explicit links.
+Variance is the correlation of subtyping relationships of complex types and the subtyping relationships of their component types. Scala supports variance annotations of type parameters of [generic classes](generic-classes.html), to allow them to be covariant, contravariant, or invariant if no annotations are used. The use of variance in the type system allows us to make intuitive connections between complex types, whereas the lack of variance can restrict the reuse of a class abstraction.
 
 ```tut
-class Stack[+T] {
-  def push[S >: T](elem: S): Stack[S] = new Stack[S] {
-    override def top: S = elem
-    override def pop: Stack[S] = Stack.this
-    override def toString: String =
-      elem.toString + " " + Stack.this.toString
-  }
-  def top: T = sys.error("no element on stack")
-  def pop: Stack[T] = sys.error("no element on stack")
-  override def toString: String = ""
-}
+class Foo[+A] // A covariant class
+class Bar[-A] // A contravariant class
+class Baz[A]  // An invariant class
+```
 
-object VariancesTest extends App {
-  var s: Stack[Any] = new Stack().push("hello")
-  s = s.push(new Object())
-  s = s.push(7)
-  println(s)
+### Covariance
+
+A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `A` is a subtype of `B`, then `List[A]` is a subtype of `List[B]`. This allows us to make very useful and intuitive subtyping relationships using generics.
+
+Consider this simple class structure:
+
+```tut
+abstract class Animal {
+  def name: String
+}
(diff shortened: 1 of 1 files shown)


mhzajac:
#733

tutorials/tour/_posts/2017-02-13-variances.md
```diff
@@ -10,31 +10,114 @@ next-page: upper-type-bounds
...
+  val dogs: List[Dog] = List(Dog("Fido"), Dog("Rex"))
+  
+  printAnimalNames(cats)
+  printAnimalNames(dogs)
+}
+```
+
+### Contravariance
+
+A type parameter `T` of a generic class can be made contravariant by using the annotation `-T`. This creates a sub-typing relationship between the class and its type parameter that is similar, but opposite to what we get with covariance. That is, for some class `Writer[T]`, making `T` contravariant implies that for two types `A` and `B` where `A` is a sub-type of `B`, `Writer[B]` is a sub-type of `Writer[A]`.
```
travis032654: I think it would be clearer to have the full class signature `class Writer[-T]` to show what it looks like.

mhzajac:
Thanks for your feedback.

@julienrf I removed that part because it referenced the complicated `Stack` example, but perhaps that was too much. Looking at it again I think mentioning something like the original sentence "This can restrict the reuse of the class abstraction." is important. Is that the one you had in mind?

Also, naming this is hard. I originally called `Request[T]` `MyClass[T]`, instead, but was worried about using something that might seem cryptic to the first-time reader.

@travis032654 I stuck with `T` as it may come as more familiar to those coming from languages that seem favor it with generics. Happy to change it though. I'll make the updates later today.

tutorials/tour/_posts/2017-02-13-variances.md
```diff
@@ -10,31 +10,114 @@ next-page: upper-type-bounds
...
+  def printAnimalNames(animals: List[Animal]): Unit = {
+    animals.foreach { animal =>
+      println(animal.name)
+    }
+  }
+  
+  val cats: List[Cat] = List(Cat("Whiskers"), Cat("Tom"))
+  val dogs: List[Dog] = List(Dog("Fido"), Dog("Rex"))
+  
+  printAnimalNames(cats)
```
mhzajac: Since it prints out multiple lines, what might look acceptable?

Something like this?
```
// prints:
// Whiskers
// Tom
printAnimalNames(cats)
```

tutorials/tour/_posts/2017-02-13-variances.md
```diff
@@ -10,31 +10,114 @@ next-page: upper-type-bounds
 previous-page: generic-classes
 ---
 
-Scala supports variance annotations of type parameters of [generic classes](generic-classes.html). In contrast to Java 5 (aka. [JDK 1.5](http://java.sun.com/j2se/1.5/)), variance annotations may be added when a class abstraction is defined, whereas in Java 5, variance annotations are given by clients when a class abstraction is used.
+Scala supports variance annotations of type parameters of [generic classes](generic-classes.html) similar to C#. In contrast to Java 5 (aka. JDK 1.5), variance annotations may be added when a class abstraction is defined, whereas in Java 5, variance annotations are given by clients when a class abstraction is used.
```
SethTisue: I would suggest moving the comparison to Java and C# to a distinct section which comes at the end of the whole section, not at the beginning.

tutorials/tour/_posts/2017-02-13-variances.md
```diff
@@ -10,31 +10,114 @@ next-page: upper-type-bounds
...
 
-In the page about [generic classes](generic-classes.html) an example for a mutable stack was given. We explained that the type defined by the class `Stack[T]` is subject to invariant subtyping regarding the type parameter. This can restrict the reuse of the class abstraction. We now derive a functional (i.e. immutable) implementation for stacks which does not have this restriction. Please note that this is an advanced example which combines the use of [polymorphic methods](polymorphic-methods.html), [lower type bounds](lower-type-bounds.html), and covariant type parameter annotations in a non-trivial fashion. Furthermore we make use of [inner classes](inner-classes.html) to chain the stack elements without explicit links.
+### Covariance
+
+A type parameter `T` of a generic class can be made covariant by using the annotation `+T`. For some class `Request[T]`, making  `T` covariant implies that for two types `A` and `B` where `A` is a sub-type of `B`, then `Request[A]` is a sub-type of `Request[B]`. This allows us to make very useful and intuitive sub-typing relationships using generics.
```
SethTisue: Yes

tpolecat:
I have found that for a lot of people the key type to look at is `Function1` which is contravariant on the left and covariant on the right. If you get this one then the others are easy. So I think a section on this might be helpful. The discussion goes something like:

> If I ask you for a `Monkey => Fruit` and you give me an `Animal => Apple` it will work fine: it still takes a monkey (because it can take any animal) and it still returns a fruit (because an apple is a fruit). This makes intuitive sense, and is a formal relationship in Scala: `Animal => Apple` is a subtype of `Monkey => Fruit`.
>
> Java programmers may recognize this relationship because it is reflected in Java's rules for overridden methods, which can be more general in argument types and more specific in return types.

tpolecat:
I also agree that `[T]` should be `[A]` everywhere.

SethTisue:
> the key type to look at is Function1

+1, this is also exactly what I reach for every time I try to explain this

tpolecat:
Minor thing, but *subtype* and *subclass* should not be hyphenated.

tpolecat:
This is a big improvement, thanks for your work on this!

mhzajac:
Updated based on everyone's feedback. I still don't like the first section all that much.

tutorials/tour/_posts/2017-02-13-variances.md
travis032654: Maybe a multiline comment?

SethTisue:
I agree that the first section would benefit from simpler language, but that can always happen in some future revision. I'm going to go ahead and hit "merge" anyway since this a huge improvement overall. Thank you for this!

PR #742, opened 2017-03-21
Rewrote compound types tour

travissarles:


SethTisue:
Maybe this should just be folded into a section on traits?

Also, for whatever it's worth, I don't recall ever hearing people talk about "compound types". If a Scala person said that to me, I would have to think for a minute and wonder what they might be referring to.

travis032654:
I agree that this goes with the traits tour and the label "compound types" is a bit too academic (it's used in the [spec](https://www.scala-lang.org/files/archive/spec/2.11/03-types.html)).

vaibhavsingh1993:
Hi,
Is this issue still available? I can help if it is.
Thanks,
Vaibhav

jvican:
Yes @vaibhavsingh1993, please go ahead :smile:

PR #743, opened 2017-03-21
Delete xml tour

travissarles:


ashawley:
Delete the translations as well?

travis032654:
Yes, I'll delete those as well.

PR #747, opened 2017-03-23
Removed duplicated instead

krizalys:
The word `instead` is used twice in the same sentence on http://docs.scala-lang.org/tutorials/tour/singleton-objects.html, it sounds unnatural.


SethTisue:
thank you!

PR #751, opened 2017-03-29
small changes to basics

travissarles:


tutorials/tour/_posts/2017-02-13-basics.md
```diff
@@ -136,7 +141,7 @@ def add(x: Int, y: Int): Int = x + y
 println(add(1, 2)) // 3
 ```
 
-Methods cannot be named with the `val` or `var` keywords.
+Notice how the return type is declared _after_ the parameter list: `: Int`. Methods cannot be named with the `val` or `var` keywords.
```
SethTisue: > Methods cannot be named with the `val` or `var` keywords.

not sure about this addition. is this really needed here? if so, I think you need to expand on this and be clearer. (else, omit?) if you want to include it, I would suggest stating it positively rather than negatively: "A method is always defined using the `def` keyword."

tutorials/tour/_posts/2017-02-13-basics.md
```diff
@@ -160,6 +165,16 @@ println("Hello, " + name + "!")
 
 There are some other differences, but for now, you can think of them as something similar to functions.
 
+Method can have multi-line expressions as well.
```
SethTisue: "Methods" or "A method". Maybe "A method's body"?

tutorials/tour/_posts/2017-02-13-basics.md
```diff
@@ -160,6 +165,16 @@ println("Hello, " + name + "!")
...
+```tut
+def getSquareString(input: Double): String = {
+  val square = input * input
+  square.toString
+}
+```
+You can use the `return` but it is not necessary or idiomatic. Expressions
```
SethTisue: I would downplay and discourage `return` more strongly. Perhaps: "The last expression in the body is the method's return value. (Scala does have a `return` keyword, but it's rarely used.)"

tutorials/tour/_posts/2017-02-13-basics.md
```diff
@@ -170,6 +185,7 @@ class Greeter(prefix: String, suffix: String) {
     println(prefix + name + suffix)
 }
 ```
+The return type of the method `greet` is `Unit` which says there's nothing meaningful to return. It's used similarly to `void` in Java and C but the main difference is that there is actually a singleton value of type `Unit`, written `()`.
```
SethTisue: perhaps:

The return type of the method `greet` is `Unit`, which says there's nothing meaningful to return. It's used similarly to `void` in Java and C. (A difference is that because every Scala expression must have _some_ value, there is actually a singleton value of type `Unit`, written `()`. It carries no information.)

PR #754, opened 2017-03-30
Rewrote sequence comprehensions section of tour

travissarles:


tutorials/tour/_posts/2017-02-13-sequence-comprehensions.md
```diff
@@ -13,58 +13,39 @@ previous-page: extractor-objects
...
-```
- 
-The for-expression in function introduces a new variable `i` of type `Int` which is subsequently bound to all values of the list `List(from, from + 1, ..., to - 1)`. The guard `if i % 2 == 0` filters out all odd numbers so that the body (which only consists of the expression i) is only evaluated for even numbers. Consequently, the whole for-expression returns a list of even numbers.
+case class User(val name: String, val age: Int)
 
-The program yields the following output:
+val userBase = List(new User("Travis", 28),
+  new User("Kelly", 33),
+  new User("Jennifer", 44),
+  new User("Dennis", 23))
```
Ichoran: Although the style of indentation in parens varies widely, I'm not sure this one is a good one for a tutorial since we don't do trailing braces in Scala, generally, and trailing parens aren't that different from trailing braces.  I'd format it as
```scala
val userBase = List(
  new User("Travis", 28),
  ...
  new User("Dennis", 23)
)
```

tutorials/tour/_posts/2017-02-13-sequence-comprehensions.md
```diff
@@ -13,58 +13,39 @@ previous-page: extractor-objects
...
-         j <- i until n if i + j == v) yield
-      (i, j);
-  foo(20, 32) foreach {
-    case (i, j) =>
-      println(s"($i, $j)")
-  }
+def foo(n: Int, v: Int) =
+   for (i <- 0 until n;
+        j <- i until n if i + j == v)
+     yield (i, j)
```
Ichoran: Again, I'm not sure this is the best style to promote as it invites confusion about the comprehension part and the yield part.  You don't indent the `else` farther than the corresponding `if`, so I'd tend to

```scala
for (...)
yield (i, j)
```
or
for (
  ...
)
yield (i, j)
```

travissarles: I agree it looks nicer but all of the examples in Odersky's book use an indent as does IntelliJ.

PR #755, opened 2017-03-31
Rewrote inner classes tour section

travissarles:


jarrodwb:
Hi Travis. I will take a look later today. At the moment I am UTC+2 in Hamburg but I should have some time later this evening.

Thanks for the contribution and talk to you soon.

jarrodwb:
Sorry not today. 😴   Tomorrow for sure.

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -31,50 +31,32 @@ class Graph {
...
-  val n2 = g.newNode
-  val n3 = g.newNode
-  n1.connectTo(n2)
-  n3.connectTo(n1)
-}
-```
- 
-We now enrich the above example with types to state explicitly what the type of the various defined entities is:
- 
+This program represents a graph as a list of nodes (`var nodes`). Each node knows which other nodes it's connected to (`connectedNodes`). The `class Node` is a _path-dependent type_ because it is nested in the `class Graph`. Therefore, everything in the `connectedNodes` must be created using the class attached to Graph.
```
jarrodwb: The `var` in (`var nodes`) seems out of place.

jarrodwb: "Each node knows which..." seems odd. Nodes don't know anything since they don't have thoughts. This would be like saying a chair knows that it has four legs.

SethTisue: I find "using the class attached to Graph" unclear, in my opinion — I'd suggest trying to find a clearer wording here, even if it requires using more words.

SethTisue: agree on `var nodes`, I would expect to see e.g. `List[Node]` instead or whatever the actual type is

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -31,50 +31,32 @@ class Graph {
...
 ```tut
-object GraphTest extends App {
-  val g: Graph = new Graph
-  val n1: g.Node = g.newNode
-  val n2: g.Node = g.newNode
-  val n3: g.Node = g.newNode
-  n1.connectTo(n2)
-  n3.connectTo(n1)
-}
+val graph1: Graph = new Graph
```
SethTisue: in this and the immediately following line, why the type annotations?

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -31,50 +31,32 @@ class Graph {
...
+val node1: graph1.Node = graph1.newNode
+val node2: graph1.Node = graph1.newNode
+val node3: graph1.Node = graph1.newNode
+node1.connectTo(node2)
+node3.connectTo(node1)
 ```
- 
-This code clearly shows that a node type is prefixed with its outer instance (which is object `g` in our example). If we now have two graphs, the type system of Scala does not allow us to mix nodes defined within one graph with the nodes of another graph, since the nodes of the other graph have a different type.
+Notice how the type of all of the nodes is `graph1.Node`. This is because when we call `graph1.newNode` which calls `new Node`, the method is using the instance of `Node` specific to the instance `graph1`.
```
SethTisue: After the "Notice how" sentence, perhaps insert: "We have explicitly supplied that type, but the compiler would infer it if we had left it out."

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -31,50 +31,32 @@ class Graph {
...
+
+If we now have two graphs, the type system of Scala does not allow us to mix nodes defined within one graph with the nodes of another graph, since the nodes of the other graph have a different type.
```
SethTisue: This is the key sentence of the entire article. I would suggest stating this very close to the beginning, rather than withholding it until the reader has had to wade through a lot of details.

So at the start, say something like "Suppose we want the compiler to prevent us from mixing up which nodes belong to what graph. Path-dependent types gives us a way to do this."

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -11,7 +11,9 @@ next-page: abstract-types
 previous-page: lower-type-bounds
 ---
 
-In Scala it is possible to let classes have other classes as members. Opposed to Java-like languages where such inner classes are members of the enclosing class, in Scala such inner classes are bound to the outer object. To illustrate the difference, we quickly sketch the implementation of a graph datatype:
+In Scala it is possible to let classes have other classes as members. Opposed to Java-like languages where such inner classes are members of the enclosing class, in Scala such inner classes are bound to the outer object. Suppose we want the compiler to prevent us, at compile time, from mixing up which nodes belong to what graph. Path-dependent types provide a solution.
```
SethTisue: s/Opposed/As opposed/

tutorials/tour/_posts/2017-02-13-inner-classes.md
```diff
@@ -31,7 +33,7 @@ class Graph {
   }
 }
 ```
-This program represents a graph as a list of nodes (`var nodes`). Each node knows which other nodes it's connected to (`connectedNodes`). The `class Node` is a _path-dependent type_ because it is nested in the `class Graph`. Therefore, everything in the `connectedNodes` must be created using the class attached to Graph.
+This program represents a graph as a list of nodes (`List[Node]`). Each node has a list of other nodes it's connected to (`connectedNodes`). The `class Node` is a _path-dependent type_ because it is nested in the `class Graph`. Therefore, all noes in the `connectedNodes` must be created using the `newNode` from the same instance of `Graph`.
```
SethTisue: s/all noes/all nodes/

PR #756, opened 2017-04-02
Rewrote annotations section of tour

travissarles:


tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
-Annotations associate meta-information with definitions.
+Annotations associate meta-information with definitions. For example, the annotation `@deprecated` before a method causes the compiler to print a warning if the method is used.
+```tut
+@deprecated
+def hello = "hola"
 
-A simple annotation clause has the form `@C` or `@C(a1, .., an)`. Here, `C` is a constructor of a class `C`, which must conform to the class `scala.Annotation`. All given constructor arguments `a1, .., an` must be constant expressions (i.e., expressions on numeral literals, strings, class literals, Java enumerations and one-dimensional arrays of them).
+hello
+```
+This will compile but the compiler will print a warning: "there was one deprecation warning
```
SethTisue: I think "print a warning" or "print a deprecation warning" is sufficient here. keeping the focus on annotations

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
-> Since Scala has no checked exceptions, Scala methods _must_ be annotated with one or more `throws` annotations such that Java code can catch exceptions thrown by a Scala method.
-
-```
-package examples
-import java.io._
-class Reader(fname: String) {
-  private val in = new BufferedReader(new FileReader(fname))
-  @throws(classOf[IOException])
-  def read() = in.read()
+## Annotations that ensure correctness of encodings
```
SethTisue: maybe just "Annotations ensuring correctness" (maybe "enforcing" instead of "ensuring"?)

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
+Certain annotations will actually cause compilation to fail if a condition(s) is not met. For example, the annotation `@tailrec` ensures that a method is [tail-recursive](https://en.wikipedia.org/wiki/Tail_call). Tail-recursion can keep memory requirements constant. Here's how it's used in a method which calculates the factorial:
```
SethTisue: in place of "Tail-recursion can keep memory requirements constant", I would suggest: "Tail-cail optimization prevents tail-recursive code from consuming stack".

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
+```tut
+def factorial(x: Int): Int = {
+
```
SethTisue: omit blank line

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
+  @tailrec
+  def factorialHelper(x: Int, accumulator: Int): Int = {
+    if (x == 1) accumulator else factorialHelper(x - 1, accumulator * x)
```
SethTisue: `if/else` on a single line is hard to read IMO. also, the curly braces around the method body can be omitted.

SethTisue: I would suggest `x <= 1` instead of `x == 1`. 0! is 1 ([source](https://en.wikipedia.org/wiki/Factorial)).

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
-  private val in = new BufferedReader(new FileReader(fname))
-  @throws(classOf[IOException])
-  def read() = in.read()
+## Annotations that ensure correctness of encodings
+Certain annotations will actually cause compilation to fail if a condition(s) is not met. For example, the annotation `@tailrec` ensures that a method is [tail-recursive](https://en.wikipedia.org/wiki/Tail_call). Tail-recursion can keep memory requirements constant. Here's how it's used in a method which calculates the factorial:
+```tut
+def factorial(x: Int): Int = {
+
+  @tailrec
+  def factorialHelper(x: Int, accumulator: Int): Int = {
```
SethTisue: since the method is nested, I'd suggest simply `helper` over `factorialHelper`

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
+    if (x == 1) accumulator else factorialHelper(x - 1, accumulator * x)
+  }
+  factorialHelper(x, 1)
 }
 ```
-
-The following Java program prints out the contents of the file whose name is passed as the first argument to the `main` method.
-
+The `factorialHelper` method has the `@tailrec` which ensures the method is indeed tail-recursive. If we were to change the implementation of `factorialHelper` to the following, it would fail:
```
SethTisue: "has the `@tailrec`" should be either "has `@tailrec`" or "has the `@tailrec` annotation", according to taste

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
 
-```
-Main.java:11: exception java.io.IOException is never thrown in body of
-corresponding try statement
-        } catch (java.io.IOException e) {
-          ^
-1 error
-```
+## Annotations affecting code generation
+Some annotations like `@inline` affect the generated code. (e.g. your jar file might have different bytes than if you hadn't used the annotation). Inlining code means that the compiler finds the calls to a function definition and moves the body of the function to all the places where the method is called. This can increase performance but also increase the size of the generated code. Using the annotation `@inline` does not ensure that a method will be inlined, but it will cause the compiler to do it iff some heuristics about the size of the generated code are met.
```
SethTisue: you mean "i.e." not "e.g."

SethTisue: is "iff" a typo here, or do you mean "if and only if"?

SethTisue: for the middle two sentences, I suggest: "Inlining means inserting the code in a method's body at the call site. The resulting bytecode is longer, but hopefully runs faster."

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
 
 ### Java Annotations ###
+This section is for Java developers.
+**Note:** Make sure you use the `-target:jvm-1.8` option with Java annotations.
 
-**Note:** Make sure you use the `-target:jvm-1.5` option with Java annotations.
-
-Java 1.5 introduced user-defined metadata in the form of [annotations](http://java.sun.com/j2se/1.5.0/docs/guide/language/annotations.html). A key feature of annotations is that they rely on specifying name-value pairs to initialize their elements. For instance, if we need an annotation to track the source of some class we might define it as
+Java 1.5 introduced user-defined metadata in the form of [annotations](https://docs.oracle.com/javase/tutorial/java/annotations/). A key feature of annotations is that they rely on specifying name-value pairs to initialize their elements. For instance, if we need an annotation to track the source of some class we might define it as
```
SethTisue: I'd suggest replacing "Java 1.5 introduced" with "Java has" or "Java also has" — the introduction is now ancient history.

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,79 +11,51 @@ next-page: default-parameter-values
...
-corresponding try statement
-        } catch (java.io.IOException e) {
-          ^
-1 error
-```
+## Annotations affecting code generation
+Some annotations like `@inline` affect the generated code. (e.g. your jar file might have different bytes than if you hadn't used the annotation). Inlining code means that the compiler finds the calls to a function definition and moves the body of the function to all the places where the method is called. This can increase performance but also increase the size of the generated code. Using the annotation `@inline` does not ensure that a method will be inlined, but it will cause the compiler to do it iff some heuristics about the size of the generated code are met.
 
 ### Java Annotations ###
+This section is for Java developers.
```
SethTisue: This needs expansion and explanation, in my opinion. If it's for Java developers, what's it doing here? In what situation would a Java developer care about this? Does it have to do a particular kind of Java/Scala interop, or what? I'd suggest stating this up front.

tutorials/tour/_posts/2017-02-13-annotations.md
```diff
@@ -11,77 +11,57 @@ next-page: default-parameter-values
 previous-page: automatic-closures
 ---
 
-Annotations associate meta-information with definitions.
+Annotations associate meta-information with definitions. For example, the annotation `@deprecated` before a method causes the compiler to print a warning if the method is used.
+```tut:fail
+object DepricationDemo extends App {
```
SethTisue: Deprication -> Deprecation

PR #757, opened 2017-04-05
Rewrote by-name parameters section of tour

travissarles:


tutorials/tour/_posts/2017-02-13-by-name-parameters.md
```diff
@@ -0,0 +1,46 @@
...
+disqus: true
+
+tutorial: scala-tour
+categories: tour
+num: 31
+next-page: annotations
+previous-page: operators
+---
+
+_By-name parameters_ are only evaluated when used. They are in contrast to _by-value parameters_. To make a parameter called by value, simply prepend `=>` to its type.
```
SethTisue: "To make a parameter called by name", you mean

tutorials/tour/_posts/2017-02-13-by-name-parameters.md
```diff
@@ -0,0 +1,46 @@
...
+
+```tut
+def goToSleep(tired: Boolean, seconds: => Int) = if (tired) Thread.sleep(seconds)
```
SethTisue: too long for a one-liner, especially given the site layout

tutorials/tour/_posts/2017-02-13-by-name-parameters.md
```diff
@@ -0,0 +1,46 @@
...
+
+def lengthOfNap = 3600 * 8
+goToSleep(tired = false, seconds = lengthOfNap)
+```
+The type of `seconds` is `=> Int` which means `seconds` is a by-name parameter. Therefore, when we call `goToSleep` it with `tired = false`, `lengthOfNap` will not be evaluated.
```
SethTisue: This is not that compelling of an example. I would suggest leading with an example where avoiding unwanted evaluation is actually crucial. Computing a number of seconds is highly unlikely to be an expensive computation.

SethTisue: What did you see as the downsides with the examples in the original version...?

travissarles: Looking back at the original example, it _does_ look better. I'll put it back.

tutorials/tour/_posts/2017-02-13-by-name-parameters.md
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+var i = 2
+
+whileLoop (i > 0) {
+  println(i)
+  i -= 1
+}  // prints 2 1
+```
+The method `whileLoop` uses multiple parameter lists to take a condition and a body of the loop. If the `condition` is true, the `body` is executed and then a recursive call to whileLoop is made. If the `condition` is false, the body is never evaluated because we prepended `=>` to the return type.
```
S11001001: `=>` wasn't appended to the return type; it was appended to `body`'s type.

tutorials/tour/_posts/2017-02-13-by-name-parameters.md
```diff
@@ -0,0 +1,40 @@
...
+
+Now when we pass `i > 0` as our `condition` and `println(i); i-= 1` as the `body`, it behaves like the standard while loop in many languages. Here is the output:
```
S11001001: Where is it?

travissarles: I'll remove that sentence since I have the output in a comment.

PR #759, opened 2017-04-06
Rewrote abstract types section of tour

travissarles:


tutorials/tour/_posts/2017-02-13-abstract-types.md
```diff
@@ -11,51 +11,46 @@ next-page: compound-types
...
+println("content = " + buf.element)
 ```
- 
-The return type of method `newIntSeqBuf` refers to a specialization of trait `Buffer` in which type `U` is now equivalent to `Int`. We have a similar type alias in the anonymous class instantiation within the body of method `newIntSeqBuf`. Here we create a new instance of `IntSeqBuffer` in which type `T` refers to `List[Int]`.
+Here the factory `newIntSeqBuf` uses an anonymous class implementation of `IntSeqBuf` (i.e. `new IntSeqBuffer`), setting `type T` to a `List[Int]`.
+
+Note that it is often possible to turn abstract type members into type parameters of classes and vice versa. Here is a version of the code above which only uses type parameters:
 
-Please note that it is often possible to turn abstract type members into type parameters of classes and vice versa. Here is a version of the code above which only uses type parameters:
-
```
som-snytt: It's nice to say please. But maybe it's not even a note. Type members and type params is a fundamental concept.

PR #761, opened 2017-04-07
Rewrote extractors section of tour

travissarles:


travissarles:
@julienrf @sethtissue any comments?

travissarles:
@julienrf @sethtisue any comments?

tutorials/tour/_posts/2017-02-13-extractor-objects.md
```diff
@@ -11,25 +11,25 @@ next-page: sequence-comprehensions
...
-There are two syntactic conventions at work here:
+object CustomerID{
 
-The pattern `case Twice(n)` will cause an invocation of `Twice.unapply`, which is used to match any even number; the return value of the `unapply` signals whether the argument has matched or not, and any sub-values that can be used for further matching. Here, the sub-value is `z/2`
+  def apply(name: String) = s"$name--${Random.nextLong}"
 
-The `apply` method is not necessary for pattern matching.  It is only used to mimick a constructor. `val x = Twice(21)` expands to `val x = Twice.apply(21)`.
+  def unapply(customerID: String): Option[String] = {
+    val name = customerID.split("--")(0)
+    if (!name.isEmpty) Some(name) else None
```
julienrf: You could use `if (name.nonEmpty) …`.

On the line above, you could also use `customerID.split("--").head`, or even `.headOption`, but `split` always returns at least one element, so I think that’s fine…

tutorials/tour/_posts/2017-02-13-extractor-objects.md
```diff
@@ -11,25 +11,25 @@ next-page: sequence-comprehensions
...
+  }
+}
+
+val customer1ID = CustomerID("Sukyoung")  // Sukyoung-23098234908
+val CustomerID(name) = customer1ID
+println(name)  //  Sukyoung
+```
+The `apply` method creates a `CustomerID` string from a `name`. The `unapply` does the inverse to get the `name` back. Although `CustomerID` doesn't take any parameters, when we call `CustomerID("Sukyoung")`, this is shorthand syntax for calling `CustomerID.apply("Sukyoung")`. When we call `val CustomerID(name) = customer1ID`, this is shorthand for `val name = CustomerID.unapply(name)`.
```
julienrf: This is not exact: `val CustomerID(name) = customer1ID` is a shorthand for `val name = CustomerID.unapply(name).get` (note the trailing `.get` call that is missing in your description).

Also, I think it is worth reminding that patterns can also be used in pattern matching (not just in val definitions).

Finally, the `unapply` method don’t have to return an `Option` (but that information is relevant for advanced users only, I think), and you can also define an `unapplySeq` method that extracts a variable number of information.

SethTisue:
> @sethtisue any comments?

(it's unlikely I'll have any time to review any of the tour PRs until May, I'm afraid)

tutorials/tour/_posts/2017-02-13-extractor-objects.md
```diff
@@ -11,32 +11,41 @@ next-page: sequence-comprehensions
...
+```tut
+val customer2ID = CustomerID("Nico")
+val customer2Name = customer2ID match {
+  case CustomerID(name) => name
+}
+```
+In fact, if we tried to unapply an invalid `CustomerID`, we would get a `scala.MatchError`:
+```tut:fail
+val CustomerID(name2) = "--asdfasdfasdf"
+name2.isEmpty
```
S11001001: This line will fail in `tut` because `name2` is unbound, which probably obscures the point you're making (where only the first line's error is relevant).

jvican:
> This rather positive effect implies that it may be worth recommending that irrefutable extractors, those explicitly designed for extraction with val instead of might-fail patterns, ought to be ascribed in this way.
  
Interesting, didn't know this trick. Thanks for sharing.

tutorials/tour/_posts/2017-02-13-extractor-objects.md
```diff
@@ -11,32 +11,41 @@ next-page: sequence-comprehensions
...
-}
-```
+object CustomerID {
 
-There are two syntactic conventions at work here:
+  def apply(name: String) = s"$name--${Random.nextLong}"
 
-The pattern `case Twice(n)` will cause an invocation of `Twice.unapply`, which is used to match any even number; the return value of the `unapply` signals whether the argument has matched or not, and any sub-values that can be used for further matching. Here, the sub-value is `z/2`
+  def unapply(customerID: String): Option[String] = {
+    val name = customerID.split("--").head
```
travissarles: @S11001001 I'm not sure what you mean, can you elaborate?

tutorials/tour/_posts/2017-02-13-extractor-objects.md
travissarles: The `val` keyword doesn't bind it? I'm seeing a MatchError when I run it in a worksheet.

SethTisue: @travissarles yeah I think you're right here

S11001001: Here is what tut emits.  Note the second error.

----

This is equivalent to `val name = CustomerID.unapply(customer2ID).get`. If there is no match, a `scala.MatchError` is thrown:

```scala
scala> val CustomerID(name2) = "--asdfasdfasdf"
scala.MatchError: --asdfasdfasdf (of class java.lang.String)
  ... 198 elided

scala> name2.isEmpty
<console>:14: error: not found: value name2
       name2.isEmpty
       ^
```

SethTisue: ah, I see. because tut is REPL based.

nbd either way I think, but I'd be fine with omitting the `name2.isEmpty` line

tutorials/tour/_posts/2017-02-13-extractor-objects.md
S11001001: Currently:

```scala
scala> val CustomerID(x) = CustomerID("a--b")
x: String = a
```

tutorials/tour/_posts/2017-02-13-extractor-objects.md
```diff
@@ -11,32 +11,49 @@ next-page: sequence-comprehensions
...
+The `apply` method creates a `CustomerID` string from a `name`. The `unapply` does the inverse to get the `name` back. When we call `CustomerID("Sukyoung")`, this is shorthand syntax for calling `CustomerID.apply("Sukyoung")`. When we call `case CustomerID(name) => customer1ID`, we're calling the unapply method.
 
-The pattern `case Twice(n)` will cause an invocation of `Twice.unapply`, which is used to match any even number; the return value of the `unapply` signals whether the argument has matched or not, and any sub-values that can be used for further matching. Here, the sub-value is `z/2`
+The unapply method can also be used to assign a value.
 
-The `apply` method is not necessary for pattern matching.  It is only used to mimick a constructor. `val x = Twice(21)` expands to `val x = Twice.apply(21)`.
+```tut
+val customer2ID = CustomerID("Nico")
+val CustomerID(name) = customer2ID
+println(name)  // prints Nico
```
S11001001: Can probably do without this `println`, as tut writes

```scala
scala> val customer2ID = CustomerID("Nico")
customer2ID: String = Nico--8217157908548994001

scala> val CustomerID(name) = customer2ID
name: String = Nico

scala> println(name)  // prints Nico
Nico
```

SethTisue: We aren't actually using the output of tut. We're just using it to verify that code snippets compile (or don't). Actually using the output is a possible future improvement.

S11001001: @SethTisue Ah, I see.

tutorials/tour/_posts/2017-02-13-extractor-objects.md
travissarles: @S11001001 I took out the name2.isEmpty as it's not helpful

tutorials/tour/_posts/2017-02-13-extractor-objects.md
travissarles: I'd like to use Scastie in the future so people can see the output and play around with the examples.

travissarles:
@julienrf any other comments?

PR #775, opened 2017-05-18
run jekyll with nix

MasseGuillaume:


heathermiller:
Could you explain why this is necessary?

MasseGuillaume:
I use a package manager called nix and It's really strict to manage dependencies. Without this file, it's not possible to contribute to the documentation.

PR #776, opened 2017-05-18
run jekyll with nix

MasseGuillaume:


SethTisue:
Could this include some instructions?

heathermiller:
Why is this necessary?

MasseGuillaume:
> Why is this necessary?
I use a package manager called nix and It's really strict to manage dependencies. Without this file, it's not possible to contribute to the documentation.

> Could this include some instructions?
The instruction is at the top of the file. I can create a follow-up PR and add a CONTRIBUTING.md to include instruction for other platforms.

SethTisue:
> The instruction is at the top of the file

But it doesn't answer the question, “Why?"

I think it's fine to have optional stuff like this lying around, it should just say what it's for and include some indication that it's optional.

PR #777, opened 2017-05-18
Fixed small template error

DiegoReiriz:
There's an error on that post, that makes it unnable to be rendered in the web browser


heathermiller:
Good catch @DiegoReiriz! Thank you :)

PR #787, opened 2017-06-01
Fix formatting of code snippets in Future docs

i:
Replaced some rogue tabs with appropriate spaces to fix the formatting of some snippets on http://docs.scala-lang.org/overviews/core/futures.html


heathermiller:
Always useful to clean this stuff up! Thank you for catching and fixing 😃

PR #791, opened 2017-06-06
Refer to 'by-name-parameters' page instead of non-existing page

Philippus:
Refer to 'by-name-parameters' page instead of non-existing 'automatic-closures' page


SethTisue:
thanks 👍

PR #794, opened 2017-06-10
Fixed next/previous links in tutorials/tour docs

devsdmf:
After the removal of the xml-processing tutorial, I realized that the next/previous links in the english tutorial tour was broken by an inexistent page, and as I fixed that, I found that the translations does not has the next/previous links too, so I decided to fix it too.


SethTisue:
thank you!

PR #797, opened 2017-06-12
Create 2017-02-13-mixin-class-composition.md

borromeotlhs:
I am suggesting this change as this seems to be what you all mean, though I could be wrong, as I'm trying to follow the written logic which seems circular. . .


SethTisue:
sorry about this PR falling through the cracks. it got fixed anyway in #899

PR #798, opened 2017-06-13
FOR THE LOVE OF GOD!! Remove Discourse from translations

heathermiller:
😱 😱 😱 😱


heathermiller:
Merged immediately to make the flood stop. I'll go and delete new threads in other languages that were created now.

cc/ @SethTisue @travissarles

PR #801, opened 2017-06-17
Show all Scaladoc options

som-snytt:
Demonstrate the styles of indentation without prejudice.

### DIFF
--- a/_style/scaladoc.md
+++ b/_style/scaladoc.md
@@ -12,43 +12,65 @@ previous-page: declarations
 
 It is important to provide documentation for all packages, classes,
 traits, methods, and other members. Scaladoc generally follows the
-conventions of Javadoc, however there are many additional features to
-make writing scaladoc simpler.
+conventions of Javadoc, but provides many additional features that
+simplify writing documentation for Scala code.
 
 In general, you want to worry more about substance and writing style
-than in formatting. Scaladocs need to be useful to new users of the code
+than about formatting. Scaladocs need to be useful to new users of the code
 as well as experienced users. Achieving this is very simple: increase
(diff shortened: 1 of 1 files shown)


style/scaladoc.md
```diff
@@ -10,43 +10,65 @@ previous-page: declarations
...
      *
-     *  This is further documentation of what we're documenting.  It should
-     *  provide more details as to how this works and what it does.
+     * This is further documentation of what we're documenting.
+     * Here are more details about how it works and what it does.
      */
-    def myMethod = {}
+    def member: Unit = ()
+
+Scaladoc style, with gutter asterisks aligned in column two:
```
olafurpg: s/ScaLadoc/Scaladoc/ ?

eddsteel:
To fix the issue where tools/editors refer to "Scaladoc style", could these have names? I've seen flat and ragged used elsewhere. Thus your options could be "Javadoc", "Flat Scaladoc", "Ragged Scaladoc".

olafurpg:
I totally agree @eddsteel, if the official style will advertise three different styles then it should give each style a name. However, I think we should first try to see if we can agree on a single style first and avoid the need for coming up with three names. Let's continue this discussion in https://contributors.scala-lang.org/t/quiet-changes-to-the-style-guide/923/6

SethTisue:
hey @som-snytt, rebase? and then let's get this merged. in the absence of any broader changes to how the style guide is managed, in the meantime we might as well merge this, it reflects the current status quo

SethTisue:
thanks, A.P.!

som-snytt:
@SethTisue N.P.

PR #803, opened 2017-06-19
upgrade Jekyll from 3.3.0 to 3.4.3

SethTisue:
3.4.3 is what GitHub Pages currently uses, as per
https://pages.github.com/versions/ , so we should use it when
building locally, too

this also fixes a problem I had on OS X where json 1.8.3 failed to
install (see e.g. https://github.com/fastlane/fastlane/issues/7673)


SethTisue:
@jvican @travissarles this is failing on Drone because the default Ruby version on the CI server is too old: `Gem::InstallError: nokogiri requires Ruby version >= 2.1.0`

I tried adding `rbenv local 2.4.1` to `.drone.yml`, since that's what I did to get this running locally on OS X, but I just get `/bin/sh: rbenv: not found`

I don't know what to do next as my familiarity with Drone, Ruby, and how the Scala Center's Drone server is configured is limited. if this were Travis-CI, I'd do `apt-get install rbenv`, I guess. is that what I do here? not sure what the next step is.

heathermiller:
Looks like this was taken care of in #811 by @cooperc0 (thanks again!)

Sorry for missing this one, @SethTisue. 

Heads up though; @fsalvi just moved this build to EPFL machines, which means we can now add plugins and use https (plugins are required for #806)

The build is here: https://scala-webapps.epfl.ch/jenkins/job/build_docs_scala_website/
(Adding a ticket to remind us to add badges etc to the README)

PR #805, opened 2017-07-12
SIP-NN: Right-Associative By-Name Operators

szeiger:
See https://github.com/scala/scala/pull/5969 for previous discussion.

### DIFF
--- a/sips/pending/_posts/2017-07-12-right-associative-by-name-operators.md
+++ b/sips/pending/_posts/2017-07-12-right-associative-by-name-operators.md
@@ -1,0 +1,90 @@
+---
+layout: sip
+discourse: true
+title: SIP-NN - Right-Associative By-Name Operators
+---
+
+**By: Stefan Zeiger**
+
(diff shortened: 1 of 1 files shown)


szeiger:
BTW, regarding the SIP creating process:

- On http://docs.scala-lang.org/sips/sip-submission.html, the links to "S(L)IP template" and "README" are broken
- It tells you to "Create a link to your SIP in the “pending sips” section of index.md". There are two index.md files, neither of which has a "pending sips" sections. Judging by previously merged SIP submissions there is no place to add the link.
- The attempt to update the Jekyll dependency mess with bundler aborted with some C compilation errors, as expected from previous attempts.

jvican:
Thanks for the SIP and the observations to the current instructions, I'll fix them when I get some time! I don't understand why I didn't get a notification for this.
  
Merging :smile:. This will be discussed in our next SIP meeting. I'll send an email on Monday to organise it.

PR #807, opened 2017-07-15
Fix parameter name in tutorial

Abrasha:


Abrasha:
@heathermiller Could you also review this one? Thanks.

heathermiller:
Good catch! Thanks a lot for the fix ❤️

PR #812, opened 2017-07-18
Fix class name example

jotafeldmann:
The class name from the text doesn't match the class name in the last example.


heathermiller:
Woohoo! Thank you very much @jotafeldmann!! The fix is much appreciated ❤️

PR #822, opened 2017-07-21
Tweak the `@stableABI` title to match the others

dwijnand:
Just adjusting an outlier:

image

### DIFF
--- a/sips/pending/_posts/2017-01-13-binary-compatibility.md
+++ b/sips/pending/_posts/2017-01-13-binary-compatibility.md
@@ -1,6 +1,6 @@
 ---
 layout: sip
-title: SIP XX - Improving binary compatibility with @stableABI
+title: SIP-NN - Improving binary compatibility with @stableABI
 discourse: true
 ---
 


heathermiller:
Thanks @dwijnand. Fixes for these little things are really helpful; I really appreciate it.

As an aside, I'm currently working on overhauling the doc site. Since you are one of the people using the SIP part of the site more than the average, is there anything you think should be improved or added to the SIP pages? I've already got a wishlist from Jorge.

dwijnand:
I'll try and have a little look, see if I can think of anything. Where would you like to receive such feedback?

heathermiller:
I'm doing a total overhaul of it. So tell me whatever you think should be different. There are no bounds on changing the SIP site. If you think I should put all information on one page, that's a valid suggestion. If you think more information is needed somewhere, also valid suggestion. Whatever you think should be done differently in terms of communicating information about the SIP process and in-process SIPs, let me know. I'm open to anything; I'm already overhauling it, so time is allocated to do whatever is needed to it.

The only requirement is being able to redirect old links to something that makes sense.

dwijnand:
@heathermiller I don't have an strong suggestions. I'm sure you'll find things like the wrong and broken link in http://docs.scala-lang.org/sips/sip-submission.html labelled "process template" that points to https://github.com/scala/slip/blob/master/slip-template.md . As well as the whole obsolete SLIP section.

If it's useful to you I'd be happy to provide feedback on the overhaul, time permitting.

dwijnand:
OCD.. https://github.com/scala/scala.github.com/pull/823

heathermiller:
Yep, those things definitely are slated to be resolved. Ok, if you think of anything that is missing or should be done differently at any point, let me know.

heathermiller:
Thanks!

SethTisue:
@heathermiller here's a SIP page usability thing: it happened just the other day on scala/scala Gitter that Haoyi couldn't find a SIP, because by default only the "pending" ones show. & I've repeatedly been confused by the same thing myself. I'd suggest showing all SIPs, of any status, right on the front page.

heathermiller:
Gah, yes, you're right about that. I've been bitten by this as well but had forgotten about it. Thanks for remembering, I'll make sure that all are listed.

PR #824, opened 2017-07-21
Fix TOC links in documentation pages

### LINKED ISSUE #819: Broken links of sidebar in Collections
I found out the sidebar of [collections](http://docs.scala-lang.org/overviews/collections/introduction.html) is broken. All of links in sidebar are missed `.html`. But It's fine when you built in local without `.html`. Sorry I am not the expert of frontend, I couldn't figure out the reason.

jrudolph:
According to https://github.com/jekyll/jekyll/issues/4145#issuecomment-158109424 `:output_ext` needs to be used to make all links generated with `page.url` contain the extension.

Fixes #819 

I haven't checked if it really works or if it breaks other links. I haven't set up jekyll right now to try out if something breaks, so maybe merge only after testing.

An alternative solution would be to configure the EPFL server to do the necessary redirecting to find documents for URLs missing the extension as explained in [the Jekyll documentation](http://jekyllrb.com/docs/permalinks/#extensionless-permalinks).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,7 +18,7 @@ keywords:
 scala-version: 2.12.0
 
 highlighter: rouge
-permalink: /:categories/:title.html
+permalink: /:categories/:title.html:output_ext
 baseurl:
 exclude: ["vendor"]
 gems:


jiminhsieh:
I built with this patch in local. It did generate link with file extension.

heathermiller:
Thanks @jrudolph and @jiminhsieh!

PR #825, opened 2017-07-24
Fix a zh-cn translation mistake about implicit conversion priority

airt:
Fix a zh-cn translation mistake about implicit conversion priority
in `zh-cn/overviews/collections/arrays`.


heathermiller:
@liufengyun, by any chance could you double-check this? 

(My Chinese skills are a bit rusty 😛 )

liufengyun:
@heathermiller The translation looks good to me.

heathermiller:
Great, thanks a lot @airt! And thanks for looking over it @liufengyun!

liufengyun:
The original translation gets confused with `inherited from` vs. `inherited by`, I also find the English text a little confusing:

> The first is defined in the `Predef` object whereas the second is defined in a class `scala.LowPriorityImplicits`, which is inherited from `Predef`.

PR #806, opened 2017-07-14
WIP: merging look and feel of scala-lang with doc site

heathermiller:
Todos before this can be merged
- [x] Restore lost SIPs/SLIP pages
- [x] Fix broken translation links on landing page
- [x] Restore footer
- [x] Restore Java to Scala guide
- [x] Restore style guide
- [x] Language dropdown box instead of unordered list
- [x] Fix syntax highlighting
- [x] Remove "what is Scala" on landing page 
- [x] Fix glossary styling
- [x] Rearrange landing page masthead (some important things are swallowed, like list of previous API docs)
- [x] Need a nav bar for docs
- [x] Overview pages need subtitles
- [x] Rework overviews landing page
- [x] Doc template
- [x] Rework TOCs
  - [x] Based on collections instead
  - [x] Expand TOC elements for current page
  - [x] Language switcher
- [x] Multipage overview templates w/ language switch
- [x] Singlepage overview templates w/ language switch
- [x] Tour w/ language switch
- [x] Tutorials w/ language switch
- [x] Fix cheatsheet styling
- [x] Tutorials landing page
- [x] SIP layouts
- [x] SIP index page
- [x] Make it possible to use tut AND rouge syntax highlighting
- [x] Sort out getting started guides
- [x] All versions of API docs page
- [x] Improve formatting of index page of style guide

Nice to haves, but not required for merge:
- Make cheatsheets table more responsive (breaks on very small screens)
- New algolia search
- Fit search bar into doc navigation bar

...more forthcoming

PR #837, opened 2017-08-01
next on iterators is side-effecting, so use parens

SethTisue:
as per discussion at
https://users.scala-lang.org/t/help-understanding-a-function-parameter-and-f-next/1480


heathermiller:
This will conflict with #806.

Have you managed to take a look at it yet, @SethTisue?

ilango123:
Thanks for this

SethTisue:
I can wait and rebase-and-merge this after #806 lands.

re: #806 generally, the landing page looks great. I've looked superficially at it, but in general, I'm fine with seeing it merged sooner than later & we can gradually sort out any regressions later, just as we've been doing over in the scala/scala-lang repo.

PR #840, opened 2017-08-03
Removed unexpected bracket

amphineko:


heathermiller:
Good catch! Thank you for the fix!!

PR #843, opened 2017-08-04
extractor-objects.md: Option[Seq[T]] is useful when the number of values isn't fixed

jbalintbiro:


SethTisue:
thanks, nice catch.

PR #847, opened 2017-08-04
Fix rebuild example to be compliant with previous section

valentinarho:
Change the rebuild functionality example to be compliant to the code created before in the tutorial.


SethTisue:
sorry — can you rebase against current master?

valentinarho:
The problem is currently already solved in the master branch, so this pull request is no more needed.

PR #848, opened 2017-08-07
Remove old wiki links

ashawley:


heathermiller:
Thank you @ashawley!! ❤️

PR #849, opened 2017-08-07
Markdown toc for style guide

ashawley:
Doesn't look like there was a reason for the table of contents to be in  HTML, so simplify?  It renders the same.


heathermiller:
Can we please keep the `{{ site.baseurl }}`? Since this is not on GitHub pages, relative linking is sometimes broken. 

Plus, if we ever want to post a version of the site for people to comment on ever again (like I did here: http://lampwww.epfl.ch/~hmiller/new-docsite) we need the absolute paths so the site can be placed in a subdirectory and for links to still work. 

(Not having `{{ site.baseurl }}` all over the place was the reason so many links were broken in the above temp site)

ashawley:
Ah, ok, that's good to know. I noticed a lot of the site was using root-relative linking, so I assumed it was ok.  I'm surprised Jekyll doesn't DTRT with links in markdown here by inserting `site.baseurl` in root-relative links... :-P

heathermiller:
Yeah, it's a super confusing situation. 

In the _data files (.yml), we can't use site variables, so we've got to use relative linking, which usually works (but not always). So that's where you might be seeing a bunch of relative links when you look around.

Anywhere else though, such as in markdown files or CSS/JS (with empty `---` `---` as their frontmatter), we can and should do the `{{ site.baseurl }}` thing. ...Hopefully Jekyll will one day get a bit smarter about linking.

I actually looked around for a link analyzing plugin for Jekyll to insert the stupid `{{ site.baseurl }}` thing but couldn't find anything that worked. (I almost wrote a plugin myself, but stopped myself...)

ashawley:
I'm so confused.  I'm looking at the links on this page http://lampwww.epfl.ch/~hmiller/new-docsite/tour/tour-of-scala.html

When I click on the first link, I got to http://lampwww.epfl.ch/~hmiller/new-docsite/unified-types.html

That's totally broken. I can confirm that.  The `/tour` part of the URL was dropped.  However, it should have worked.  The markup was:

```html
<a href="unified-types.html">every value is an object</a>
```

Is there some Javascript on the site interacting with links?  When I disable Javascript, the link works.  What???

heathermiller:
Yeah, on that site there's JavaScript fixing relative linking problems. ...In that case, the JavaScript broke things more than it fixed them.

Though most other links wouldn't work w/out the JS futzing.

ashawley:
Ah, ok.  That makes me feel a little more sane, then.

So, without the JS futzing, won't the links on this page work fine, then?

heathermiller:
Yep, they should. (This is one case where relative linking actually works 🙄 )

heathermiller:
Merci! ❤️

PR #850, opened 2017-08-07
Remove unnecessary refs to docs.scala-lang.org

ashawley:
- Abbreviate/shorten the URLs in Markdown when possible
- Preserve non-relative links per #849 discussion
- Fix page anchors that were incorrectly using hyphen instead of a dash to separate words.


heathermiller:
Merci 🙂

PR #851, opened 2017-08-08
Fix broken link of "Problem with this page".

jiminhsieh:
I found there are some broken links in "Problem with this page?" of right sidebar.

For example, http://docs.scala-lang.org/glossary/index.html and http://docs.scala-lang.org/api/all.html.

Those pages would get empty from `page.relative_path`, due to not all pages or markdowns in collection.

After I did some research, I found out this [comment](https://github.com/jekyll/jekyll/issues/2897#issuecomment-55149803), the author proposed this [walkaround solution](https://github.com/jekyll/jekyll/issues/2897#issuecomment-55157824).

So I replaced `page.relative_path` to this solution. I am not sure there is better solution or not. If there is a better solution, please just close it. : )

### DIFF
--- a/_includes/sidebar-toc-glossary.html
+++ b/_includes/sidebar-toc-glossary.html
@@ -5,6 +5,6 @@
       <div id="toc"></div>
 		</div>
 		<hr>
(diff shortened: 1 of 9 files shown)


heathermiller:
😍 THANK YOU! I recently noticed this! 
(And was super annoyed by it, Jekyll should be smart enough to do this automatically, right!?)

SethTisue:
nice work, thank you

PR #853, opened 2017-08-08
Fix broken links to "Anonymous function syntax"

ashawley:
Anonymous function syntax was dropped from the tour in #734, and one link was fixed in #788 but this is the rest.


heathermiller:
Thank you @ashawley!

_tour/tour-of-scala.md
```diff
@@ -19,7 +19,7 @@ Scala is a modern multi-paradigm programming language designed to express common
...
+Scala is also a functional language in the sense that [every function is a value](unified-types.html). Scala provides a [lightweight syntax](basic.html#functions) for defining anonymous functions, it supports [higher-order functions](higher-order-functions.html), it allows functions to be [nested](nested-functions.html), and supports [currying](currying.html). Scala's [case classes](case-classes.html) and its built-in support for [pattern matching](pattern-matching.html) model algebraic types used in many functional programming languages. [Singleton objects](singleton-objects.html) provide a convenient way to group functions that aren't members of a class.
```
ashawley: Oops, should be `basics.html#functions`

I'll get that in a follow-up commit.

I guess one downside of long lines, is that GitHub refuses to highlight the character diffs after some threshold?

PR #854, opened 2017-08-08
Fix 404 links

ashawley:
Wget can check for bad links pretty easily with

```
$ wget -X /ba,/es,/ja,/ko,/pl,/pt-br,/sips --recursive --spider http://localhost:4000
```


heathermiller:
Thank you! 

I've also been using a Jekyll plugin for this too (locally):
https://github.com/scala/scala.github.com/blob/e8ee124c1310ed93819331a06683067d6a13f4fc/Gemfile#L3

I think you're already familiar with it, no? In any case, it finds a lot, including broken images. Anyway, there's more work to do on this one...

ashawley:
Cool, that's handy.  I had not heard of html-proofer.  It even does other stuff besides link checking.  I'll try it out.  The benefit of wget is its spidering takes like 2 seconds to check the whole site locally.

PR #855, opened 2017-08-08
Changes to SIP33

soronpo:
* Numbered the SIP, as given by the committee.
* Fixed bad view of code snippets.
* Added feature interaction with repeated parameters.
* Added related issues.
* Fixed comment in infix code snippet (between Left expansion and Right expansion)


heathermiller:
🎉 Merci!

PR #859, opened 2017-08-11
Fix code style by removing `:fail` from lower-type-bounds.md

ayurchuk:
Existing documentation layout:
screen shot 2017-08-11 at 13 58 58
screen shot 2017-08-11 at 14 53 19
Expected one:
screen shot 2017-08-11 at 13 59 50
screen shot 2017-08-11 at 14 54 52

### DIFF
--- a/_tour/lower-type-bounds.md
+++ b/_tour/lower-type-bounds.md
@@ -18,7 +18,7 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
 
 Here is an example where this is useful:
 
-```tut:fail
+```tut
 trait Node[+B] {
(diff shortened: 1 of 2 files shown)


heathermiller:
Ah! Thanks for catching this. Actually, I think the solution is for me to add `tut:fail` to this silly little Jekyll plugin I made so that this is correctly rewritten to a `scala` fenced code block after tut runs... Hang on, let me see...

heathermiller:
@ayurchuk I think I got it! Fixed it in this little Jekyll plugin we have for making syntax highlighting place nicely with tut. See #860 

Thank you so much for reporting this and for taking a first step at a fix! It's sincerely appreciated! ❤️

PR #861, opened 2017-08-15
Update unified-types.md

lierdakil:
Removed a part which probably was left hanging there after an earlier edit


heathermiller:
Thank you!

heathermiller:
❤️

Issue #862, opened 2017-08-15
tour lists both "For Comprehensions" and "Sequence Comprehensions"


lierdakil:
... which are basically the same. One of those probably should be removed.

I should mention though, that supposed "new version" -- "for comprehensions" -- doesn't mention that comprehensions without `yield` are (a) allowed and (b) return `()`.

PR #863, opened 2017-08-15
Remove old "sequence comprehensions" in tour

lierdakil:
Fixes https://github.com/scala/docs.scala-lang/issues/862

Note: this also re-introduces stuff that was in "sequence comprehensions" back into "for comprehensions". I'm not familiar with scala (that's why I'm reading the tour in the first place), so example might be horrifying. Sorry if that's the case.


heathermiller:
Ah, thanks for catching this! I presume it has something to do with the unfinished rewrite of the Scala Tour that was recently done :/ I'll have to have a deeper look to see what happened; indeed there shouldn't be 2 documents laying around.

SethTisue:
nice catch — thanks!

PR #865, opened 2017-08-15
fixes for tour/Lower type bounds

lierdakil:
Firstly, `Node.prepend` should have a return type annotation, since it's assumed to be `Unit` if not specified, and that's not what one would want *at all* here.

Secondly, having swallows extend from `Mammal` seems wrong on too many levels.


heathermiller:
Hahahahahaha 🤣

heathermiller:
Thank you :)

PR #871, opened 2017-08-28
Update mixin-class-composition.md

SirMullich:
`Iter` should be replaced with `RichStringIter` or other way around. Class name in code is not consistent with class name in explanation. Also, I suggest using some literal string instead of `args(0)` because it is not obvious for beginners what `args(0)` is.


SethTisue:
thank you for this! as it turned out, though, I think #873 is a better fix

PR #873, opened 2017-08-29
Change "Iter" class name for "RichStringIter"

voidance:
The comments refer to the "RichStringIter" class name which I guess makes more sense for the code example given than "Iter".


SirMullich:
You spotted the same mistake as me :) 
[https://github.com/scala/docs.scala-lang/pull/871](#871)

SethTisue:
thank you!

PR #874, opened 2017-08-29
Minor typo correction

voidance:


heathermiller:
Good catch! Thanks so much! 🎉

PR #878, opened 2017-09-07
Make code readable in the web version

shepard8:
Wrapped the comments at 80 characters so that the code is readable in the web documentation. (The code was clipped at about 90 characters and it was very uncomfortable to manually slide horizontally).


shepard8:
My bad for not checking first if this hadn't already been done.

Indeed,  #866 already fixes this. I close.

PR #879, opened 2017-09-07
"Message" not in agreement with example

AlvinOscar:
The statement implies that "Message" was in the previous example, but the class in the example is "Book"


heathermiller:
Thank you so much for catching this @AlvinOscar!!

Issue #621, opened 2016-11-09
A guide to binary compatibility for library authors


lrytz:
It would be great to have a guide about binary compatibility for library authors: what is a good policy to follow for your libraries, how to use version numbers, how to check binary compatibility using MiMa, details on what kind of changes are / aren't binary compatible.

sjrd:
I've wanted to write such a guide for quite some time, now. But as with all things, time is lacking. I'd be happy to point a few ideas to someone motivated, though, as well as review such a guide.

jvican:
I think this is a fantastic idea and everyone in the ecosystem would benefit from it. I would extend it to source compatibility also. Despite not having a tool to check it, it'd be good to give developers some rule of thumbs and guidelines.

zjffdu:
Any update on this ? I also need some guidance on binary compatibility. Thanks

heathermiller:
What about this? http://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html

heathermiller:
So far, no one has stepped up to do this. 

All we have at the moment is a document which discusses [binary compatibility of Scala releases](http://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html). Not yet something specifically aimed at authors of libraries. 

However, maybe info from this guide is useful for general library authors? @sjrd @lrytz could you add a few sentences explaining what you had in mind for a document aimed specifically at library authors?

sjrd:
There's a bunch of stuff to say about the topic.

* Explain why binary compatibility is an issue for library authors. Basically explain the diamond problem (in the dependency graph) and dependency hell when libraries evolve in binary incompatible ways.
* Talk about how language features impact binary compatibility: fields in traits, case classes, default parameters, type inference of result types, etc.
* Understand how to use MiMa, and how to interpret its output. Be able to diagnose whether its error messages can be safely filtered out or not.
* Give strategies on how to define public APIs so that they can be later evolved in source and binary compatible ways: avoid default parameters (prefer overloads), avoid case classes unless they are really meant for pattern matching (in which case they're frozen forever), seal/finalize everything possible, especially traits, use abstract classes instead of traits if new fields are expected in the future, etc.

There's *a lot* to say about this.

heathermiller:
Thanks for the list, Séb. It's really helpful for someone to potentially use as a starting point.

jvican:
I may step up to do this if someone does not beat me to it.

[...]

PR #881, opened 2017-09-11
Guide on binary compatibility for library authors

jatcwang:
Added guide on binary compatibility, with library authors as the target audience. Addresses #621

After building you can see the guide on `http://localhost:4000/tutorials/binary-compatibility-for-library-authors.html`

TODOs:

* Able to navigate to the actual guide from the homepage
* "Designing for Evolution" section - I'm setting up an example repo to present this information with more detail (and actual code examples that show the breakages). Will link it in the main guide after I'm happy with it **DONE**

Let me know if there are things that I've missed. I'm also up for reorganization if you think it will help with the presentation/delivery.


_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+TODO
+
+## Versioning Scheme - Communicate binary compatiblity breakages
+
+We recommend using the following schemes to communicate binary and source compatibility to your users:
+
+* Any release with the same major version are **Binary Backwards Compatible** with each other
+* A minor version bump signals new features and **may contain minor source incompatibilities** that can be easily fixed by the end user
+* Patch version for bugfixes and minor behavioural changes
+* For **expreimental library versions** (where the major version is `0`, such as `v0.1.0`), a minor version bump **may contain both source and binary breakages**
```
ktoso: typo, experimental

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+Some examples:
+
+* `v1.0.0 -> v2.0.0` is <span style="color: red">binary incompatible</span>. Cares needs to be taken to make sure no evicted versions are still in the `v1.x.x` range to avoid runtime errors
+* `v1.0.0 -> v1.1.0` is <span style="color: blue">binary compatible</span> and maybe source incompatible
+* `v1.0.0 -> v1.0.1` is <span style="color: blue">binary compatible</span> and source compatible
+* `v0.4.0 -> v0.5.0` is <span style="color: red">binary incompatible</span> and maybe source incompatible
+* `v0.4.0 -> v0.4.1` is <span style="color: blue">binary compatible</span> and source compatible
+
+Many libraries in the Scala ecosystem has adopted this versioning scheme. A few examples are [Akka](http://doc.akka.io/docs/akka/2.5/scala/common/binary-compatibility-rules.html),
```
ktoso: re:Akka: Not really, the above explained scheme is much much weaker than what Akka guarantees.

Here it said:
`+* `v0.4.0 -> v0.5.0` is <span style="color: red">binary incompatible</span> and maybe source incompatible`
In Akka this would be fully binary compatible, we do not talk about source compatibility, but usually it would be kept as well. Only deprecated things (that were such for at least 2 full release cycle) could be removed in such release

ktoso: Ah, sorry! I see that you mean the `0.` prefix here.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+recompile our code without causing any compilation errors, `v2.0.0` is said to be source compatible with `v1.0.0`.
+
+Two library versions are said to be **Binary Compatible** if the compiled bytecode of these versions are compatible. Using the example above, removing a class will render two version
+binary incompatible too, as the compiled bytecode for v2.0.0 will no longer contain the removed class.
+
+When talking about two versions being compatible, the direction matters too. If we can use `v2.0.0` in place of `v1.0.0`, `v2.0.0` is said to be **backwards compatible** with `v1.0.0`. Conversely,
+if we say that any library release of `v1.x.x` will be forwards compatible, we can use `v1.0.0` anywhere where `v1.1.0` was originally used.
+For the rest of the guide, when the 'compatible' is used we mean backwards compatible, as it is the more common case of compatibility guarantee.
+
+An important note to make is that while breaking source compatibility normally results in breaking binary compatibility, they are actually orthorgonal
```
DaniRey: Typo: orthogonal

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Why binary compatibility matters
+
+Let's look at an example where binary incompatibility between versions of a library can have catastrophic consequences:
+
+Our application depends on library `A` and `B`. Both `A` and `B` depends on library `C`. Initially both `A` and `B` depends on `C v1.0.0`.
+
+![Initial dependency graph]({{ site.baseurl }}/resources/images/library-author-guide/before_update.png){: style="width: 280px; margin: auto; display: block;"}
+
+Some time later, we see `B v1.1.0` is now available and we upgraded the version in our `build.sbt`. Our code compiles and seems to work so we push it to production and goes home for dinner.
```
DaniRey: Times and Typo:
Some time later, we see `B v1.1.0` is available and upgrade its version in our `build.sbt`. Our code compiles and seems to work so we push it to production and go home for dinner.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+Why did we get a `NoSuchMethodError`? Remember that `A v1.0.0` is compiled with `C v1.0.0` and thus calls methods availble in `Cv1.0.0`. While `B` and
+our App has been recompiled with available classes/methods in `C v2.0.0`, `A v1.0.0`'s bytecode hasn't changed - it still calls the same method that is now missing in `C v2.0.0`!
+
+This situation can only be resolved by ensuring that the chosen version of `C` is binary compatible with all other evicted versions of `C`. In this case, we need a new version of `A` that depends
+on `C v2.0.0` (or any other future `C` version that is binary compatible with `C v2.0.0`).
+
+Now imagine if our App is more complex with lots of dependencies themselves depending on `C` (either directly or transitively) - it becomes extremely difficult to upgrade any dependencies because it now
+pulls in a version of `C` that is incompatible with the rest of the versions of `C` in our dependency tree! In the example below, we cannot upgrade `D` because it will transitively pull in `C v2.0.0`, causing breakages
+due to binary incompatibility. This inability to upgrade any packages without breaking anything is common known as **Dependency Hell**.
```
DaniRey: Typo: commonly

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+This is what we call **Binary Incompatibility Errors** - The bytecode interface used for compilation differs and is incompatible with the bytecode provided during runtime.
+
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+Since the classloader only loads the first match of a class, having multiple versions of the same library in the classpath is redundant. 
+Therefore when deciding which JARs to use for compilation, SBT only selects one version from each library (by default the highest), 
+and all other versions of the same library are **evicted**. When packaging applications, the same versions of libraries that was used for compiling the 
+application is packaged and used during runtime.
+
+Two library versions are said to be **Source Compatible** if switching one for the other does not incur any compile errors. For example, If we can switch from `v1.0.0` of a dependency to `v2.0.0` and
```
DaniRey: I would prefer if the examples would use the proposed versioning scheme. I think it's easier for the reader to follow and internalize it's rules.

jatcwang: Yeap good catch. That's my intention too.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+title: Binary Compatibility for library authors
+
+discourse: true
+permalink: /tutorials/:title.html
+---
+
+## Introduction
+
+A diverse and comprehensive set of libraries is important to any productive software ecosystem. While it is easy to develop and distribute Scala libraries, good library authorship goes far
+beyond just writing code and publishing them.
```
jvican: I'd write "beyond just writing code and publish it".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+In this guide, we will cover the important topic of **Binary Compatibility**:
+
+* How binary incompatibility can cause production failures in your applications
+* How library authors can avoid breaking binary compatibility, and/or convey breakages clearly to library users when they happen
```
jvican: I would split this last item in two:
* How library authors can avoid breakages in binary compatibility
* How to reason about the impact of binary incompatible changes

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+Before we start, first we need to understand how code is compiled and executed on the Java Virtual Machine (JVM).
```
jvican: <s>first we need to understand</s> -> let's understand

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+## The JVM execution model
+
+Code compiled to run on the JVM is compiled to a platform-independent format called **JVM bytecode** and stored in **Class File** format (with `.class` extension) and these class files are stored
```
jvican: <s>Code compiled to run on the JVM</s> -> Scala
<s>stored in **Class File** format</s> -> stored in `.class` files.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+in JAR files. The bytecode is what we refer to as the **Binary** format.
+
+When application or library code is compiled, their bytecode invokes named references of classes/methods from their dependencies instead of including the dependencies' actual bytecode
```
jvican: I prefer "When an application depends on a library, its compiled bytecode references to the library's bytecode"

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+(unless inlining is explicitly requested). During runtime, the JVM classloader will search through the provided class files for classes/methods referenced by and invoke them.
```
jvican: <s>unless inlining is explicitly request</s>. Things to note here:
* You're only talking about Scala's inlining.
* The previous sentence is true always. Bytecode dependencies are always references.
  
For clarity, I would encourage you to devote a new paragraph or section to describe what Scala's inlining is for, and how it breaks binary compatibility.

jvican: Also, I'd write "When a JVM application is started, the JVM classloads the application's dependencies and the application, in a process typically called (dynamic) linking. This is the process during which the JVm resolves the bytecode references between the application and its dependencies." instead of "During runtime, the JVM classloader will search through the provided class files for classes/methods referenced by and invoke them.".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+Let's illustrate with an example:
```
jvican: Let's illustrate **it** :smile:

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+in JAR files. The bytecode is what we refer to as the **Binary** format.
+
+When application or library code is compiled, their bytecode invokes named references of classes/methods from their dependencies instead of including the dependencies' actual bytecode
+(unless inlining is explicitly requested). During runtime, the JVM classloader will search through the provided class files for classes/methods referenced by and invoke them.
+
+Let's illustrate with an example:  
+
+We got an application `App` that depends on `A` which itself depends on library `C`. When starting the application we need to provide the class files
+for all of `App`, `A` and `C` (something like `java -cp App.jar:A.jar:C.jar:. MainClass`). If we did not provide `C.jar`, or if we provided a `C.jar` that does not contain certain classes/methods
+which `A` expected to exist in library `C`, we will get an exception when our code attempt to invoke the missing classes/methods.
```
jvican: we will get a **classloading** exception
when our code attempt**s**
  
It would be cool that you list the exceptions that users can get `InvocationTargetException`, `ClassNotFoundException`, `MethodNotFoundException` are the ones that come to mind now.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+This is what we call **Binary Incompatibility Errors** - The bytecode interface used for compilation differs and is incompatible with the bytecode provided during runtime.
```
jvican: Nitpick: use the em or the en dash, instead of `-`.
  
I would write:
"This is what we call **Binary Incompatibility Errors**. A error caused by binary incompatibility happens when the compiled class files references a name that cannot be resolved during runtime" (or even better: during class linking -- I think this is the way it's called formally, but don't remember.)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+Since the classloader only loads the first match of a class, having multiple versions of the same library in the classpath is redundant.
```
jvican: What about "When the JVM loads a given class, the first class file that is matched in the classpath is used. If another library of the classpath also has a class file with the same name, it'll be ignored. It is good practice to make sure that this case doesn't happen, since it can only cause trouble."?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+We got an application `App` that depends on `A` which itself depends on library `C`. When starting the application we need to provide the class files
+for all of `App`, `A` and `C` (something like `java -cp App.jar:A.jar:C.jar:. MainClass`). If we did not provide `C.jar`, or if we provided a `C.jar` that does not contain certain classes/methods
+which `A` expected to exist in library `C`, we will get an exception when our code attempt to invoke the missing classes/methods.
+
+This is what we call **Binary Incompatibility Errors** - The bytecode interface used for compilation differs and is incompatible with the bytecode provided during runtime.
+
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+Since the classloader only loads the first match of a class, having multiple versions of the same library in the classpath is redundant. 
+Therefore when deciding which JARs to use for compilation, SBT only selects one version from each library (by default the highest),
```
jvican: Uf, I wouldn't mention this. First, it doesn't give a lot of value. Second, it's factually not correct, since it can depend on the resolution tool and the semantics that you use (e.g. Ivy and Maven have different semantics and they are really complex...).

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: 👍

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+and all other versions of the same library are **evicted**. When packaging applications, the same versions of libraries that was used for compiling the 
+application is packaged and used during runtime.
+
+Two library versions are said to be **Source Compatible** if switching one for the other does not incur any compile errors. For example, If we can switch from `v1.0.0` of a dependency to `v2.0.0` and
+recompile our code without causing any compilation errors, `v2.0.0` is said to be source compatible with `v1.0.0`.
+
+Two library versions are said to be **Binary Compatible** if the compiled bytecode of these versions are compatible. Using the example above, removing a class will render two version
+binary incompatible too, as the compiled bytecode for v2.0.0 will no longer contain the removed class.
+
+When talking about two versions being compatible, the direction matters too. If we can use `v2.0.0` in place of `v1.0.0`, `v2.0.0` is said to be **backwards compatible** with `v1.0.0`. Conversely,
```
jvican: What about "Binary compatibility can be forwards and/or backwards." instead of  "When talking about two versions being compatible, the direction matters too."?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+if we say that any library release of `v1.x.x` will be forwards compatible, we can use `v1.0.0` anywhere where `v1.1.0` was originally used.
+For the rest of the guide, when the 'compatible' is used we mean backwards compatible, as it is the more common case of compatibility guarantee.
```
jvican: when the **word** 'compatible".
  
I think you should use 'binary compatible' instead of just plain 'compatible' :smile:

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+An important note to make is that while breaking source compatibility normally results in breaking binary compatibility, they are actually orthogonal
```
jvican: I think you can reword this with one of my previous suggestions and make this paragraph much nicer to read? 👍

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+if we say that any library release of `v1.x.x` will be forwards compatible, we can use `v1.0.0` anywhere where `v1.1.0` was originally used.
+For the rest of the guide, when the 'compatible' is used we mean backwards compatible, as it is the more common case of compatibility guarantee.
+
+An important note to make is that while breaking source compatibility normally results in breaking binary compatibility, they are actually orthogonal 
+(breaking one does not imply breaking the other). See below for more examples (TODO: make sure we have examples?)
+TODO: more facts?
+
+## Why binary compatibility matters
+
+Let's look at an example where binary incompatibility between versions of a library can have catastrophic consequences:
```
jvican: Answering the title first is the best IMO: "binary compatibility matters because it prevents an application's execution. This disruption can destroy ecosystems and force library maintainers to go through the process of updating their libraries so that users can continue using them".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+Our application depends on library `A` and `B`. Both `A` and `B` depends on library `C`. Initially both `A` and `B` depends on `C v1.0.0`.
+
+![Initial dependency graph]({{ site.baseurl }}/resources/images/library-author-guide/before_update.png){: style="width: 280px; margin: auto; display: block;"}
+
+Some time later, we see `B v1.1.0` is now available and we upgraded the version in our `build.sbt`. Our code compiles and seems to work so we push it to production and goes home for dinner.
+
+Unfortunately at 2am, we got frantic calls from customers saying that our App is broken! Looking at the logs, you find lots of `NoSuchMethodError` is being thrown by some code in `A`!
```
jvican: I would potentially remove this sentence (though I like the humouristic touch) and just say: "What can go wrong, right?".
  
And then you explain the error, as your following paragraphs do.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+![Binary incompatibility after upgrading]({{ site.baseurl }}/resources/images/library-author-guide/after_update.png){: style="width: 280px; margin: auto; display: block;"}
+
+Why did we get a `NoSuchMethodError`? Remember that `A v1.0.0` is compiled with `C v1.0.0` and thus calls methods available in `Cv1.0.0`. While `B` and
+our App has been recompiled with available classes/methods in `C v2.0.0`, `A v1.0.0`'s bytecode hasn't changed - it still calls the same method that is now missing in `C v2.0.0`!
+
+This situation can only be resolved by ensuring that the chosen version of `C` is binary compatible with all other evicted versions of `C`. In this case, we need a new version of `A` that depends
+on `C v2.0.0` (or any other future `C` version that is binary compatible with `C v2.0.0`).
+
+Now imagine if our App is more complex with lots of dependencies themselves depending on `C` (either directly or transitively) - it becomes extremely difficult to upgrade any dependencies because it now
```
jvican: I would make 'App' lowercase :smile:

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+When run standalone in the command line, it will compare the .class files in the two provided JARs and report any binary incompatibilities found. Most library authors use the [SBT plugin](https://github.com/typesafehub/migration-manager/wiki/Sbt-plugin)
+to help spot binary incompatibility between library releases. (Follow the link for instructions on how to use it in your project)
+
+## Designing for Evolution - without breaking binary compatibility
+
+TODO
+
+## Versioning Scheme - Communicate binary compatibility breakages
+
+We recommend using the following schemes to communicate binary and source compatibility to your users:
```
jvican: We **strongly"" recommend.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+The [Migration Manager for Scala](https://github.com/typesafehub/migration-manager) (MiMa) is a tool for diagnosing binary incompatibilities between different library versions.
+
+When run standalone in the command line, it will compare the .class files in the two provided JARs and report any binary incompatibilities found. Most library authors use the [SBT plugin](https://github.com/typesafehub/migration-manager/wiki/Sbt-plugin)
+to help spot binary incompatibility between library releases. (Follow the link for instructions on how to use it in your project)
+
+## Designing for Evolution - without breaking binary compatibility
+
+TODO
+
+## Versioning Scheme - Communicate binary compatibility breakages
```
jvican: What if you try to explain what a versioning scheme is and maybe quote the semantic versioning spec? I feel like a small introduction here would make the reading much much nicer for beginners.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+
+We recommend using the following schemes to communicate binary and source compatibility to your users:
+
+* Any release with the same major version are **Binary Backwards Compatible** with each other
```
jvican: Okay, this is how I propose you write this:

* Any library versioned with a starting `0` can break binary and source compatibility at any time, as of the [semantic versioning](link here) spec.
* A change in the **major version number** signals at least one **binary incompatibility**, either forwards and/or backwards. For example, code compiled against `v1.0.0` is not compatible with `v2.0.0`.
* A change in the **minor version number** signals at least one **source incompatibility**, either forwards and/or backwards. For example, code compiled against the sources of a dependency `v1.0.0` is not ensure to compile with `v1.1.0` of the same library.
* A change in the patch version (the `x` in `1.0.x`) is both binary and source compatible. Patch releases often include bug fixes and minor behavioural changes.
  
There may be cases in which libraries break source compatibility and decide to bump up the major version number. These cases, where not common, are defined per library.

jvican: I think the new format is much more readable and more "structured". The dumber the bullet lists are, the better ^^

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,140 @@
...
+* `v1.0.0 -> v1.0.1` is <span style="color: blue">binary compatible</span> and source compatible
+* `v0.4.0 -> v0.5.0` is <span style="color: red">binary incompatible</span> and maybe source incompatible
+* `v0.4.0 -> v0.4.1` is <span style="color: blue">binary compatible</span> and source compatible
+
+Many libraries in the Scala ecosystem has adopted this versioning scheme. A few examples are [Akka](http://doc.akka.io/docs/akka/2.5/scala/common/binary-compatibility-rules.html),
+[Cats](https://github.com/typelevel/cats#binary-compatibility-and-versioning) and [Scala.js](https://www.scala-js.org/).
+
+### Explanation
+
+Why do we use the major version number to signal binary compatibility releases?
```
jvican: to signal **binary incompatible** releases

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Or something along these lines... I just feel I need a clear answer to the title here ^^

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: You can then explain how binary compatibility is the absence of any binary incompatible error. And you can say that binary incompatible errors are backwards and forwards depending on the direction of the arrow. And even illustrate so with a small graph :smile:

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Damn, this comment is actually for the next paragraph. I reviewed the document but your last update offsetted it by one line.

jvican:
@jatcwang I'm really not a fan of the GitHub way of reviewing this. If you prefer it, I can add all my suggestions as a commit to this PR, and then you review my commit and change the things you don't like.
  
I feel that would be a much faster way of reviewing it.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: One thing is not clear to me is the status we give to forward binary compatibility. We should probably say that it's often the case that libraries do not promise forward binary compatibility, and that therefore they don't influence major version numbers. I'll think about this in the next hours since now I'm saturated, but perhaps @sjrd can shed some light on this. He's the original author of this proposal.

sjrd:
I will not have time this week to review this, because of a paper deadline. I will get to it next week.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: Forward compat is generally not an issue and does not count per se in the versioning number I am advocating.

However, checking forward binary compatibility with MiMa is the best approximation we have of guaranteed (100%) *backward source* compat. Indeed, the most common source for backward source incompatibility are additions of public fields/methods (which influence implicit resolution), and that is forbidden if you enforce forward binary compat. Therefore, I would in general encourage checking forward bin compat across patch releases (so between 1.2.x and 1.2.y), which is the best thing we have to enforce backward source compat in patch releases.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: That's better. One important point I want deliver is that application code is calling the "bytecode interface" of the library, which means modifying the internal of methods is OK and won't break bin-compat. What's the right technical term for this?

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Are there different levels of "impact` for binary incompatibility?

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: If I'm new to the JVM world in general, then I'd read that as "On startup, JVM will load all your classes and tell you if methods/classes are missing". However that's not true - the exceptions are thrown only when the missing class/method is invoked. I wanted to stress that fact. However I'm all for using the right terminologies

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Yeap I think your description is much more accurate. I'd not use "it is good practice" though as most of the time the tooling will take care of it for you by eviction.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: But this is not true for classes -- all classes that a given class uses are specified in the constant pool. If you try to load a class whose constant pool references to a class that has failed to resolve, the JVM will fail at that very same moment. You're right that this is not the case for methods, which have a more "lazy" resolution.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Yeap I'm aware that there are different semantics between tools, but can you customize the classloader behaviour? My understanding is that resolvers decides the list of JARs to put on the classpath, but classloaders behaves the same (which is why I specifically mentions SBT)

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: I would call it "binary interface" though I don't know the academic term. If you want to make this more obvious, I would create a new paragraph (or even better: enclose it in a <note> -- there are some other blog posts that do it).

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Feel free to make the edit - I can draw the small graph :)

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Not really. In hindsight, that last bullet point doesn't explain what I wanted to. Let me reword it: "How to reason about the impact of changes in their code". And I would put it in the second position instead of the last one.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: `App` is the name of our application, just like `A`, `B` are the names of the libraries. Can wrap it in inline-code it so it's more obvious.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I didn't want to mention semantic versioning at all because it brings in the wrong correlation and might confuse the reader. 

`MAJOR version when you make incompatible API changes,` 

In the versioning scheme we're recommending, major version bump has nothing to do with API changes. (From my point of view, API changes = source incompatibility)

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: > My understanding is that resolvers decides the list of JARs to put on the classpath, but classloaders behaves the same
  
Yeah, and this is correct. What I meant is that the resolution changes the order of the artifacts, and therefore the semantics. I wouldn't mention sbt because sbt is not the only Scala build tool. If I were you, I would create a new section for "evictions" that explains it in a more broad sense, and then I would explain that "build tools usually take care of resolving evictions automatically if they can -- otherwise, they delegate this task to the user, who in a conflict has to choose which library to use".

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: but yeah, we can briefly explain what versioning schemes are

jvican: > I didn't want to mention semantic versioning at all because it brings in the wrong correlation and might confuse the reader.
  
I'm not sure about this. The versioning scheme that we recommend **is** a more constrained subset of semantic versioning. I feel mentioning semantic versioning here is a must, it's a widely used standard. We should make it obvious that our versioning scheme enforces more information in every Scala version, which is extremely advantageous.

> (From my point of view, API changes = source incompatibility)
  
Not really, API changes happen because of both binary and source incompatible changes.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: * I agree with @sjrd regarding forward compatibility not being as relevant for most userland libraries
* I am still on the fence about minor versions. I know a lot of libraries like to use minor version to signal new features. Perhaps we make minor version a mandatory bump if forward compatibility is broken? (which normally coincides with adding new functionality like new methods).
* Does the SBT MiMa plugin currently support easy checking for forward compatibility?
* Just an idea: I think we should extend MiMa SBT plugin to automatically suggest a new version number based on the result of compatibility check. (Although I'm not sure if there's a way to check source compatibility..)

jatcwang:
@jvican That'll be awesome thank you. I've left comments on all of your feedback but feel free to make changes as you see fit.

jatcwang:
I have just uploaded a repo to demonstrate common causes of binary incompability + more detailed explanation that I can't cover in the main guide. The intention is to provide a link to this repo from the main guide.

https://github.com/jatcwang/binary-compatibility-guide

Please raise issues in that repo for anything you think I've missed.

@jvican I am happy to update my PR to address your feedback so let me know if you're struggling to find time.

jvican:
I'm struggling to find time atm, but I'll review further updates on this PR.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Is that true? It still seems to delay loading a class until instantiation is required.

I tested with the following code:

```
package foo
import bar.Bar

object Main extends App {
  println("hello")
  val foo = new Foo("hello")
  foo.bar
}

class Foo(value: String) {
  def bar: Bar = new Bar()
}

```
(+ empty class Bar in another file)

If I remove the `foo.bar` call, no exception is thrown despite `Bar` being referenced in the method signature. At least that's how I interpret "a given class uses are specified"

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I am using `this`, does it makes more sense in context (with the updated section)?

jatcwang:
Pushed some commits that hopefully addresses all your feedback @jvican. (Much appreciated - it reads better now)

heathermiller:
🎉 🎉 
@jvican can you have a look at this when you're back in the office?

And again, sincerest thanks for taking the initiative on this @jatcwang ❤️

heathermiller:
@sjrd too, if you're interested in/have time to provide a review, I know it'd be appreciated.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: You're right @jatcwang, TIL this. I thought lazy loading only applied to methods and not classes.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Sure, it works.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+
+![Forwards and Backwards Compatibility]({{ site.baseurl }}/resources/images/library-author-guide/fowards_backwards_compatibility.png){: style="width: 65%; margin: auto; display: block"}
+
+`C v1.1.0 ` is **Forwards Binary Compatible** with `v1.0.0` if we can use `v1.0.0`'s JAR at runtime instead of `v1.1.0`'s JAR without any binary compatibility errors.
+
+`C v1.2.0 ` is **Backwards Binary Compatible** with `v1.1.0` if we can use `v1.2.0`'s JAR at runtime instaed of `v1.1.0`'s JAR without any binary compatibility errors..
+
+## Why binary compatibility matters
+
+Binary Compatibility matters because failing to maintain it makes life hard for everyone.
```
jvican: What about "Breaking binary compatibility has bad consequences on the ecosystem around the software:"?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+
+* End users has to update all library versions in their whole transitive dependency tree such that they are binary compatible, otherwise binary compatibility errors will happen at runtime
```
jvican: <s>has</s> have

jvican: What about "have to update versions transitively in all their dependency tree such that they are binary compatible. This process is time-consuming and error prone, and it can change the semantics of end program."?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+* Library authors are forced to update the dependencies of their library so users can continue using them, greatly increases the effort required to maintain libraries
```
jvican: I think this part is not clear enough. What do you mean exactly?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+
+Constant binary compatibility breakages in libraries, especially ones that are used by other libraries, is detrimental to our ecosystem as they require a lot of effort
```
jvican: "as they require time and effort from end users and maintainers of dependent libraries"

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+
+### An example of "Dependency Hell"
+
+Our application `App` depends on library `A` and `B`. Both `A` and `B` depends on library `C`. Initially both `A` and `B` depends on `C v1.0.0`.
+
+![Initial dependency graph]({{ site.baseurl }}/resources/images/library-author-guide/before_update.png){: style="width: 50%; margin: auto; display: block;"}
+
+Some time later, we see `B v1.1.0` is available and upgrade its version in our build. Our code compiles and seems to work so we push it to production and go home for dinner.
+
+Unfortunately at 2am, we got frantic calls from customers saying that our application is broken! Looking at the logs, you find lots of `NoSuchMethodError` is being thrown by some code in `A`!
```
jvican: <s>got</s>get

jvican: <s>is</s>are being thrown

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+Common exceptions from classloading failures includes 
+`InvocationTargetException`, `ClassNotFoundException`, `MethodNotFoundException`, and `AbstractMethodError`.
+
+Let's illustrate this with an example:
+
+Consider an application `App` that depends on `A` which itself depends on library `C`. When starting the application we need to provide the class files
+for all of `App`, `A` and `C` (something like `java -cp App.jar:A.jar:C.jar:. MainClass`). If we did not provide `C.jar` or if we provided a `C.jar` that does not contain some classes/methods
+which `A` calls, we will get classloading exceptions when our code attempts to invoke the missing classes/methods.
+
+These are what we call **Binary Incompatibility Errors**. An error caused by binary incompatibility happens when the compiled bytecode references a name that cannot be resolved during runtime
```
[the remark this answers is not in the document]
jatcwang: I have no idea what you're talking so I just rewrote it and add the missing full stop at the end 😛

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,211 @@
...
+
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+### Evictions
+When a class is needed during execution, the JVM classloader loads the first matching class file from the classpath (any other matching class files are ignored).
+Because of this, having multiple versions of the same library in the classpath is generally undesireable:
+
+* Unnecessary application size increase
```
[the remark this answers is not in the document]
jatcwang: If the build tool bundles all different version of a library with the application (however it is distributed), then you have unnecessary bloat in your bundle. Happy to reword or remove if you find it confusing

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: reworded if that is clearer? mainly talking about users asking library authors to bump the dependency version to align the versions to be bin-compat

jatcwang:
Updated according to @jvican's second round of review (thanks!) + grammar check.

_overviews/tutorials/binary-compatibility-for-library-authors.md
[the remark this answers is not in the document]
jvican: Yes, I think that it requires some clarification :smile:. Also, consider adding my remark, I think it's valid.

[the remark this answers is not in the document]
jatcwang: updated :)

heathermiller:
@jvican... Can you please have a look at the rest?

heathermiller:
(PS, thanks very much @jatcwang for your patience on this! And thank you for the fixes!)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+How can we, as library authors, spare our users of runtime errors and dependency hell?
+
+* Use **Migration Manager** (MiMa) to catch unintended binary compatibility breakages before releasing a new library version
+* **Avoid breaking binary compatibility** through careful design and evolution of your library interfaces
+* Communicate binary compatibility breakages clearly through **versioning**
+
+## MiMa - Checking binary compatibility against previous library versions 
+
+The [Migration Manager for Scala](https://github.com/typesafehub/migration-manager) (MiMa) is a tool for diagnosing binary incompatibilities between different library versions.  
+It works by comparing the class files of two provided JARs and report any binary incompatibilities found.
```
jvican: I would try to make more explicit that it support forwards and backwards bincompat detection with something like this: "It can tell a library author whether a change causes any kind of binary compatibility in both directions: forwards and backwards."

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+By incorporating [MiMa SBT plugin](https://github.com/typesafehub/migration-manager/wiki/Sbt-plugin) into your SBT build, you can easily check whether 
+you have accidentally introduced binary incompatible changes. Detailed instruction on how to use the SBT plugin can be found in the link.
+
+We strongly encourage every library author to incorporate MiMa into their library release workflow.
+
```
jvican: Someone at some point mentioned that MiMa can be used as the best approximation of source (in)compatibility detection. I think it was a comment in this very same PR.
  
It may be interesting mentioning that here and also explaining that detecting source incompatibility automatically in Scala is extremely difficult because of complex scoping and certain language features like implicits, named parameters, etc.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* Case classes
+* Default methods on traits (doesn't cause breakages since 2.12)
+
+Techniques you can use to avoid breaking binary compatibility:
+
+* Annotate public method's return type explicitly
+* Mark methods as package private when you want to remove a method or modify its signature
+* Don't turn on inlining for methods from dependencies
+
+For brevity of this guide, detailed explanation and runnable code examples can be found in [Binary Compatibility Code Examples & Explanation](https://github.com/jatcwang/binary-compatibility-guide).
```
jvican: That repo is great :smile:. Thanks for putting that together.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+users to easily reason about the impact of updating a library, without needing to read the detailed release note.
+
+In the following section, we will outline a versioning scheme based on Semantic Versioning that we **strongly encourage** you to adopt for your libraries. The rules listed below are **in addition** to 
+Semantic Versioning v2.0.0.
+
+### Recommended Versioning Scheme
+
+* If backward **binary compatibility** is broken, **major version number** must be increased
+* If backward **source compatibility** is broken, **minor version number** must be increased
+* A change in **patch version number** signals **no binary nor source incompatibility**. According to SemVer, patch versions should contain only bug fixes that fix incorrect behavior so major behavioral
```
jvican: I would add to this point: "but in this scheme we recommend adding any change that is source and binary compatible" to emphasize that we don't only want the patch version number to be used for bugfixes.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+Library authors use versioning schemes to communicate compatibility guarantees between library releases to their users. Versioning schemes like [Semantic Versioning](http://semver.org/)(SemVer) allow
+users to easily reason about the impact of updating a library, without needing to read the detailed release note.
+
+In the following section, we will outline a versioning scheme based on Semantic Versioning that we **strongly encourage** you to adopt for your libraries. The rules listed below are **in addition** to 
+Semantic Versioning v2.0.0.
+
+### Recommended Versioning Scheme
+
+* If backward **binary compatibility** is broken, **major version number** must be increased
```
jvican: You miss a '.' at the end of every item in the enumeration. I think it's a convention to put either ';` or `.` instead of not ending the sentence with any punctuation mark.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* If backward **source compatibility** is broken, **minor version number** must be increased
+* A change in **patch version number** signals **no binary nor source incompatibility**. According to SemVer, patch versions should contain only bug fixes that fix incorrect behavior so major behavioral
+change in method/classes should result in a minor version bump.
+* When major version is `0`, a minor version bump **may contain both source and binary breakages**
+* Some libraries may take a harder stance on maintaining source compatibility, bumping the major version number for ANY source incompatibility even if they are binary compatible
```
jvican: This is not a recommendation, just a clarification, so I think it should be in an independent paragraph. The gist would be: "being stricter than this scheme is possible and encouraged when possible, though most of the times it's not required; here's how...".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+Some examples:
+
+* `v1.0.0 -> v2.0.0` is <span style="color: red">binary incompatible</span>. Cares needs to be taken to make sure no evicted versions are still in the `v1.x.x` range to avoid runtime errors
```
jvican: Let's enumerate here the consequences of using every version. I suggest removing "Cares needs to be taken to make sure evicted versions are still in the `v1.x.x` range to avoid runtime errors" by clearer consequences: "End users and library maintainers need to update all their dependency graph to `2.0.0`. ". I would do the same for every other version transition; what do you think?
  
For example, the last item in the enumeratino would have "Users and library authors can migrate to this version without any API problem".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* `v1.0.0 -> v1.1.0` is <span style="color: blue">binary compatible</span> and maybe source incompatible
+* `v1.0.0 -> v1.0.1` is <span style="color: blue">binary compatible</span> and source compatible
+* `v0.4.0 -> v0.5.0` is <span style="color: red">binary incompatible</span> and maybe source incompatible
+* `v0.4.0 -> v0.4.1` is <span style="color: blue">binary compatible</span> and source compatible
+
+Many libraries in the Scala ecosystem has adopted this versioning scheme. A few examples are [Akka](http://doc.akka.io/docs/akka/2.5/scala/common/binary-compatibility-rules.html),
+[Cats](https://github.com/typelevel/cats#binary-compatibility-and-versioning) and [Scala.js](https://www.scala-js.org/).
+
+If this version scheme is followed, reasoning about binary compatibility is now very simple:
```
jvican: <s>very simple</s> simpler

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Ensure major versions of all versions of a library in the dependency tree are the same
```
jvican: What do you mean by this? Can you reword so that's it's clearer? It gives me the impression this is not one of the pains that are now easier with the recommended scheme.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* Pick the latest version and evict the rest (This is the default behavior of SBT).
```
jvican: Why do we mention this here? How is this important to end users implementing this scheme?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
```
jvican: In my opinion, this is one of the most important parts of the document and we need to make a good case to synthesize all the advantages to use this scheme over others. To me, the biggest advantage to users is determinism: it's clearer the consequences of making a upgrading/downgrading a library in their projects. They can reason better about the changes produced in every release. They express the rich kind of incompatibilities in Scala.
  
Do you think that adding these instead of the ones you mention is better?

jvican: I see you do a similar reasoning in "Explanation". I think it may be worth considering merging this with the "Conclusion" section instead of putting it here. Do you agree?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+### Explanation
+
+Why do we use the major version number to signal binary incompatible releases?
+
+From our [example](#why-binary-compatibility-matters) above, we have learned two important lessons:
+
+* Binary incompatibility releases often leads to dependency hell, rendering your users unable to update any of their libraries without breaking their application
```
jvican: Not only binary. I would say: "API incompatibilities often lead to dependency hell".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* If a new library version is binary compatible but source incompatible, the user can simply fix the compile errors in their application and everything will work
```
jvican: <s>simply</s>

jvican: s/everything will work/their application should work/g

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+Therefore, **binary incompatible releases should be avoided if possible** and be more noticeable when they happen, warranting the use of the major version number. While source compatibility
```
jvican: <s>be more noticeable</s> be "properly documented"

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+is also important, if they are minor breakages that do not require effort to fix, then it is best to let the major number signal just binary compatibility.
```
jvican: <s>if they are minor breakages that do not require effort to fix</s>
  
The previous statement is true no matter whether that conditional statement is true or not. The scheme signals the most severe API change, in this case breaking binary compatibility, and users should be prepared for source incompatible changes too.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+These are what we call **Binary Incompatibility Errors** -- errors that happen when the compiled bytecode references a name that cannot be resolved during runtime.
+
+Before we look at how to avoid binary incompatibility errors, let us first
+establish some key terminologies we will be using for the rest of the guide.
+
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+### Evictions
+When a class is needed during execution, the JVM classloader loads the first matching class file from the classpath (any other matching class files are ignored).
+Because of this, having multiple versions of the same library in the classpath is generally undesirable:
```
sjrd: This mixes up different things. Consider replacing that paragraph with something more obviously accurate, such as

> Unless using special tools such as OSGI), only one version of any given library can be loaded in the same JVM. If several are provided on the classpath, the contents of the first one will shadow the later ones.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Need to fetch and bundle multiple library versions when only one is actually used
+* Unexpected runtime behavior if the order of class files changes
+
+Therefore, when resolving JARs to use for compilation and packaging, most build tools will pick only one version of each library and **evict** the rest.
+
+### Source Compatibility
+Two library versions are **Source Compatible** if switching one for the other does not incur any compile errors or unintended behavioral changes at runtime.
```
sjrd: source compatible <ins>with each other</ins>

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+* Unexpected runtime behavior if the order of class files changes
+
+Therefore, when resolving JARs to use for compilation and packaging, most build tools will pick only one version of each library and **evict** the rest.
+
+### Source Compatibility
+Two library versions are **Source Compatible** if switching one for the other does not incur any compile errors or unintended behavioral changes at runtime.  
+For example, If we can upgrade `v1.0.0` of a dependency to `v1.1.0` and recompile our code without any compilation errors, `v1.1.0` is source compatible with `v1.0.0`.
+
+### Binary Compatibility
+Two library versions are **Binary Compatible** if the compiled bytecode of these versions can be interchanged without causing binary compatibility errors.
```
sjrd: binary compatible <ins>with each other</ins>

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+### Source Compatibility
+Two library versions are **Source Compatible** if switching one for the other does not incur any compile errors or unintended behavioral changes at runtime.  
+For example, If we can upgrade `v1.0.0` of a dependency to `v1.1.0` and recompile our code without any compilation errors, `v1.1.0` is source compatible with `v1.0.0`.
+
+### Binary Compatibility
+Two library versions are **Binary Compatible** if the compiled bytecode of these versions can be interchanged without causing binary compatibility errors.  
+For example, if we can replace the class files of a library's `v1.0.0` with the class files of `v1.1.0` without any binary compatibility errors during runtime,
+`v1.1.0` is binary compatible with `v1.0.0`.
+
+**NOTE:** While breaking source compatibility normally results in binary compatibility breakages as well, they are actually orthogonal -- breaking one does not imply breaking the other.
```
sjrd: I don't think this is the best place to mention that. It disrupts the flow of explaining concepts with something that is thought-provoking and confusing to people. It should come later, and with an example.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Forwards and Backwards Compatibility
+
+There are two "directions" when we describe compatibility of a library release:
+
+**Backwards Compatible** means that a newer library version can be used in an environment where an older version is expected. When talking about binary and source compatibility,
+this is the common and implied direction.
+
+**Forwards Compatible** means that an older library can be used in an environment where a newer version is expected.  
+Forward compatibility is generally not upheld for userland libraries. It is only important in situations where an older version of a library is commonly used at runtime against code that is compiled with newer versions. (for example, [Scala's standard library](http://docs.scala-lang.org/overviews/core/binary-compatibility-of-scala-releases.html))
```
sjrd: Why/where is the Scala standard library more often replaced by an earlier version than other libraries?

Answer: it isn't. It's just that the Scala std lib has very broad exposure, and they want to support that feature for users who do put themselves in situations where they do that. However, those users would replace *other libraries* with earlier versions just as much as the Scala std lib. There is nothing special about the Scala std lib here.

The only well-known example in the "open" community around Scala is sbt: sbt loads itself with its declared version of Scala, and a bunch of dependencies (such as Zinc). It will then classloads sbt plugins, which will inherit from that hard-coded version of the Scala std lib *and any other dependency of sbt*, no matter what the sbt plugin was compiled with. (At least, I'm pretty sure that's what happens in sbt; I haven't actually tested that hypothesis. Maybe @jvican knows better.)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+## MiMa - Checking binary compatibility against previous library versions 
+
+The [Migration Manager for Scala](https://github.com/typesafehub/migration-manager) (MiMa) is a tool for diagnosing binary incompatibilities between different library versions.  
+It works by comparing the class files of two provided JARs and report any binary incompatibilities found.
+
+By incorporating [MiMa SBT plugin](https://github.com/typesafehub/migration-manager/wiki/Sbt-plugin) into your SBT build, you can easily check whether 
+you have accidentally introduced binary incompatible changes. Detailed instruction on how to use the SBT plugin can be found in the link.
+
+We strongly encourage every library author to incorporate MiMa into their library release workflow.
```
sjrd: their library release workflow  -> the CI scripts of their library

Doing that at the time of releasing is much too late.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: I probably said that.

Basically, checking *both* forward and backward bincompat with MiMa is the closest tool we've got to checking backward source compatibility. Because it prevents, for example, from adding public methods, which is one of those things that can break source compat.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Evolving code without breaking binary compatibility
+
+Binary compatibility breakages can often be avoided through careful use of certain Scala features as well as some techniques you can apply when modifying code.
+
+Some language features may break binary compatibility:
```
sjrd: The features themselves don't break binary compatibility. They easily allow for scenarios where "obviously" (supposedly, in fact) backward source compatible changes actually result in catastrophic binary incompatible changes.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,215 @@
...
+
+* Default parameter values for methods or classes
+* Case classes
+* Default methods on traits (doesn't cause breakages since 2.12)
+
+Techniques you can use to avoid breaking binary compatibility:
+
+* Annotate public method's return type explicitly
+* Mark methods as package private when you want to remove a method or modify its signature
+* Don't turn on inlining for methods from dependencies
```
sjrd: I don't think this list here brings any value. They require examples and explanations for them to make any sense. Just send to the other guide and don't mention this list here.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: I would have comments to give to the readme of that other repo, but I don't have any easy interface to give comments online like I have with a PR review. What should I do?

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: In English typography, there are no hard rules on the proper use of bulleted lists.

The only well-accepted rule I have seen is that, if the full list can read as a full sentence, you can write the list as if bullets were basically not semantically relevant. Therefore, the rest of the punctuation and capitalization should be applied as if they were all part of the same sentence. This means that a bullet list should look as follows:
* items should not start with a capital letter,
* they should all end with a comma, and
* the next-to-last item ends with ", and" while the last one ends with a period.

Indeed, you can remove the bullets from the above list, and it still reads as a correct sentence.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: Depending on how serious you are about source compatibility, a backward source compatible release must have exactly the same public/protected API as the previous version. It cannot even add methods. Given that SemVer itself recommends that *adding functionality* should bump MINOR, it's pretty clear to me that adding a public method should bump MINOR anyway.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: Bumping the major version for ANY source incompatibility is probably nonsense. If you really mean *ANY* (and I guess you mean it, because it's uppercase), then adding a public method means you need to bump MAJOR. I don't any library will ever be that strict.

Actually, I would argue that a library doing this is doing a *disservice* to its community, as it has no means to effectively convey *binary incompatible changes* anymore in a way that is independent from source incompatible changes. This means that every single new public method breaks the ecosystem, since I cannot know from the version number whether I am moving into dependency hell or not.

sjrd: MAJOR could be bumped for "severe" source breaking changes (for some definition of "severe"). Removing public APIs could be considered "severe", for example. But we cannot give any objective metric about this.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sjrd: No, this is not accurate. That's the whole point of having bin compat in MAJOR and source compat in MINOR. Source incompatibility does ***NOT*** result in dependency hell. Only binary incompat does.

I think I explained that at length, and pretty well, in the Contributors forum at some point.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Cool, thanks for letting me know! Myself I've always used `;` and then, for the last item in an enumeration, `; and,`. I'll use comma from now on.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Correct, I don't know why I said otherwise in the beginning.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: I also had comments and I created a new ticket in the issue tracker.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: Yes, I can confirm this. I brought it up in our last discussion of forward binary compatibility https://contributors.scala-lang.org/t/question-about-changes-in-the-scalac-release-cycle/1093/6. You explained it correctly. In reality, what happens is that global plugins and local plugins are classloaded in independent classloaders whose parent is the sbt classloader, but this does not change the behaviour that we're describing here.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: Yeap just create tickets :)

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I'm gonna use full stops as they are all full sentences

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I think I'll remove this paragraph to avoid confusion.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I'm just gonna take out this sentence. I feel like I'm overexplaining at various places when the reader can simply figure it out later on their own

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: What's inaccurate about the paragraph? I want to minimize the amount of keywords mentioned especially when they doesn't help the key message of the guide (e.g. OSGI)

jatcwang:
Changes pushed. I addressed everything apart from @sjrd's comment about JVM classloading and bullet points 😛

jvican:
@jatcwang Thanks :heart:.

Can you check if the document is in good shape now? @sjrd

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,221 @@
...
+
+How can we, as library authors, spare our users of runtime errors and dependency hell?
+
+* Use **Migration Manager** (MiMa) to catch unintended binary compatibility breakages before releasing a new library version
+* **Avoid breaking binary compatibility** through careful design and evolution of your library interfaces
+* Communicate binary compatibility breakages clearly through **versioning**
+
+## MiMa - Checking binary compatibility against previous library versions 
+
+The [Migration Manager for Scala](https://github.com/typesafehub/migration-manager) (MiMa) is a tool for diagnosing binary incompatibilities between different library versions.
```
gmethvin: This is now https://github.com/lightbend/migration-manager

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,221 @@
...
+users to easily reason about the impact of updating a library, without needing to read the detailed release note.
+
+In the following section, we will outline a versioning scheme based on Semantic Versioning that we **strongly encourage** you to adopt for your libraries. The rules listed below are **in addition** to 
+Semantic Versioning v2.0.0.
+
+### Recommended Versioning Scheme
+
+* If backward **binary compatibility** is broken, **major version number** must be increased.
+* If backward **source compatibility** is broken, **minor version number** must be increased.
+* A change in **patch version number** signals **no binary nor source incompatibility**.
```
gmethvin: The wording is is a bit weird here. This should be "neither binary nor source incompatibility" or "no binary or source incompatibility".

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Versioning Scheme - Communicating compatibility breakages
+
+Library authors use versioning schemes to communicate compatibility guarantees between library releases to their users. Versioning schemes like [Semantic Versioning](http://semver.org/)(SemVer) allow
+users to easily reason about the impact of updating a library, without needing to read the detailed release note.
+
+In the following section, we will outline a versioning scheme based on Semantic Versioning that we **strongly encourage** you to adopt for your libraries. The rules listed below are **in addition** to 
+Semantic Versioning v2.0.0.
+
+### Recommended Versioning Scheme
```
gmethvin: I would also like to encourage projects to document their versioning scheme and binary compatibility rules. Can we add a short section suggesting how a project might do this? If it follows this scheme, we should recommend linking to this document in the README. If not, the README should ideally include a short section explaining the rules for that project.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,221 @@
...
+* Case classes
+
+You can find detailed explanations, runnable examples and tips to maintain binary compatibility in [Binary Compatibility Code Examples & Explanation](https://github.com/jatcwang/binary-compatibility-guide).
+
+Again, we recommend using MiMa to double check that you have not broken binary compatibility after making changes.
+
+## Versioning Scheme - Communicating compatibility breakages
+
+Library authors use versioning schemes to communicate compatibility guarantees between library releases to their users. Versioning schemes like [Semantic Versioning](http://semver.org/)(SemVer) allow
+users to easily reason about the impact of updating a library, without needing to read the detailed release note.
```
gmethvin: "...release note**s**."

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,221 @@
...
+## What are Evictions, Source Compatibility and Binary Compatibility?
+
+### Evictions
+When a class is needed during execution, the JVM classloader loads the first matching class file from the classpath (any other matching class files are ignored).
+Because of this, having multiple versions of the same library in the classpath is generally undesirable:
+
+* Need to fetch and bundle multiple library versions when only one is actually used
+* Unexpected runtime behavior if the order of class files changes
+
+Therefore, when resolving JARs to use for compilation and packaging, most build tools will pick only one version of each library and **evict** the rest.
```
gmethvin: It might be useful to explain that the default scheme for this in sbt (and gradle, and many other build tools) is to take the latest version, since most libraries are backwards but not forwards compatible.

jvican:
Okay, I think it’s about time to merge this PR, what’s your take on this? @jatcwang

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I did initially mention this but was asked to remove it. What do you think @jvican?

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: I'm trying to keep the word count down to not dilute the main takeaways of this tutorial. 

I think having another "maintaining an OSS Scala project" guide will be a good place to cover topics such as publishing to repositories and documenting versioning schemes. Speaking of which, I think it's a good idea to name this versioning scheme (e.g. "Scala Binary SemVer") instead of "that versioning scheme mentioned in that tutorial"

jatcwang:
@jvican, I've addressed some stuff @gmethvin picked up. (there's one question about mentioning SBT/maven's default eviction strategy - I can add it if you think it's good)

Also just want to make sure that the page is indexable by search engine bots, but we can just merge it and see whether it appears :)

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: I believe you can add "The most common solution in build tools is to take the latest version (sbt, gradle, etc)." and will still be good 👍

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: I agree, while @gmethvin's suggestion is good I think it may be too much for this guide.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jvican: @jatcwang I think Sébastien's suggestion here makes sense, just make sure you add a paragraph for "(such as OSGi)" so that it has less importance. But the wording is superior in precision :smile:.

_overviews/tutorials/binary-compatibility-for-library-authors.md
jatcwang: done

heathermiller:
@jvican please let me know when we can merge this :)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -0,0 +1,222 @@
...
+According to SemVer, patch versions should contain only bug fixes that fix incorrect behavior so major behavioral
+change in method/classes should result in a minor version bump.
+* When major version is `0`, a minor version bump **may contain both source and binary breakages**.
+
+Some examples:
+
+* `v1.0.0 -> v2.0.0` is <span style="color: #c10000">binary incompatible</span>. 
+  End users and library maintainers need to update all their dependency graph to remove all dependency on `v1.0.0`.
+* `v1.0.0 -> v1.1.0` is <span style="color: #2b2bd4">binary compatible</span>. Classpath can safely contain both `v1.0.0` and `v1.1.0`. End user may need to fix minor source breaking changes introduced
+* `v1.0.0 -> v1.0.1` is <span style="color: #2b2bd4">binary compatible</span>. This is a safe upgrade that does not introduce binary or source incompatibilities.
```
sjrd: is <ins>source and</ins> binary compatible

jatcwang: done :)

jvican:
Well, after 153 comments (without including this one), it's time to ship this valuable guide. Once again, I thank you for this awesome work, this is an unprecedent contribution from the Community and I'm really happy to see that there exists tenacious people out there that helps us improve Scala :smile:. 

:shipit:

SethTisue:
kudos to @jatcwang for tackling this and to everybody for the inspiring group effort here

jatcwang:
Thanks for all your help. Every suggestion and comment was appreciated :) and also thanks for all your awesome work in the scala ecosystem - many silent users like me appreciate it :)

PR #883, opened 2017-09-20
Update building-a-scala-project-with-intellij-and-sbt.md

RobvH:
It appears that the code example in Running The Project had been changed but the second portion, to change the code and demonstrate SBT watcher, had not.


SethTisue:
thank you!

PR #884, opened 2017-09-21
SIP-ZZ: Opaque type SIP

jvican:
This PR adds the latest changes to the opaque types SIP, an effort driven by @non and myself.
  
The new encoding differs significantly from our previous iteration, and it addresses feedback that we got from the SIP Committee and other Scala compiler hackers.


jvican:
I'm merging this so that Darja can link to it. But please, feedback is very welcome.

laughedelic:
It would be nice if there was an open PR with ongoing changes, so that one could subscribe to it. Is there any other way?

jvican:
Subscribe to the Scala Contributors thread for updates.

PR #894, opened 2017-10-09

Ironpencil94:
Change to class example "Complex" toString()-Method. 
Original would have printed  -i as "01i" instead of "0-1i"


SethTisue:
nice catch, thanks!

PR #564, opened 2016-08-27
Added Bosnian translations for tour and cheatsheet

sake92:
Added Bosnian translations for following:
- tutorials/tour
- cheatsheet

Since Croatian and Serbian are very similar languages to Bosnian, they will also benefit from this, I hope. :)

PR #895, opened 2017-10-11
Update Bosnian tutorials

sake92:
I've updated Bosnian translations of tutorials to reflect changes in original ones. 
No idea how `Gemfile.lock` ended in diff, probably when I merged the changes... :/

IDK if this is the right place, but I found some minor issues along the way:
- from English version of "Basics" page, when Bosnian is selected it redirects to `ba/tour/sics.html`. Probably a regex used somewhere or what? I guess other languages don't have it so 404 is returned.
- in section "Nested **methods**" method is nowhere mentioned, just **functions**. A bit confusing for beginners IMHO.
- in section "Traits", the `Iterator` definition is repeated. On purpose?
- there's a typo in "Upper Type Bounds" section, referring to non-existing `Cage` class, probably meant `PetContainer`.


_tour/abstract-types.md
```diff
@@ -1,4 +1,4 @@
----
+
```
heathermiller: This seems accidental. It would break the Abstract Types page, I think.

heathermiller:
Wow, this is a huge undertaking @sake92!! 

Thanks for the heads up about `ba/tour/sics.html`. Not sure what's happening; I'll look into it.

About the other issues; I agree that method being mentioned nowhere in the "Nested methods" section is confusing. We should probably go through and replace all of that with "method" instead of "function." About the Traits section; you're right that the `Iterator` definition is repeated. Thanks for catching this! And yep, you're right about the typo in the "Upper Type Bounds" section.  

Would you be up for fixing these minor issues on the English side as well?

Other than that, I found a small typo in the frontmatter. There was a bug; we had `----` when it should've been `---` but it seems it's been replaced with a newline. Not sure, but that would break the English Abstract Types page, no?

heathermiller:
For the record, I'll assume that your updates to the Bosnian translation are good! It wouldn't hurt though to have another Bosnian-speaker have a pass through the PR though, if you know of someone that might be up for it. Just to catch any typos or anything.

_tour/abstract-types.md
sake92: That's probably because I copied the headers from English versions.. :smile:
Three dashes aren't missing. It's just that git-diff rendered them after deleted section.  
I'm using Bash on Windows 10 and it renders ok.
If there are errors when merged I'll fix it.

sake92:
Thanks, @heathermiller. I'll make another PR for fixing those minor issues. [First time](https://github.com/scala/docs.scala-lang/pull/564) when I did these translations, Vojin Jovanović reviewed it (since Bosnian, Croatian and Serbian are almost the same 😆 ).  
I didn't translate all comments inside the code examples, most of them are obvious...

Also, somehow I skipped the "Classes" tutorial, so this could wait until I finish that also.

sake92:
@heathermiller, I've finished the "Classes" section. This is ready to be reviewed/merged now.

heathermiller:
Thank you! I'll merge this as-is.

Do you know anyone that may be up for passing over this once through? Just to catch typos etc?

PR #896, opened 2017-10-15
fix a wording issue

### LINKED ISSUE #876: A joint use of both features
"A joint use of both features facilitates the definition of new statements without using meta-programming facilities such as macros."

Which two features? 

They are actually mentioned in other languages but are missing here in the English version.

SethTisue:
fixes #876


heathermiller:
👍 Good point!

PR #897, opened 2017-10-18
Redirect /reference.html to /api/all.html

ashawley:
Doesn't seem like there's any reason for this page.

- http://docs.scala-lang.org/reference.html

The API links are old, and the rest of the links are in the site's navigation, so there's nothing worth preserving, IMO.  Seems it should redirect to:

https://docs.scala-lang.org/api/all.html

The referece page is still pretty prominent in google search results:

https://www.google.com/search?q=scala+api

Not sure why there is the "API API" in the title for the new page in the search results.

Google search: scala api


SethTisue:
LGTM!

PR #899, opened 2017-10-22
Fix typo: 'Iter' -> 'RichStringIter' on the CLASS COMPOSITION WITH MIXINS page

x0st:
Corrected a typo on the [CLASS COMPOSITION WITH MIXINS](http://docs.scala-lang.org/tour/mixin-class-composition.html) page.


SethTisue:
good catch, thanks!

PR #902, opened 2017-10-22
Fix Tour of Scala link

McBrainy:
The URL http://docs.scala-lang.org/tutorials/tour/tour-of-scala.html redirects to a version in another language (not English), which is undesirable since the link appears on an English-language page.


heathermiller:
Ah, darn, I wonder what happened! Regardless, *thank you very much* for catching this, debugging it, and making a fix! We really appreciate it! ❤️

PR #905, opened 2017-10-25
Split SIP33 into two SIPs

soronpo:
As requested by the October 2017's SIP committee, the SIP was split into two:
1. SIP33 now only specify infix types precedence (maintains its number)
2. SIPNN (no numbered) specifies prefix types.


SethTisue:
@soronpo I'm supposed to make sure that you received the feedback from the SIP committee — apparently you did :-)

could you also take a minute to review the parts of https://github.com/scala/docs.scala-lang/pull/907 that pertain to your SIP and verify that @darjutak represented them accurately?

soronpo:
> could you also take a minute to review the parts of #907 that pertain to your SIP and verify that @darjutak represented them accurately?

They were represented as the commented by the committee (aside for a minor error which I commented on now). _Accurately_ depends on my perspective which might be different than the committee's, but that should not part of the minutes 😄

jvican:
I'm reviewing this next Monday and I'll give @soronpo some feedback on the motivation as we discussed privately.

heathermiller:
(@jvican: @darjutak and I will hold you to it! 😈 )

_sips/sips/2017-10-25-adding-prefix-types.md
```diff
@@ -0,0 +1,163 @@
...
+  type N1 = -N //Not working
+  type N2 = ~N //Not working
+  type N3 = !N //Not working
+  type N4 = +N //Not working
+}
+```
+
+---
+
+## Proposal
```
jvican: Can you move this after the motivation section?

_sips/sips/2017-10-25-adding-prefix-types.md
```diff
@@ -0,0 +1,163 @@
...
+PrefixType ::= [`-' | `+' | `~' | `!'] SimpleType
+CompoundType ::= PrefixType
+              |  AnnotType {with AnnotType} [Refinement]
+              |  Refinement
+```
+
+---
+
+## Motivation
+Developers expect terms and types to be expressed the same for mathematical and logical operations.
```
jvican: Reading just this line I have several questions:
  
1. Can you elaborate on why developers expect that?
1. Perhaps also make a point to make the language more consistent?
1. Can you explain in which use cases or domain are prefix types more likely to be useful?
1. Can you come up with more interesting and *shorter* examples that show the value of prefix types?
1. Can you find other programming languages that support prefix types and tell us about their experience using them?
  
In my opinion, this motivation is not enough to gauge attention from the Committee... motivations are self-contained explanations that research on why the language is lacking without a certain lanaguage feature. They are usually rich in details and examples, and they try to be as objective as possible. For example, saying "developers expect terms and types to be expressed the same for mathematical and logical operations" is not backed up by any empirical or objective data, and it sounds overly subjective.
  
I feel that you can make a much better case for prefix types if you turn your statement around: try to go to the data and analysis first, and then you draw some conclusions from there. Prefix types look like a language feature that has not perceived lots of attention in other languages -- in this case, I think it's important to explain in which scenarios (with concrete examples) would they be useful.

_sips/sips/2017-10-25-adding-prefix-types.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Motivating examples
+
+The [singleton-ops library](https://github.com/fthomas/singleton-ops) with [Typelevel Scala](https://github.com/typelevel/scala) (which implemented [SIP-23](http://docs.scala-lang.org/sips/pending/42.type.html)) enable developers to express literal type operations more intuitively. For example:
+
+```scala
```
jvican: Can you add text that complements this snippet of code? For examples, saying "let's define foo and bar for..." or something similar. You can draw inspiration from SIP-35 Opaque types.

_sips/sips/2017-10-25-adding-prefix-types.md
```diff
@@ -0,0 +1,163 @@
...
+* [Typelevel Scala Issue #157](https://github.com/typelevel/scala/issues/157)
+
+Both SIP23 implementation and Dotty's implementation of literal types currently fail compilation when infix types interact with a negative literal type.
+```scala
+type ~~[A, B]
+type good = 2 ~~ 2
+type bad = 2 ~~ -2 //Error:(9, 20) ';' expected but integer literal found.
+type work_around = 2 ~~ (-2) //works for Typelevel scala, but fails in Dotty
+```
+It is not yet clear if this is an implementation issue, or if the spec should be changed to allow this as well.
```
jvican: I think it's important that this is sorted out before presenting the proposal to the Committee (you want to give the impression that you've researched and understood future interactions with prefix types). In particular, why is this happening?
  
You may also ask the Dotty team about more clarifications in the issue tracker if you cannot find the root of the problem. /cc @smarter

soronpo: After investigating this further for TLS, I figured out that the SIP23 proposal (and implementation) should be changed to support infix types and negative literal types. Interestingly, implementing prefix types resolves this issue, because a `-` prefix type has a special case to handle negative literal types. Here is a branch I made that fixes SIP23:
https://github.com/soronpo/scala/blob/3aeb19a7ff9e9c2a1dda61899d4c6482cdd8343e/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala#L1052
You can see that a Spec change is required to handle `NegativeNumericLiteral` inside a compound type.
The prefix expressions do the same thing, so that is why you don't have such a bug with infix expression and negative literal values.

_sips/sips/2017-10-25-adding-prefix-types.md
soronpo: :+1:

_sips/sips/2017-10-25-adding-prefix-types.md
soronpo: :+1:

_sips/sips/2017-10-25-adding-prefix-types.md
soronpo: I understand what you mean. The problem is justifying this can be just as easily a chicken and egg issue. Nobody uses prefix types because they are not available and the alternative is "good enough". Will a survey passed among the community be acceptable by the committee?

I cannot tell the committee it is not a _niche_ because it is. It will only serve the typelevel community, but it allow us to think _less_ about the language, in general. Currently, the we have to think differently for typelevel operations, because the language made it different. 

Other than the potential for confusion with variance annotation, I don't understand the reluctance include this feature in the language:
* Prefix types are _not_ for doing the same thing but different. They add something that we can't do at all. 
* The implementation is very simple and should not affect compiler in any way.

_sips/sips/2017-10-25-adding-prefix-types.md
milessabin: I think that SIP-23 should handle negated numeric literal types more smoothly than it does at present (see the example at the head of https://github.com/typelevel/scala/issues/157). However I think that the prefix type proposal goes beyond and is somewhat orthogonal to this, so I think it deserves a separate SIP.

jvican: @soronpo Can you update this last part and we merge? Thanks for your hard work Oron.

_sips/sips/2017-10-25-adding-prefix-types.md
jvican: > I understand what you mean. The problem is justifying this can be just as easily a chicken and egg issue. Nobody uses prefix types because they are not available and the alternative is "good enough".
  
If this hasn't been tested before enough, how are we sure that it merits being accepted in the Scala Language Specification? :smile:
  
> Other than the potential for confusion with variance annotation, I don't understand the reluctance to include this feature in the language:
  
Let's see what the Committee thinks. I'll discuss it with Martin too. Thank you for your feedback and updating this document!

PR #908, opened 2017-10-26
correct a couple of SIP committee affiliations

SethTisue:


SethTisue:
@dragos unless you actually prefer to be listed as "independent"

SethTisue:
the Drone failure is unrelated to my change

dragos:
Looks good to me!

PR #909, opened 2017-10-27
SIP-34 accepted

darjutak:
Please merge the change in vote-status for SIP-34.
Thanks
Darja

### DIFF
--- a/_sips/sips/2017-07-12-right-associative-by-name-operators.md
+++ b/_sips/sips/2017-07-12-right-associative-by-name-operators.md
@@ -3,7 +3,7 @@ layout: sip
 discourse: true
 title: SIP-NN - Right-Associative By-Name Operators
 
-vote-status: pending
+vote-status: accepted
 permalink: /sips/:title.html
 ---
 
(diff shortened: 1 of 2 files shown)


_sips/sips/2017-07-12-right-associative-by-name-operators.md
```diff
@@ -3,7 +3,7 @@ layout: sip
 discourse: true
 title: SIP-NN - Right-Associative By-Name Operators
```
jvican: Are you sure this is SIP-34? I think it's 33.

darjutak: the 33 is match infix?

http://docs.scala-lang.org/sips/make-types-behave-like-expressions.html

darjutak: and thank you for the git compliment! I will frame it ((((o:::
I had THE best teacher, with incredible patience and presence = @heathermiller <3

jvican: Yup, but since the numbers cannot be sequential because there were other SIPs numbered in the meanwhile. This would be SIP 36, if I'm not mistaken!

darjutak: screen shot 2017-10-27 at 13 53 28
I don't know if you are wrong, but contradictory for sure (o:

jvican: Oh my! I read the title of the SIP wrong. Yeah, this is correct!

PR #912, opened 2017-10-31

jfwm2:
Not sure if that change is scala-compliant, but if im<0, we should use the absolute value of im to avoid printing a leading minus sign.


SethTisue:
shorter if instead of adding an `abs` call, you change the `if` to read `(if (im < 0) "" else "+")`. nbd either way I guess?

jfwm2:
Indeed shorter!

SethTisue:
good catch, thanks!

PR #915, opened 2017-11-01
Fix cover image of 'Scala Puzzlers' book

Philippus:
it was showing the cover of 'Scala in Depth'


SethTisue:
good catch — thanks!

PR #919, opened 2017-11-03
add return type other than Unit onto the prepend() function in trait …

justinpermar:
…Bird

I don't think this page intends to have a Unit return type for prepend. Without it, prepend() doesn't do anything useful: ```scala> birdList.prepend(new EuropeanSwallow)``` returns Unit. So add the return type so prepend() does what the reader expects it to. With the suggested change, I get ```scala> birdList.prepend(new EuropeanSwallow)
res16: Node[Bird] = ListNode(EuropeanSwallow(),ListNode(AfricanSwallow(),Nil()))```


SethTisue:
good catch, thanks!

PR #922, opened 2017-11-07
Removed Bosnian translation redirects

srdjanm88:
I clicked a 'normal' looking link (http://docs.scala-lang.org/tutorials/tour/singleton-objects.html) to the docs in an English [blog post](http://www.jesperdj.com/2016/01/08/scala-access-modifiers-and-qualifiers-in-detail/) about access qualifiers, and ended up with the Bosnian translation. I forwarded the link to some scala gitter members and they all reported the same thing happening. I'm pretty sure this is unintended, left over from copy-pasting the articles under `_tour/` to translate as none of the other languages have it.

I just removed the lines from each of the Bosnian articles to bring them in line with the other languages, though it may be worth considering redirecting to the different translations based on the user's browser language?

### DIFF
--- a/_ba/tour/abstract-types.md
+++ b/_ba/tour/abstract-types.md
@@ -12,7 +12,6 @@ next-page: compound-types
 previous-page: inner-classes
 prerequisite-knowledge: variance, upper-type-bound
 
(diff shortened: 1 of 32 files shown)


SethTisue:
nice catch!! thanks

PR #926, opened 2017-11-12
java.lang.ArrayIndexOutOfBoundsException: 0

Kobenko:


SethTisue:
the intention here was that you pass a single argument on the command line when you run the app. if you don't pass one, you get the `ArrayIndexOutOfBoundsException`

I tend to agree, though, that it just distracts from the main idea of this part of the tour.

how about just `extends StringIterator("Scala")`? this avoids the unnecessary extra `val`, plus it's "Scala" not "SCALA"

Kobenko:
Yes, passing a string value to an argument would be a good option

Kobenko:
@SethTisue how often do you measure the branches? I have several more suggestions for improving the tour.

SethTisue:
thank you for this!

merging PRs is done on no particular schedule by Heather, myself, and others. I usually don't merge things until they've been sitting for a few days, to allow time for community review and feedback.

there is certainly plenty of room for improvement left on the tour — looking forward to further suggestions and pull requests!

PR #928, opened 2017-11-16
Updated following implementation experience in Typelevel Scala

milessabin:


heathermiller:
cc/ @jvican

jvican:
I've been summoned and I approve!

smarter:
I see this got merged last week but http://docs.scala-lang.org/sips/pending/42.type.html is still out of date, how come?

jvican:
That's the old link. New link is here: http://docs.scala-lang.org/sips/42.type.html.

smarter:
Old one should be delete then, because if I google SIP 23 that's what I find.

PR #929, opened 2017-11-17
Fixed code to match comments

salborrelli:
Comments on lines 53 and 54 make me think that the program prints something as output, which obviously wasn't the case in the previous version. I've added the calls to "println" that are necessary to match the corresponding comments.


heathermiller:
Good point! Thanks for noticing and going through the trouble of submitting a fix! We really appreciate it! ❤️

salborrelli:
You're welcome 😃

PR #930, opened 2017-11-20
Update LinearSeq Link From IndexedSeq

shankarshastri:
LinearSeq used to refer to the wrong URL.


SethTisue:
good catch, thanks!

PR #932, opened 2017-11-21
Fix code comment to print the correct value

MarounMaroun:
Fix output in comment, should be `(2, 3)` and not `(x, y)`:

```scala
val point1 = new Point(2, 3)
point1.x  // 2
println(point1)  // prints (2, 3) <- here's my change, was (x, y)
```


SethTisue:
good catch. thanks!

PR #934, opened 2017-11-21
Fix Typo: `decls` to `decl` + Upgrad from deprecated API usage

reza-samei:
`ru.typeOf[X].decls(???)` fails in compile time, because `decls` method takes nothing as input and return `MethodScope`. The right method to call is: `decl(name: Name): Symbol`.

And also I've changed `ru.nme.CONSTRUCTOR` to `ru.termNames.CONSTRUCTOR` based on the deprecation-warning on my Scala compiler: `2.11.8`


marconilanna:
Superseded by #935 and #946

SethTisue:
@reza-samei thanks for your attention to this, and sorry that we were slow about reviewing/merging

PR #935, opened 2017-11-24
Fix code sample in reflection overview

marconilanna:
Fix the following issues:

0. Deprecated methods `declarations`/`declaration` replaced with `decls`/`decl`;
0. `decls` used instead of `decl`
0. Links to current Scaladoc;
0. Typo in Japanese docs: `meber`

### DIFF
--- a/_overviews/reflection/overview.md
+++ b/_overviews/reflection/overview.md
@@ -132,7 +132,7 @@ The second step involves obtaining a `ClassMirror` for class `Person` using
 the `reflectClass` method. The `ClassMirror` provides access to the
 constructor of class `Person`.
 
-    scala> val ctor = ru.typeOf[Person].decls(ru.nme.CONSTRUCTOR).asMethod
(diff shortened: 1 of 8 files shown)


_overviews/reflection/overview.md
```diff
@@ -171,7 +171,7 @@ which makes all classes and types available that are loaded by the classloader
 that also loaded the class of `p` (`Purchase`), which we need in order to
 access member `shipped`.
 
-    scala> val shippingTermSymb = ru.typeOf[Purchase].decls(ru.TermName("shipped")).asTerm
+    scala> val shippingTermSymb = ru.typeOf[Purchase].declaration(ru.TermName("shipped")).asTerm
```
hrhino: I think that `declaration` is [deprecated in favor of `decl`](https://github.com/scala/scala/blob/v2.12.4/src/reflect/scala/reflect/api/Types.scala#L97-L99) these days...

marconilanna:
@hrhino thanks for point that out. fixed.

SethTisue:
thank you!

PR #945, opened 2017-12-05
Fix typo in by-name-parameters.md

qbahers:
Replace `have the the advantage that` with `have the advantage that`.


SethTisue:
good catch, thank you!

PR #947, opened 2017-12-05
Update CI to check for dead links

OlivierBlanvillain:
This PR adds dead link detection to the CI (via `htmlproofer`) and fixes the ~200 dead links currently on docs.scala-lang.org.

I couldn't find a better alternative to 925e42038002f5a338f955a6f383c2e07e870213 which adds empty pages when there is no translations. As a results, the CI is expected to fail because these new "empty pages" point to non existing md files on master with their "Problem with this page? Please help us fix it!" links.

I've tried to split the mess in sensible commits, it probably easier to review commit by commit rather than look at the entire diff.


SethTisue:
I'm able to view the Drone log now, I wasn't before for some reason.

Before we can merge this, we'll need _some_ solution that permits CI to pass. @OlivierBlanvillain do you have a suggestion of how to handle that...?

OlivierBlanvillain:
There is a chicken an egg problem with newly added pages: new files are not on <https://github.com/scala/docs.scala-lang/blob/master/*> until a PR is merged, but the generated pages contain links to github blobs for the edit button... htmlproofer has a `--url-ignore` flag, I will try that! (I don't have all the ruby stuff setup on this machine)

SethTisue:
oh, so we should just merge, and then future CI runs will pass? if so, that's good enough for me

OlivierBlanvillain:
Yes, bug you will have the same issue every time a new page is added, I think it's worth addressing before merging :)

SethTisue:
is this ready for final review/merge...?

OlivierBlanvillain:
Someone broke a tut between since the PR was opened, and tut are not checked by the CI on master. I won't have time to come back to it until end of next week

SethTisue:
I'll take care of the `tut` thing separately.

SethTisue:
as of #984, PR validation runs are now green

PR #949, opened 2017-12-09
Update binary compatibility documentation.

hrhino:
We (as evidenced by the mima filters) don't guarantee binary compatibility for `scala.reflect.io` or `scala.runtime`, as well; both are marked as "not for external consumption".

I'm not sure which packages are "we'll change them if we need to" as opposed to "no guarantees whatsoever!", so let me know if this seems over-eager.

Also, `scala-library-all` [has been removed](https://github.com/scala/scala/pull/5790) in 2.12.3-M2, so add a note about that.


SethTisue:
thanks!

PR #950, opened 2017-12-14
Avoid `_` inner parameter names in higher-kinded types

Atry:
See https://gitter.im/scala/contributors?at=5a2f68980163b02810269751 and https://github.com/lampepfl/dotty/issues/2041 for the motivation


SethTisue:
I don't think there's sufficient consensus on this to enshrine it in the style guide. I've never seen this style used or recommended elsewhere. (Is it in wide use and I'm just not aware...?)

SethTisue:
discussion thread at https://contributors.scala-lang.org/t/do-the-syntaxes-for-existential-types-and-higher-kinded-types-confuse-you/1403/1

SethTisue:
I don't think there's sufficient consensus on this to enshrine it in the style guide.

PR #955, opened 2017-12-17
Fixed typo in higher-order-functions.md

yannayl:


SethTisue:
good catch — thank you!

PR #959, opened 2017-12-20
Update testing-scala-with-sbt-on-the-command-line.md

goodgoodwish:
This scala/scalatest-example.g8 project doesn't seems to exists anymore.
We may use: sbt new scala/scala-seed.g8


jvican:
Is the patch updated with the new link?

goodgoodwish:
No.  
scala/scala-seed.g8 is not exactly same as scala/scalatest-example.g8.
We'd better have scala/scalatest-example.g8 restored.

SethTisue:
perhaps scalatest-example.g8 never actually existed, but was just something intended to be created...?

goodgoodwish:
Probably you're correct.

I don't know how to create scalatest-example.g8 yet.
I'll wait for you, or I'll do it when I become stronger in future.

PR #961, opened 2017-12-23
Add "previous" And "next" To Tour Pages

wmonk:
Hi! I just started reading the tour page, and one thing that I really wanted was a way to jump to the next page, without scrolling up and clicking something in the contents block to the right hand side. 

As far as I can tell, each page in the "tour" section implements front matter that specifies the previous and next pages, so this will just work for each of them.

screen shot 2017-12-23 at 09 28 50

### DIFF
--- a/_layouts/tour.html
+++ b/_layouts/tour.html
@@ -9,7 +9,18 @@ includeCollectionTOC: true
 		<div class="content-primary documentation">
 			<div class="inner-box">
 				{{content}}
-			</div>
+
+                <div class="two-columns">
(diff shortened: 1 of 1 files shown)


SethTisue:
fantastic! thanks so much for tackling this

jvican:
Wow, this is really nice,thank you @wmonk :smile:.

wmonk:
🎉  no problem!

SethTisue:
@wmonk see #977, does this need adjusting...?

PR #962, opened 2017-12-24
Update scalafiddle/scastie links

hrhino:
[scala-js-fiddle](http://www.scala-js-fiddle.com/) is now ScalaFiddle; [Scala Kata](http://www.scalakata.com/) redirects to Scastie.


SethTisue:
thanks!

PR #963, opened 2017-12-24

thiaguten:
The predefined "error" method no longer exists in the current version of Scala.


SethTisue:
thanks for the fix!

PR #966, opened 2017-12-30
Update files.md

lanjiann:
Follow the Scala API, ```Option[+A]``` should be an abstract class, NOT a trait.

Conceptually, this is important. When you define a type, it should be a class or an abstract class. When you annotate a type, it can be a class, an abstract class, or a trait.

### DIFF
--- a/_style/files.md
+++ b/_style/files.md
@@ -39,7 +39,7 @@ file. One common example is that of a sealed trait and several
 sub-classes (often emulating the ADT language feature available in
 functional languages):
 
-    sealed trait Option[+A]
+    sealed abstract class Option[+A]
 
     case class Some[A](a: A) extends Option[A]
 


SethTisue:
It seems to me that `sealed trait` is pretty common for this, although I do certainly also see `sealed abstract class` around. But stylistically, I tend to think of `trait` as normal code and `abstract class` as a JVM-level optimization (possibly a premature optimization). Which version is more "normal" isn't a settled question, IMO.

Even if we did change the code as you suggest, the text just before it would need a parallel change; it currently says "sealed trait and several sub-classes".

Perhaps it would make the most sense to change "sealed trait" to "sealed trait or abstract class" in that sentence and leave the code alone? `trait` vs `sealed class` is not the main point being addressed in this passage.

lanjiann:
Yes, I think "sealed trait or abstract class" is good.

I still want to discuss the concept things, but I think I shouldn't post it here. I may create a question in Stack Overflow.

SethTisue:
closing for inactivity, but we can definitely reopen this if you want to return to it!

martijnhoekstra:
@SethTisue did you intend to re-open?

PR #967, opened 2017-12-30
Update declarations.md

lanjiann:
In the **Modifiers** subsection there is a method definition ```foo()```, which does NOT follow a rule of this style guide.


SethTisue:
thanks!

PR #969, opened 2018-01-03
Add support for direct ScalaFiddle integration into document code examples

ochrons:
This adds a ScalaFidde integration Jekyll plugin to the docs site. With this plugin it's trivial to convert existing code examples into fully interactive fiddles. A few examples are provided in the `basics.md` file.

## Description

Good documentation is critical for the success of Scala. Code examples are an important part of documentation as they illustrate concepts in a clear and understandable way. The problem is that examples are static, so you can show only a limited set of features in a single example.

Making code examples interactive allows readers to not only read the code, but make changes and experiment. This enhances learning as readers become experimenters and can try out different things quickly.

ScalaFiddle was built to be embedded from the very beginning. But before these new [integration plugins](https://github.com/scalafiddle/scalafiddle-core/tree/master/integrations) it was quite hard to actually embed a fiddle. The new Jekyll plugin makes it easy by just surrounding a code block in the documentation with `{% scalafiddle %} ... {% endscalafiddle %}`. For further information, refer to the integration documentation in the link above.

## Performance

The impact on loading time of documentation pages due to ScalaFiddle integration is neglible. The page loads a single JavaScript file (2.5kb gzipped) from CloudFlare CDN, which then adds `Run` buttons to each fiddle. Nothing else is loaded at this point.

When the reader clicks a `Run` button to interact with the fiddle, the actual ScalaFiddle embedded editor is loaded (via CDN, less than 300kB in total, cached forever by the browser). The editor automatically compiles and runs the code. Since compilation results are cached, they will be loaded immediatelly from the CDN cache. The first compilation utilizes Scala.js full optimization to make the downloaded JS code very small (for these simple examples typically less than 20kB). When changes are made, the code is compiled using "fast optimization" and typically the results are delivered in under 400ms (typically less than 50kB gzipped). For more information on why ScalaFiddle is faster(!) than running Scala on your own computer, read [my recent blog post](https://medium.com/@otto.chrons/what-makes-scalafiddle-so-fast-9a3edf33ed4d).

Overall, this will give the reader a good impression on Scala compilation speed. 

## Next steps

This PR adds only a few fiddles, but there are a lot more code examples that could be turned into interactive fiddles using the plugin. Since there is no overhead in making examples interactive, every code example that can be a fiddle, should be!

Since ScalaFiddle runs in the browser, you can do more than just `println`. With suitable templates providing helper functions it's easy to provide better visualizations for data structures or even interactive HTML to illustrate the power of Scala.


heathermiller:
Sounds like a good idea! Mind if I have a look at it next week? I'm on the road this week + a zillion deadlines.

heathermiller:
THIS LOOKS GREAT @ochrons! Even works like a charm on mobile :) 

Thank you!

retronym:
Nice!

A few minor styling suggestions:

The `Run` button can obscure the code:

image

The `Edit on Scalafiddle` button look a bit pixelated to me:

image

heathermiller:
@ochrons this is a nitpick, but perhaps there's a way to shrink the ScalaFiddle button and lift it up into the code block's padding? (Just thinking out loud...)

ochrons:
Actually I did notice the same problem and experimented with some other margin/padding settings, including making it go partially above the code block. I'll continue with the experiments and make the modifications to keep it away from the code :)

The "Edit on ScalaFiddle" image would need another high-res version for "retina" displays that scale everything 2x. I'll look into this, too!

Thanks for the comments!

SethTisue:
very cool, kudos Otto

PR #970, opened 2018-01-04
Fix return type of forecastInFahrenheit

rjdestigter:
forecastInFahrenheit returns a `Seq[Double]`, not a `Double`


SethTisue:
good catch! thanks

PR #972, opened 2018-01-05
Change the type of return value.

xllh:


SethTisue:
thanks, but #970 beat you to it 😃

Issue #975, opened 2018-01-07
out-of-order: def foldLeft[B](z: B)(op: (A,B) => B): B


chaokunyang:
In https://docs.scala-lang.org/style/declarations.html
There is a out-of-order in MULTIPLE PARAMETER LISTS For type inference: 
```def foldLeft[B](z: B)(op: (A,B) => B): B``` should be changed to ```def foldLeft[B](z: B)(op: (B,A) => B): B```

SethTisue:
agree. would you like to submit a pull request with the correction...?

chaokunyang:
Of course, it would be my pleasure. I'll make a pull request later.

chaokunyang:
the [pull request](https://github.com/scala/docs.scala-lang/pull/976) has be submitted

PR #976, opened 2018-01-09
fix out-of-order: def foldLeft[B](z: B)(op: (A,B) => B): B

chaokunyang:
fix out-of-order:  ```def foldLeft[B](z: B)(op: (A,B) => B): B```  to ```def foldLeft[B](z: B)(op: (B, A) => B): B```
The details is in https://github.com/scala/docs.scala-lang/issues/975


SethTisue:
nice, thanks!

PR #978, opened 2018-01-10
rewrote singleton objects tour

som-snytt:
Rebased Travis's work.

### DIFF
--- a/_tour/singleton-objects.md
+++ b/_tour/singleton-objects.md
@@ -13,11 +13,13 @@ previous-page: pattern-matching
 redirect_from: "/tutorials/tour/singleton-objects.html"
 prerequisite-knowledge: classes, methods, private-methods, packages, option
 ---
+An object is a class that has exactly one instance. It is created lazily when it is referenced, like a lazy val.
 
-A singleton object is an instance of a new class. There is exactly one instance of each singleton object. They do not have constructors so they cannot be instantiated.
+As a top-level value, an object is a singleton.
 
+As a member of an enclosing class or as a local value, it behaves exactly like a lazy val.
 # Defining a singleton object
-The simplest form of an object is the keyword `object` and an identifier:
+An object is a value. The definition of an object looks like a class, but uses the keyword `object`:
 ```tut
 object Box
 ```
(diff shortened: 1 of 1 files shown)


SethTisue:
the addition of the `.ruby-version` file is accidental?

_tour/singleton-objects.md
```diff
@@ -10,67 +10,101 @@ num: 13
...
 ```
+The `object Email` contains a factory `fromString` which creates an `Email` instance from a String. We return it as an `Option[Email]` in case of parsing errors.
 
-This illustrates another feature: in the context of `private`, a class and its companion are friends. `object X` can access private members of `class X`, and vice versa. To make a member *really* private to one or the other, use `private[this]`.
+Note: If a class or object has a companion, both must be defined in the same file. To define companions in the REPL, either define them on the same line or enter `:paste` mode.
+
+## Notes for Java programmers ##
 
-For Java convenience, methods, including `var`s and `val`s, defined directly in a singleton object also have a static method defined in the companion class, called a *static forwarder*. Other members are accessible via the `X$.MODULE$` static field for `object X`.
+`static` members in Java are modeled as ordinary members of a companion object in Scala.
```
jvican: @som-snytt I think this is a very important concept, and there are several deletions that explain how people can encode statics in Scala 🤔 Do you think it's reasonable that we emphasize it more in the new section?

SethTisue:
ping @som-snytt, shall we try to get this across the finish line?

som-snytt:
@SethTisue We shall! or maybe go for a field goal.

SethTisue:
no way. touchdown or go home.

_tour/singleton-objects.md
som-snytt: Sorry for the late reply. Re-reading the diff, I see that static forwarding is defined on the next line 110. I'm not sure if it's less visible in the diff, or if it really needs more words.

som-snytt:
I did that thing where the snap is to some guy who pitches it forward to the QB.

I have no idea what I'm doing, but I deleted the spurious file from the old contribution and amended my commit text.

I think static forwarders are adequately addressed, but no doubt someone might like to add more words later.

SethTisue:
> I think static forwarders are adequately addressed, but no doubt someone might like to add more words later.

@jvican followup PR if you like

SethTisue:
Homer Zuckerman: There he is!
Arable: That's Som Snytt.
Avery Arable: He's terrific.
Lurvy: He's radiant.
Mrs. Zuckerman: Well, he's clean anyway. That buttermilk certainly helped.

som-snytt:
TIL snytt is norwegian for suckling pig.

Issue #977, opened 2018-01-10
Navigation links are broken in "Tour of Scala" at http://docs.scala-lang.org


elpinal:
The bottom links, "next →" and "← previous", of each page of "Tour of Scala" are broken since those URLs are not given the suffix ".html".
For example, http://docs.scala-lang.org/tour/basics.html is accessible, but http://docs.scala-lang.org/tour/basics, which is "next →" in http://docs.scala-lang.org/tour/tour-of-scala.html indicates, responds with "404 Not Found error".

Note that running `bundle exec jekyll serve -I` in local works correctly.

Madusudanan:
This issue seems to be sitewide. The below URLs also do not work without the .html extension for example.

http://docs.scala-lang.org/sips/all

http://docs.scala-lang.org/overviews/core/architecture-of-scala-collections

I guess it has to do with the permalink, I think 

```
# Permalinks
permalink:        pretty
```
should fix it - https://jekyllrb.com/docs/permalinks/#permalink-style-examples. And it should work both ways i.e with and without the .html extension.

But this is tricky to test as it works in local by default.

elpinal:
It is important to use URLs in a consistent format. I'd suggest that use just `.html` URLs and fix all links which lack `.html` extensions and point wrong paths.

A naive way may solve it: adding ` | append: '.html'` to `{{page.previous-page}}` and `{{page.next-page}}` in `_layouts/tour.html`. If using `.html` eternally, I think it's enough.

Madusudanan:
I agree on consistent URL format but I disagree on the `.html` for all pages. It doesn't look clean and can lead to errors (IMHO). We can have jekyll to handle both, i.e with and without html automatically, but without the `.html` as the default and only use clean URLs . Maybe the community can take a call on the naming convention.

elpinal:
>  I disagree on the .html for all pages
1. I didn't mean `.html` for __all__  pages. And I didn't suggest the change of permalinks.
2. I'd prefer URLs without `.html` to with `.html` (FYI).
3. __All I request is fix the bottom links, "previous" and "next", in "Tour of Scala" of https://www.scala-lang.org.__

> We can have jekyll to handle both

I disagree with allowing both formats, i.e. with and without `.html`. A mixture of these formats makes visitors confusing.

I apologize for the obscure explanation.

Madusudanan:
> I disagree with allowing both formats, i.e. with and without .html. A mixture of these formats makes visitors confusing.

[...]

PR #982, opened 2018-01-10
fix non-compiling tut code in tour

SethTisue:


SethTisue:
now that #947 has been merged, Drone is failing, but it's because of #977, not because of the change in this PR.

PR #714, opened 2017-02-25
New packages and imports tour

travissarles:

PR #984, opened 2018-01-10
add missing translations of tour sections

SethTisue:
they still need to actually be translated, but in the meantime,
we need placeholders or html-proofer fails

followup to #714


SethTisue:
screen shot 2018-01-10 at 11 47 06 am

PR #986, opened 2018-01-11
Rewrite Currying tour as Multiple Parameter Lists

sujithjay:
replaces #732 . This is a work in progress; raising a PR to elicit comments.


sujithjay:
cc: @SethTisue @jvican 
Please take a look. 

PS: The build failure is due to the stale link '/tour/currying.html' in files, other than the one in English. I will fix them asap.

_tour/multiple-parameter-lists.md
```diff
@@ -0,0 +1,44 @@
...
+partof: scala-tour
+
+num: 10
+next-page: case-classes
+previous-page: nested-functions
+
+redirect_from: "/tutorials/tour/multiple-parameter-lists.html"
+---
+
+Methods may define multiple parameter lists. When a method is called with a fewer number of parameter lists, then this will yield a function taking the missing parameter lists as its arguments.
```
jvican: "This is formally known as currying." and then maybe a link to the currying wikipedia page? https://en.wikipedia.org/wiki/Currying

_tour/multiple-parameter-lists.md
```diff
@@ -0,0 +1,44 @@
...
+
+Here is an example, defined in [Traversable](/overviews/collections/trait-traversable.html) trait from Scala collections:
+
+```
+def foldLeft[B](z: B)(op: (B, A) => B): B
+```
+
+`foldLeft` applies a binary operator `op` to an initial value `z` and all elements of this traversable, going left to right. Here is an example of its usage:
```
jvican: I would mention here one of the primary reasons why currying is useful: to fix the parameter `z` and reuse it in the call site several times. For example, when we're using the same parameter `z` all the time, it's better to reuse the same function.

I'm trying to find a good example that illustrates this scenario, but cannot come up with a good one. Do you happen to have a good one in mind? :smile:

_tour/multiple-parameter-lists.md
```diff
@@ -0,0 +1,44 @@
...
+val numbers = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+val res = numbers.foldLeft(0)((m: Int, n: Int) => m + n)
+print(res) // 55
+```
+
+Starting with an initial value of 0, `foldLeft` applies the function `(m: Int, n: Int) => m + n` to each element in the List and the previous accumulated value.
+
+Multiple parameter lists have a more verbose invocation syntax; and hence should be used sparingly. Suggested use cases include:
+
+1. ##### Implicit parameters
```
jvican: Is this rendered correctly in markdown?

_tour/multiple-parameter-lists.md
```diff
@@ -0,0 +1,44 @@
...
+    To specify certain parameters in a parameter list as `implicit`, multiple parameter lists should be used. An example of this is:
+
+    ```
+    def execute(arg: Int)(implicit ec: ExecutionContext) = ???
+    ```
+
+2. ##### Single functional parameter
+    In case of a single functional parameter, like `op` in the case of `foldLeft` above, multiple parameter lists allow a concise syntax to pass an anonymous function to the method.
```
jvican: I think it would be good to elaborate more here and maybe add a good example, as I described in the previous comment.

jvican:
Thanks for picking this up @sujithjay, it's appreciated.

_tour/multiple-parameter-lists.md
tfinneid: I agree, but I would also suggest changing the title to "Multiple Parameter Lists (Currying)". That would make it easy for people to recognise sections according to specific names they have heard of, with regards to functional programming. One often hears about the formal names for features of the language.

In general, I think that titles should contain a pedagogical title part, such as "multiple parameter lists" and a pat containint some sort of reference to the more formal name.

sujithjay: @tfinneid , I am not sure how much value there is in adding the formal name to the title as well. Although I do not see a downside to it either (other than rather long section names). I would like @jvican 's opinion on this before I make my edits.

jvican: I see value in having the technical term in the title just for those that are already familiar with the term and want to see what the Scala tour says about it, or how Scala allows to encode currying in the language. :smile:

_tour/multiple-parameter-lists.md
Adowrath: I think that being able to reuse the parameter is just one of the advantages here.
Another is the special syntax you can get, `list.foldLeft(z) { (acc, value) => /**/ }`.

A bigger advantage is, though, that `scalac` doesn't use type information in one parameter for another parameter's inference.
With a `def foldLeft[B](z: B, op: (B, A) => B)`, you can't write `List(1, 2).foldLeft(0, _ + _)`. It complains about the first `_` having a missing parameter type.

jvican: > A bigger advantage is, though, that scalac doesn't use type information in one parameter for another parameter's inference.

sujithjay:
cc: @jvican @Adowrath @tfinneid @SethTisue 
Please take a look

jvican:
I defer merging to @SethTisue

tfinneid:
@sujithjay, as @jvican also states, I think the title should also contain the formal name of the language feature.

SethTisue:
thanks, @sujithjay, for taking this on!

SethTisue:
followup PR to further clarify the relationship between multiple parameter lists and currying: https://github.com/scala/docs.scala-lang/pull/2490

PR #993, opened 2018-01-16
language tour: say what the intent and audience are

SethTisue:


SethTisue:
note that historically, we used to get tons of Disqus comments from people saying "this language tutorial sucks, how dare you call this a tutorial, etc etc etc". hopefully the consistent use of "tour" in the URLs and pages has brought down the outrage levels since then. regardless, I think it's still useful to pre-emptively set expectations.

PR #998, opened 2018-01-17
Rewrote implicits section of the tour #746

mghildiy:


mghildiy:
I added code snippets for explaining implicit class. Build fails for that code:
```tut
package prac.implicits
 
object StringNuggets {
  implicit class StringImprovements(val s: String) {
    def increment = s.map(c => (c + 1).toChar)
  }
}

package prac.implicits

import prac.implicits.StringNuggets._

object ImplicitDemo {

  def main(args: Array[String]): Unit = {
    println("Einstein".increment)
  }
}
```
How to work around it?

SethTisue:
it appears to me that tut doesn't support package declarations. are they necessary here? if the example can do without them, you can remove them. if they are necessary, you can disable tut by removing the `tut` after the triple backquote at the top of the code snippet in question

mghildiy:
Do I need to remove older implicit files(implicit-parameters.md,implicit-conversions.md) from language specific directories(_ba\tour,_es\tour,_ko\tour etc) too?It seems jekyll build tries to link these files to the older implcit files I have deleted from  _tour, and hence fails.

mghildiy:
Build passed. Ready for review.

mghildiy:
Can anybody review it?

jvican:
Sorry for the delay @mghildiy, I'll have a look at these over this w-e.

mghildiy:
Ok.
Are there any more opportunities,like in library,design,tools etc?

NthPortal:
I'd like to note (as I did in the original PR) that the Swing example doesn't not require an implicit conversion to work in 2.12, because of support for Java 8 lambdas.

Unfortunately, I don't have any ideas for an example to use instead.

jvican:
There are many issues you could help with in https://github.com/scala/docs.scala-lang/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc :smile: Pick the one you like the most. If you prefer another style of issue, lemme know.
  
In particular, I think that working on a robust way to have the spec in both PDFs and ebook format would be great, see this ticket: https://github.com/scala/bug/issues/10218. I believe that this can have a high impact, making the spec easier to read in different formats is always a win.

mghildiy:
Hi,

In addition to documentation related issues, I am also keen to work on
issues related to language features,libraries or tools development for
community.

jvican:
@mghildiy Let's chat privately, can you ping me in Gitter?

SethTisue:
two things that would need fixing before this could be mergeable:

* the changes to the headers aren't right. (Travis drafted his changes before the big site overhaul, so you should base the headers on what the other tour sections have, not on what's in Travis's draft.)
* you shouldn't put very long `/** comments */` inside code blocks, because they won't line wrap. it's better for these texts to be outside, in the body text, and keep any comments inside _very_ brief.

mghildiy:
Code blocks means 'tut' sections. Can you please mention here the line no which long comments.

SethTisue:
an example very long comment:

```
/** To show how implicit parameters work, we first define monoids for strings and integers. The implicit keyword indicates that the corresponding object can be used implicitly, within this scope, as a parameter of a function marked implicit. */
```

there are a number of similar such comments in the diffs.

mghildiy:
Hi,

What should be the order of file navigation?Previous page for Implicit-conversions is 'explicitly-typed-self-references'(which though doesn't exist, and I guess it should be 'self-types'). And implcits too has same previous page.

SethTisue:
> What should be the order of file navigation

the existing order at http://docs.scala-lang.org/tour/tour-of-scala.html seems fine to me

I just noticed that if this PR were merged, we'd have _three_ sections on implicits. did you mean to remove some of them...?

mghildiy:
implicit parameters and implicit conversions are now part of implicits.md. So I would remove these two older files.

SethTisue:
new PR is #1003

mghildiy:
@NthPortal , do you have a good example in mind for the section you want to be modified?

NthPortal:
@mghildiy Unfortunately I don't, sorry. My only thought is that there's probably a nice implicit conversion in the standard library that can be used; perhaps something in `Predef` (something like `Array[A] => Seq[A]`, but less complicated).

PR #999, opened 2018-01-18
Update arrays.md

lonewar:
`scala.LowPriorityImplicits`, which is inherited from `Predef`
is changed to:
`scala.LowPriorityImplicits`, which is inherited by `Predef`


SethTisue:
good catch. thanks!

PR #1002, opened 2018-01-22
New SIP: Byname implicit arguments.

milessabin:


jvican:
I'm gonna merge this without a proper review so that it's accessible in the website and can be tweeted about. If I find some time soon, I'll try to dig into the details (I'm interested).

PR #1003, opened 2018-01-26
Rewrote implicits section of the tour #746

mghildiy:


mghildiy:
I have few questions about this change.
I removed older implicit files(implicit-parameters.md,implicit-conversion.md),introduced new file(implicit.md) with previous-page pointing to self-types and next-page pointing to polymorphic-methods. 
But build failed.
Error traces suggests that I need to do similar changes for all language specific directories(ko,es,pl etc).
If yes, then how would I get language translations for implicit.md?

SethTisue:
I've answered your question at https://contributors.scala-lang.org/t/multi-language-files/1519

SethTisue:
it looks like #1018 will be merged first and you'll need to rebase.

SethTisue:
#1018 has been merged.

som-snytt:
I was going to fold changes to http://docs.scala-lang.org/tutorials/FAQ/finding-implicits.html on top of this PR, but I'm not sure this is a good foundation. I doubt whether the examples are explanatory.

I also think it's useful to have two pages for params and conversions.

For params, we need an easy typeclass example that is meaningful.

I see the current page links to that faq page; so my task is to inline it. However, I'll do that from head.

mghildiy:
To clarify things more, now that #1018 changes are in master, and as part of this PR I removed implicit-parameters.md, so now I would manually need to take new content from implicit-parameters.md and put them in implicit.md(implicit-parmeters.md is still deleted).

mghildiy:
Can anyone please confirm my last comment?

SethTisue:
Hmm, well, I've lost track of what the benefit even is of merging the two sections into one...? Is there even a consensus on making that change, or was it just an idea Travis had? It doesn't seem like a clear improvement to me. I'm not strongly against, either... I just don't get it. Maybe it just needs to be explained to me.

I'm not sure what other improvements might remain to be incorporated.

mghildiy:
I think it would be better if we close this PR.Instead, I would prefer to first really understand what latest changes in implicit-parameters.md bring to table,and then if merging of implicit files into one still makes sense, we can go further.

SethTisue:
@mghildiy okay — thanks for working on this, sorry about the overlap in work that ended up happening

PR #1004, opened 2018-01-26
Mention implicit classes and string interpolation in the tour intro (fixes #992)

### LINKED ISSUE #992: make "string interpolation" and "implicit classes" part of the tour
@smarter points out on Gitter that:

> http://docs.scala-lang.org/overviews/core/implicit-classes.html is pretty bite-sized and http://docs.scala-lang.org/overviews/core/string-interpolation.html isn't so big either compared to some pages of the tour, and it's a pretty useful feature of Scala to know about

I agree, these could become part of http://docs.scala-lang.org/tour/tour-of-scala.html instead

mghildiy:


_tour/tour-of-scala.md
```diff
@@ -53,3 +53,9 @@ In practice, the development of domain-specific applications often requires doma
 Scala is designed to interoperate well with the popular Java Runtime Environment (JRE). In particular, the interaction with the mainstream object-oriented Java programming language is as smooth as possible. Newer Java features like [annotations](annotations.html) and Java generics have direct analogues in Scala. Those Scala features without Java analogues, such as [default](default-parameter-values.html) and [named parameters](named-arguments.html), compile as close to Java as they can reasonably come. Scala has the same compilation model (separate compilation, dynamic class loading) like Java and allows access to thousands of existing high-quality libraries.
 
 Please continue to the [next page](basics.html) in the Contents menu to read more.
+
+## Implicit classes in Scala ##
+Scala offers a language construct called implicit class which simplifies the creation of classes that provide extension methods to another type.Classes annotated with the implicit keyword are referred to as implicit classes.For an implicit class to work, its name must be in scope and unambiguous.
```
SethTisue: missing spaces after the periods in this line

I'd suggest using backticks here: "Classes annotated with the `implicit` keyword"

_tour/tour-of-scala.md
```diff
@@ -53,3 +53,9 @@ In practice, the development of domain-specific applications often requires doma
...
+
+## Scala offers string interpolation ##
+With mechanism of string interpolation, Scala enables users too embed variables in string literals.Scala offers three string interpolation methods out of the box:  `s`, `f` and `raw`.Prepending `s` to any string literal allows the usage of variables directly in the string.Prepending `f` to any string literal allows the creation of simple formatted strings.The raw interpolator is similar to the `s` interpolator except that it performs no escaping of literals within the string.
```
SethTisue: too embed -> to embed

missing spaces after periods again

you can just say "string interpolators" instead of "string interpolation methods"

instead of "out of the box", I would suggest "in the standard library"

I feel like the level of detail here is probably inappropriate for this page. It isn't necessary to cover `f` and `raw` here, 95% or even 99% of the time what you see in actual code is just the `s` interpolator.

The details remain available at http://docs.scala-lang.org/overviews/core/string-interpolation.html

SethTisue:
also, this is a good PR and I expect we'll be able to merge it after a bit of tweaking, but it doesn't fix #992. what I meant in #992 was to add new sections to the tour that cover these features

mghildiy:
'what I meant in #992 was to add new sections to the tour that cover these features'.
So you  mean that in #992, we need to introduce new .md files dedicated solely to implicit classes and sring interpolation.

SethTisue:
right — not written from scratch, just relocated out of overviews

SethTisue:
replaced by #1012

PR #1005, opened 2018-01-27
Adds first pass at JDK compatibility table

### LINKED ISSUE #1000: add some kind of "JDK compatibility" guide
it would say that Scala 2.10.x works fine on Java 6, 7, and 8 (is there a minimum version that works well on 8?), and 2.10.7 has minimal Java 9 support that is unlikely to improve further

likewise for 2.11.x; 2.11.12 is the release that added minimal Java 9 support

2.12.x requires Java 8 minimum; 2.12.4 is the first version with minimal Java 9 support

and then maybe some content inspired by https://github.com/scala/scala-dev/issues/139 about remaining issues and/or missing features

I'm sure there's more to say, but that could be the initial skeleton

this would go under http://docs.scala-lang.org/overviews/index.html

colindean:
Information has not been fully verified beyond comments in #1000.

I'll happily fix the data given more input from those with a longer memory!

I think I'd prefer to represent this with a chart of some kind, perhaps a Gantt-like chart or something, but I can't think quickly of a good way to _script_ generation of this kind of chart so that it's easily updatable in the future.

Fixes #1000

### DIFF
--- a/_data/overviews.yml
+++ b/_data/overviews.yml
@@ -89,6 +89,10 @@
         The Scala REPL is a tool (<code>scala</code>) for evaluating expressions in Scala.
         <br><br>
         The <code>scala</code> command will execute a source script by wrapping it in a template and then compiling and executing the resulting program
+    - title: JDK Compatibility
+      icon: coffee
+      url: "jdk-compatibility/overview.html"
+      description: "Tables and more explaining Scala's use of the Java Virtual Machine and Java Development Kit."
 
 - category: Parallel and Concurrent Programming
   description: "Complete guides covering some of Scala's libraries for parallel and concurrent programming."
--- a/_data/overviews.yml
+++ b/_data/overviews.yml
@@ -89,10 +89,10 @@
         The Scala REPL is a tool (<code>scala</code>) for evaluating expressions in Scala.
         <br><br>
         The <code>scala</code> command will execute a source script by wrapping it in a template and then compiling and executing the resulting program
-    - title: JDK Compatibility
+    - title: JDK Version Compatibility
       icon: coffee
       url: "jdk-compatibility/overview.html"
-      description: "Tables and more explaining Scala's use of the Java Virtual Machine and Java Development Kit."
+      description: "Which Scala versions work on what JDK versions"
 
 - category: Parallel and Concurrent Programming
   description: "Complete guides covering some of Scala's libraries for parallel and concurrent programming."
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -20,7 +20,7 @@ This table shows the first Scala release in each series that functions on each J
 | 9           | 2.12.4, 2.11.12, 2.10.7                              |
 | 8           | 2.12.0, 2.11.0, 2.10.0                               |
 | 7           | 2.11.0, 2.10.0                                       |
-| 6           | 2.10.0                                               |
+| 6           | 2.11.0, 2.10.0                                       |
 
 ### JDK 9 compatibility notes
 
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -22,4 +22,10 @@ This table shows the first Scala release in each series that functions on each J
 | 7           | 2.11.0, 2.10.0                                       |
 | 6           | 2.10.0                                               |
 
-Note that JDK 9 support requires minimum sbt version 1.1.0, or 0.13.17 in the 0.13.x series.
+### JDK 9 compatibility notes
+
+As of Scala 2.12.4 and 2.11.12, **JDK 9 support is incomplete**. Notably, `scalac` will not enforce the restrictions of the Java Platform Module System, which means that code that typechecks may incur linkage errors at runtime.
+
+JDK 9 support requires minimum sbt version 1.1.0, or 0.13.17 in the 0.13.x series.
+
+For more information on JDK 9 compatibility, watch the ["Support JDK 9"](https://github.com/scala/scala-dev/issues/139 "scala/scala-dev #139") issue on GitHub.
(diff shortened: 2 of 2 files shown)


adriaanm:
Thanks! @retronym could you take a look?

_overviews/jdk-compatibility/overview.md
```diff
@@ -0,0 +1,25 @@
...
+permalink: /overviews/jdk-compatibility/:title.html
+---
+
+Scala runs primarily on the Java Virtual Machine (JVM). As Scala and the JVM improve independently over time, Scala drops compatibility with older versions of the Java Developer Kit (JDK) in order to focus development efforts on supporting new JVM features that benefit Scala.
+
+This table shows the first Scala release in each series that functions on each JDK.
+
+| JDK version | First Scala release supporting JDK version per series |
+|:-----------:|:-----------------------------------------------------|
+| 9           | 2.12.4, 2.11.12                                      |
```
retronym: Thanks for getting the ball rolling.

The Java 9 row needs an asterisk to mention that support is incomplete. Notably, scalac will not enforce the restrictions of the Java Platform Module System, which means that code that typechecks may incur linkage errors at runtime.

colindean: Added a full section on it:

https://github.com/scala/docs.scala-lang/pull/1005/files#diff-c509fc8681047cdea35034d7993f8ef2R25

_data/overviews.yml
```diff
@@ -89,6 +89,10 @@
         The Scala REPL is a tool (<code>scala</code>) for evaluating expressions in Scala.
         <br><br>
         The <code>scala</code> command will execute a source script by wrapping it in a template and then compiling and executing the resulting program
+    - title: JDK Compatibility
+      icon: coffee
+      url: "jdk-compatibility/overview.html"
+      description: "Tables and more explaining Scala's use of the Java Virtual Machine and Java Development Kit."
```
SethTisue: The title and description are considerably broader than the actual content.

What do you think about:

```
title: JDK Version Compatibility
description: "Which Scala versions work on what JDK versions"
```

If the content were later to become broader, the description could be broadened as well.

_overviews/jdk-compatibility/overview.md
SethTisue: 2.10.7 includes some minimal changes intended to improve JDK 9 support, in fact, that's the whole reason 2.10.7 even exists

_overviews/jdk-compatibility/overview.md
```diff
@@ -0,0 +1,31 @@
...
+| 8           | 2.12.0, 2.11.0, 2.10.7                               |
```
SethTisue: earlier 2.10.x releases work fine on Java 8

_overviews/jdk-compatibility/overview.md
```diff
@@ -0,0 +1,31 @@
...
+| 7           | 2.11.0, 2.10.0                                       |
+| 6           | 2.10.0                                               |
+
```
SethTisue: I think this whole table may need a rethink.

There are two kinds of information to be conveyed: when support for a particular JDK was _added_, and when it was _dropped_.

Re support being _added_, to me the table seems to give the impression that 2.11.x doesn't work on Java 6. But it does.

Re support being _dropped_, that information is missing entirely. One of the most important jobs for this page to do, IMO, is convey that e.g. if you want to use Scala 2.12, you won't be able to use JDK 7 or less.

I'm don't actually know what the ideal format for conveying this information is, I'm just suggesting that it ought to be conveyed somehow. Whether in a single combined table somehow, or two tables, or what, idk.

SethTisue:
In my review I put a comment saying that the table doesn't convey when support was dropped, but I see now that it does. I deleted that comment.

Maybe I got confused because of a few inaccuracies in the table. 2.10.7 works on JDK 9. The 2.11 series works on JDK 6. Java 8 support landed earlier in the 2.10.x series than 2.10.7 — I think maybe even 2.10.0 worked? Not entirely sure, but certainly 2.10.6 works fine on Java 8. The release notes for these release should have the information.

retronym:
Maybe we can simplify this a little by only stating the compatibility (and caveats) between the latest release in each series with each JDK version.

colindean:
Is there an authoritative source of compatibility information within the
Scala source code? Could we set up some kind of build or manual step
that retrieves some file present in the repo at each tagged revision of
scala/scala that describes the JDKs with which that version is
compatible? Single-sourcing this information _somehow_ would keep it up
to date; maintaining this page is a concern.

SethTisue:
> Is there an authoritative source of compatibility information within the
> Scala source code?

I'm afraid there isn't. and I don't think building new automation that covers this would have good cost/benefit. I think we'll just have to compile it manually and update it manually.

(basically everything in this entire repo has a tendency to contain minor errors and to drift out of date gradually... this page will hardly be unique in that respect. it's okay, we manage.)

_data/overviews.yml
colindean: 👍

_overviews/jdk-compatibility/overview.md
colindean: Addressed in https://github.com/scala/docs.scala-lang/pull/1005/commits/979f07aa1bcdf3db5faa962c54479254e8e71d3d

_overviews/jdk-compatibility/overview.md
colindean: Addressed in https://github.com/scala/docs.scala-lang/pull/1005/commits/979f07aa1bcdf3db5faa962c54479254e8e71d3d

colindean:
I tried to fill in the table as presently structured as best as I could based on comments here.

> Maybe we can simplify this a little by only stating the compatibility (and caveats) between the latest release in each series with each JDK version.

I'm concerned about the upkeep necessary for the docs team with this solution. Changing the latest release has to happen with each new release where this table as written states "this is the first release of this series that supports this JDK." That fact is ~never out of date.

SethTisue:
thank you, @colindean, for tackling this — much appreciated. I hope other folks will add to it over time.

colindean:
🎉

PR #1006, opened 2018-01-29
used proper singular. A instead of AN

IDispose:
It should be A 'SmallAnimal' instead of An 'SmallAnimal'


SethTisue:
thank you!

PR #1013, opened 2018-02-04
try to clarify with a compiling example.

martijnhoekstra:
https://gitter.im/scala/scala?at=5a773dcde217167e2c3f4ea0 shows people not understanding this page. Tried to clarify. 

The error message with the unimplemented `Ordered` doesn't give a hint that the problem is with the implementation of Ordered, rather than with `xs <= ys`.

Further, the `Int => Ordered[Int]` implementation leads to ambiguous implicits because of the implementation in `Predef`

### DIFF
--- a/_tour/implicit-conversions.md
+++ b/_tour/implicit-conversions.md
@@ -23,7 +23,13 @@ Implicit conversions are applied in two situations:
 In the first case, a conversion `c` is searched for which is applicable to `e` and whose result type conforms to `T`.
 In the second case, a conversion `c` is searched for which is applicable to `e` and whose result contains a member named `m`.
 
-If an implicit method `List[A] => Ordered[List[A]]` like the one below
+If an implicit method `List[A] => Ordered[List[A]]` is in scope, as well as an implicit method `Int => Ordered[Int]`, the following operation on the two lists of type `List[Int]` is legal:
+
+```
+List(1, 2, 3) <= List(4, 5)
+```
+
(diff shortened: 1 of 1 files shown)


_tour/implicit-conversions.md
```diff
@@ -23,21 +23,22 @@ Implicit conversions are applied in two situations:
...
+implicit def list2ordered[A](x: List[A])
+    (implicit elem2ordered: A => Ordered[A]): Ordered[List[A]] =
+  new Ordered[List[A]] { 
+    //replace with a more useful implementation
+    def compare(that: List[A]): Int = 1
+  }
 ```
 
-assuming the implicit methods `list2ordered` and `int2ordered` defined below are in scope:
+is in scope, as well as an implicit method `Int => Ordered[Int]`, one of which is provided automatically
```
hrhino: I would not split the sentence around the code block.

_tour/implicit-conversions.md
```diff
@@ -23,21 +23,22 @@ Implicit conversions are applied in two situations:
...
+through `scala.Predef.intWrapper`, the following operation on the two lists of type `List[Int]` is legal:
```
hrhino: Probably better to explain _which_ is provided by `intWrapper`.

martijnhoekstra: Ha, I meant that `intWrapper` provides "some" `Int => Ordered[Int]` out of many possible implementations, but indeed this reads totally ambiguous.

SethTisue:
did you know we have `tut` in this repo? you can add `tut` to the triple quotes before a piece of code, and then we can be sure it compiles.

martijnhoekstra:
@SethTisue hint taken.

PR #1014, opened 2018-02-05
add caveat that for methods

martijnhoekstra:
cleaned up the terminology a bit.

Maintained "higher order function" for both methods and functions, but add a warning, and don't explicitly call a method a function


martijnhoekstra:
prompted by someone actually being confused: https://gitter.im/scala/scala?at=5a77f2756117191e61f5f23b

PR #1018, opened 2018-02-14
Try to clarify implicit parameters tour

martijnhoekstra:
* Don't explain what a monoid and a semigroup are. This is not the place for that.
* Don't make `Monoid[A]` extend `Semigroup[A]`. This only complicates the example for no gain.
* Explain the lookup procedure.
* Use "Companion object" instead of "companion module"
* Don't mention templates. It's an uncommonly used 
* Don't extend `App`. It doesn't simplify the example.

Further improvements could be to show an example of the second rule of implicit lookup, and/or a chained implicit lookup.

### DIFF
--- a/_tour/implicit-parameters.md
+++ b/_tour/implicit-parameters.md
@@ -18,7 +18,9 @@ A method can have an _implicit_ parameter list, marked by the _implicit_ keyword
 The places Scala will look for these parameters fall into two categories:
 
 * Scala will first look for implicit definitions and implicit parameters that can be accessed directly (without a prefix) at the point the method with the implicit parameter block is called.
-* Then, it falls back to all members that are marked implicit in the companion object of the implicit parameter.
+* Then it looks for members marked implicit in all the companion objects associated with the implicit candidate type.
+
+A more detailed guide to where scala looks for implicits can be found in [the FAQ](//docs.scala-lang.org/tutorials/FAQ/finding-implicits.html)
 
 In the following example we define a method `sum` which computes the sum of a list of elements using the monoid's `add` and `unit` operations. Please note that implicit values can not be top-level.
 
--- a/_tour/implicit-parameters.md
+++ b/_tour/implicit-parameters.md
@@ -13,52 +13,58 @@ previous-page: self-types
 redirect_from: "/tutorials/tour/implicit-parameters.html"
 ---
 
-A method with _implicit parameters_ can be applied to arguments just like a normal method. In this case the implicit label has no effect. However, if such a method misses arguments for its implicit parameters, such arguments will be automatically provided.
+A method can have an _implicit_ parameter list, marked by the _implicit_ keyword at the start of the parameter list. If the parameters in that parameter list are not passed as usual, scala will look if it can get an implicit value of the correct type, and if it can, pass it automatically.
 
-The actual arguments that are eligible to be passed to an implicit parameter fall into two categories:
+The places scala will look for these parameters fall into two categories:
 
-* First, eligible are all identifiers x that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
-* Second, eligible are also all members of companion modules of the implicit parameter's type that are labeled implicit.
+* First, eligible are all identifiers that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
+* Second, eligible are also all members of companion objects of the implicit parameter's type that are labeled implicit.
 
-In the following example we define a method `sum` which computes the sum of a list of elements using the monoid's `add` and `unit` operations. Please note that implicit values can not be top-level, they have to be members of a template.
+In the following example we define a method `sum` which computes the sum of a list of elements using the monoid's `add` and `unit` operations. Please note that implicit values can not be top-level.
 
 ```tut
-abstract class SemiGroup[A] {
+abstract class Monoid[A] {
   def add(x: A, y: A): A
-}
-abstract class Monoid[A] extends SemiGroup[A] {
   def unit: A
 }
-object ImplicitTest extends App {
+
+object ImplicitTest {
   implicit object StringMonoid extends Monoid[String] {
     def add(x: String, y: String): String = x concat y
     def unit: String = ""
   }
+  
   implicit object IntMonoid extends Monoid[Int] {
     def add(x: Int, y: Int): Int = x + y
     def unit: Int = 0
   }
+  
   def sum[A](xs: List[A])(implicit m: Monoid[A]): A =
     if (xs.isEmpty) m.unit
     else m.add(xs.head, sum(xs.tail))
-
-  println(sum(List(1, 2, 3)))       // uses IntMonoid implicitly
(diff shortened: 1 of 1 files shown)


_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
 redirect_from: "/tutorials/tour/implicit-parameters.html"
 ---
 
-A method with _implicit parameters_ can be applied to arguments just like a normal method. In this case the implicit label has no effect. However, if such a method misses arguments for its implicit parameters, such arguments will be automatically provided.
+A method can have an _implicit_ parameter list, marked by the _implicit_ keyword at the start of the parameter list. If the parameters in that parameter list are not passed as usual, scala will look if it can get an implicit value of the correct type, and if it can, pass it automatically.
 
-The actual arguments that are eligible to be passed to an implicit parameter fall into two categories:
+The places scala will look for these parameters fall into two categories:
```
hrhino: nit: capitalize "Scala" consistently

_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
...
 
-* First, eligible are all identifiers x that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
-* Second, eligible are also all members of companion modules of the implicit parameter's type that are labeled implicit.
+* First, eligible are all identifiers that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
+* Second, eligible are also all members of companion objects of the implicit parameter's type that are labeled implicit.
```
hrhino: I know it was like this, but the phrasing "eligible are all identifiers that ..." sounds like a 19th century legal text. 

Why not something like...

> ... two categories:
> - Implicit definitions and implicit parameters that can be referenced at the point of the method call without a prefix, and
> - Members of companion objects of the implicit parameter's type that are marked implicit.

This elides the "parts" bit from the spec but I think that is needlessly complicated to mention in the tour.

_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
...
+`Monoid` defines an operation called `add` here, that combines a pair of `A`s and returns another `A`, together with an operation called `unit` that is able to create some (specific) `A`.
+
+To show how implicit parameters work, we first define monoids `StringMonoid` and `IntMonoid` for strings and integers, respectively. The `implicit` keyword indicates that the corresponding object can be used implicitly.
 
-A monoid, modeled by `Monoid` here, is a semigroup with a distinguished element of `A`, called `unit`, that when combined with any other element of `A` returns that other element again.
+The method `sum` takes a `List[A]` and returns an `A`, which takes the initial `A` from `unit`, and combines each next `A` in the list to that with the `add` method. Making the parameter `m` implicit here means we only have to provide the `xs` parameter when we call the method if scala can find a an implict `Monoid[A]` to use for the implicit `m` parameter.
```
hrhino: consistent capitalization opportunity

_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
...
 
-Method `sum` takes a `List[A]` and returns an `A`, which represents the result of applying the monoid operation successively across the whole list. Making the parameter `m` implicit here means we only have to provide the `xs` parameter at the call site, since if we have a `List[A]` we know what type `A` actually is and therefore what type `Monoid[A]` is needed. We can then implicitly find whichever `val` or `object` in the current scope also has that type and use that without needing to specify it explicitly.
+> First, eligible are all identifiers that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
 
-Finally, we call `sum` twice, with only one parameter each time. Since the second parameter of `sum`, `m`, is implicit, its value is looked up in the current scope, based on the type of monoid required in each case, meaning both expressions can be fully evaluated.
+`IntMonoid` is an identifier, it can be accessed directly in `main`, and it is an implicit definition. It is also of the correct type, so it's passed to the sum method automatically.
```
hrhino: Maybe,

> Because `IntMonoid` is a value which can be referenced directly inside of `main`, is marked implicit, and has the correct type, it's passed automatically to `sum`.

Otherwise, "sum" should probably be "`sum`".

_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
...
+abstract class Monoid[A] {
   def add(x: A, y: A): A
-}
-abstract class Monoid[A] extends SemiGroup[A] {
   def unit: A
 }
-object ImplicitTest extends App {
+
+object ImplicitTest {
   implicit object StringMonoid extends Monoid[String] {
```
hrhino: There are some gotchas with `implicit object`s last time I checked, and the type of `StringMonoid` isn't _really_ `Monoid[String]`. I think this should be `implicit val StringMonoid: Monoid[String] = ...`.

_tour/implicit-parameters.md
martijnhoekstra: The swing I took at this also addresses the two below comments. WDYT?

_tour/implicit-parameters.md
```diff
@@ -13,52 +13,58 @@ previous-page: self-types
...
+A method can have an _implicit_ parameter list, marked by the _implicit_ keyword at the start of the parameter list. If the parameters in that parameter list are not passed as usual, Scala will look if it can get an implicit value of the correct type, and if it can, pass it automatically.
 
-The actual arguments that are eligible to be passed to an implicit parameter fall into two categories:
+The places Scala will look for these parameters fall into two categories:
 
-* First, eligible are all identifiers x that can be accessed at the point of the method call without a prefix and that denote an implicit definition or an implicit parameter.
-* Second, eligible are also all members of companion modules of the implicit parameter's type that are labeled implicit.
+* Scala will first look for implicit definitions and implicit parameters that can be accessed directly (without a prefix) at the point the method with the implicit parameter block is called.
+* Then, it falls back to all members that are marked implicit in the companion object of the implicit parameter.
```
jvican: What about "Then it looks for members marked implicit in all the companion objects associated with the implicit candidate type"? A link to https://docs.scala-lang.org/tutorials/FAQ/finding-implicits.html would be useful for those that want to know more.

martijnhoekstra: Linking to the implicit lookup FAQ there is a great idea.

The phrasing "in all the companion objects associated with the implicit candidate type" isn't as "human readable" as I'd like it to be, especially for the tour. It still reads a bit as legalese. I'll think on it some more. Maybe not saying much about where it looks exactly (but link to the FAQ for details) is another option.

mghildiy:
Won't it interfere with:
https://github.com/scala/docs.scala-lang/pull/1003

As part of this work, I would remove older files related to implicits, and there would be a single implicit.md.

SethTisue:
@mghildiy yes, you'll have a bit of merging work to do in order to carry #1003 forward.

SethTisue:
thanks Martijn & reviewers, nice work here.

PR #1023, opened 2018-02-20
Update naming-conventions.md

b1n0ys:
updated the naming convention for scala object

### DIFF
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -44,6 +44,17 @@ Object should be named in lower camel case :
     object inc {
       def apply(x: Int): Int = x + 1
     }
+ 
+Unless these are companion Objects, 
+in which case Object name should be same as class name(which then be upper camel case)
+ 
+    class Car (val year : Int, val miles : Int) {
+       ....
+    }
+     
+    object Car {
+      def create(year: Int) = new Car(year, 0)
+    }
 
 ## Packages
 
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -32,10 +32,7 @@ This mimics the Java naming convention for classes.
 
 ## Objects
 
-Object names are like class names (upper camel case).
-
-An exception is when mimicking a package or function.
-This isn't common. Example:
+Object should be named in lower camel case :
 
     object ast {
       sealed trait Expr


SethTisue:
can you explain your thinking here? it seems me the original text is correct. perhaps we can make it clearer though.

b1n0ys:
isn't it correct to have all objects in scala to be names with lower camel
case

unless they are companion Objects (in which case the name should be same as
the class and hence Upper camel case)
isn't that the exception case.


Thanks & Regards,
Binoy Sundaran

SethTisue:
>isn't it correct to have all objects in scala to be names with lower camel case

no. one normally writes `object FooBar`, not `object fooBar`

b1n0ys:
I guess then that's the confusion, you are explaining a case where we
specifically use 'object' keyword.

In the case we are defining a companion object or creating a singleton
object (which is a special case i would say)

but for normal instance Objects , people should follow lower camel casing

i am talking about the objects created like these.
val foo = new Foo


Thanks & Regards,
Binoy Sundaran

SethTisue:
ah, I see. yes, there is sometimes confusion in conversations about Scala whether the word "object" refers to `object` specifically or just to any value (since although Scala has both value types and reference types, they're all considered "objects" in that, unlike primitive value types in Java, you can call methods on them)

SethTisue:
later on the same page it says "Method, Value and variable names should be in lower camel case", which is intended to cover the `val foo = new Foo` case. but perhaps some wording and/or structural changes would help make this clearer

_style/naming-conventions.md
```diff
@@ -47,6 +44,17 @@ This isn't common. Example:
     object inc {
       def apply(x: Int): Int = x + 1
     }
+ 
+Unless these are companion Objects, 
+in which case Object name should be same as class name(which then be upper camel case)
+ 
+    class Car (val year : Int, val miles : Int) {
```
jvican: If the template is empty, can we remove it?

_style/naming-conventions.md
```diff
@@ -32,10 +32,7 @@ This mimics the Java naming convention for classes.
 
 ## Objects
 
-Object names are like class names (upper camel case).
-
-An exception is when mimicking a package or function.
-This isn't common. Example:
+Object should be named in lower camel case :
```
jvican: I think it's commonplace in the Scala community to always use upper camel case for `object`. I've rarely seen lower camel case for objects, I've only seen it for `package object`s.

jvican:
Oh sorry, just read the last comments. I see the confusion. Probably the best thing is to rewrite the whole paragraph to avoid it.

jvican:
@b1n0ys Do you have time to follow up on this? :smile:

SethTisue:
closing for inactivity. 

see also https://contributors.scala-lang.org/t/what-happened-to-module/1790

PR #1024, opened 2018-02-22
packages-and-imports.md - remove next button

coryveilleux:
packages-and-imports.md is the last page in the tour, so I'm removing the next link/button.


SethTisue:
nice catch, thank you 👍

coryveilleux:
I only noticed because the docs are so good and I made it to the last page. Thank you!

PR #1026, opened 2018-02-24
Fix redundant sentences and one example.

houcros:


jvican:
Thanks for this change! :smile:

PR #1030, opened 2018-03-08
Some more opaque type SIP improvements

non:
This PR does a few things:

 - Tries to add more motivation for opaque types
 - Adds another example (`IArray`) of how opaque types can be used
 - Adds a more detailed comparison of opaque types and value classes

Review by @jvican. Sorry it's so last-minute!

### DIFF
--- a/_sips/sips/2017-09-20-opaque-types.md
+++ b/_sips/sips/2017-09-20-opaque-types.md
@@ -763,11 +763,12 @@ pattern:
  3. Return the array, which is now treated as immutable.
 
 In this pattern, the vast majority of time is spent in the third step,
-where the arrays compactness and speed of iteration/access provide
+where the array's compactness and speed of iteration/access provide
 major wins over other data types.
 
 This pattern is currently only enforced by convention. However, opaque
-types create an opportunity to provide more safety without overhead:
+types create an opportunity to provide more safety without incurring
+any overhead:
 
 ```scala
 package object ia {
--- a/_sips/sips/2017-09-20-opaque-types.md
+++ b/_sips/sips/2017-09-20-opaque-types.md
@@ -886,7 +886,15 @@ possibly the runtime performance).
 By contrast, replacing `String` with `Digits` is guaranteed to have no
 impact (all occurances of `Digits` are guaranteed to be erased to
 `String`). Aside from the ergonomics of calling the `fromString` and
-`asString` methods, there's no runtime impact of any kind.
+`asString` methods, there's no runtime impact versus using the
+underlying type.
+
+One wrinkle to the above is that built-in primitive types will
+naturally box in some situations (but not others). For example
+`List[Double]` will be represented as a `List[java.lang.Double]`,
+`(Double, Double)` will be represented as a `scala.Tuple2$mcDD$sp`,
+and so on. In these cases, an opaque type will exhibit the same
+behavior.
 
 ### Default visibility
 
--- a/_sips/sips/2017-09-20-opaque-types.md
+++ b/_sips/sips/2017-09-20-opaque-types.md
@@ -32,7 +32,7 @@ Some use cases for opaque types are:
    are expanded in-place.
 
  * New numeric classes, such as unsigned integers. There would no
-   longer need to be a boxing overhead for such classes. So this is
+   longer need to be a boxing overhead for such classes. This is
    similar to value types in .NET and `newtype` in Haskell. Many APIs
    currently use signed integers (to avoid overhead) but document that
    the values will be treated as unsigned.
--- a/_sips/sips/2017-09-20-opaque-types.md
+++ b/_sips/sips/2017-09-20-opaque-types.md
@@ -27,9 +27,15 @@ time.
 
 Some use cases for opaque types are:
 
- * New numeric classes, such as unsigned ints. There would no longer
-   need to be a boxing overhead for such classes. So this is similar
-   to value types in .NET and `newtype` in Haskell.
+ * Implementing type members while retaining parametricity. Currently,
+   concrete `type` definitions are treated as type aliases, i.e. they
+   are expanded in-place.
+
+ * New numeric classes, such as unsigned integers. There would no
+   longer need to be a boxing overhead for such classes. So this is
+   similar to value types in .NET and `newtype` in Haskell. Many APIs
+   currently use singed integers (to avoid overhead) but document that
+   the values will be treated as unsigned.
 
  * Classes representing units of measure. Again, no boxing overhead
    would be incurred for these classes.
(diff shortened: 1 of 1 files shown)


_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -27,9 +27,15 @@ time.
...
 
- * New numeric classes, such as unsigned ints. There would no longer
-   need to be a boxing overhead for such classes. So this is similar
-   to value types in .NET and `newtype` in Haskell.
+ * Implementing type members while retaining parametricity. Currently,
+   concrete `type` definitions are treated as type aliases, i.e. they
+   are expanded in-place.
+
+ * New numeric classes, such as unsigned integers. There would no
+   longer need to be a boxing overhead for such classes. So this is
```
johnynek: The “So this” is a weird start to the sentence. Maybe delete “So”?

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,178 @@ This example is somewhat similar to `Logarithm` above. Other
...
+By default, a value class' constructor and accessor are public. It
+*is* possible to restrict access, using a `private` constructor and
+`private` accessor, but this makes the comparison between opaque types
+and value classes less attractive:
+
+```scala
+package object pkg {
+  opaque type XCoord = Int
+
+  case class private[pkg] YCoord (private[pkg] val Int) extends AnyVal
```
johnynek: Is the first private in the right place here? I think it should be after YCoord.

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,178 @@ This example is somewhat similar to `Logarithm` above. Other
...
+`null` (otherwise this could create ambiguities between their boxed
+and unboxed representations when wrapping `AnyRef` values).
+
+By contrast, when seen from the "outside" opaque types extend
+`Any`. Their bounds are the same as those of a type parameter without
+explicit bounds (e.g. `A <: Any >: Nothing`).
+
+### Size of artifacts produced
+
+Since value classes do have a runtime reprsentation, they do increase
```
johnynek: Representation is misspelled.

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,178 @@ This example is somewhat similar to `Logarithm` above. Other
...
+the size of runtime artifacts produced (whether a JAR file, or a
+javascript file). Their methods are also compiled to multiple
+representations (i.e. they support both the boxed and unboxed forms
+via extensions methods). Again, this comes at a cost.
+
+By contrast, opaque types have no inherent runtime footprint. The
```
johnynek: An example might be useful here. That said, maybe this is actually overstated. If we have methods they have to go somewhere so that will be a class, and so really we are arguing for smaller code. I think it could be true, but it is not 100% obvious without knowing the full encoding. Lastly, I do think reduction of compiled code side is important for scala. It is not uncommon to see it as a complaint and some say one of the reasons scala has not been popular for android developers.

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,179 @@ This example is somewhat similar to `Logarithm` above. Other
...
+```scala
+package object pkg {
+  opaque type XCoord = Int
+
+  case class private[pkg] YCoord (private[pkg] val n: Int) extends AnyVal
+
+  // in both cases, we'd need public factory constructors
+  // to allow users to produce values of these types.
+}
+```
```
jvican: I'd add another paragraph here saying: "This default of opaque types optimizes for the most sanest case when defining wrapper types."

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,179 @@ This example is somewhat similar to `Logarithm` above. Other
...
+### LUB differences
+
+Value classes extend `AnyVal` by virtue of the syntax used to define
+them. One reason this is necessary is that value classes cannot be
+`null` (otherwise this could create ambiguities between their boxed
+and unboxed representations when wrapping `AnyRef` values).
+
+By contrast, when seen from the "outside" opaque types extend
+`Any`. Their bounds are the same as those of a type parameter without
+explicit bounds, i.e. `A <: Any >: Nothing`.
```
jvican: Can you make more explicit why this is important for users/library authors? :smile:

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,207 @@ This example is somewhat similar to `Logarithm` above. Other
...
+ * `Monoid[Alphabetic]`
+
+In many cases users won't mind the fact that this code will box, but
+there will certainly be an impact on the bytecode produced (and
+possibly the runtime performance).
+
+By contrast, replacing `String` with `Digits` is guaranteed to have no
+impact (all occurances of `Digits` are guaranteed to be erased to
+`String`). Aside from the ergonomics of calling the `fromString` and
+`asString` methods, there's no runtime impact of any kind.
```
sjrd: You should mention that this is only true when the underlying type is an `AnyRef` type. If the underlying type is a primitive, e.g., `Double`, you will trading boxing a primitive `double` inside an `Alphabetic` for boxing the same primitive `double` inside a `java.lang.Double`.

non: Added a section to hopefully make this clearer.

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,179 @@ This example is somewhat similar to `Logarithm` above. Other
...
+    -lower - 1
+  }
+}
+```
+
+This example is a bit different from others: there's no attempt to
+enrich the `IArray` type with syntactic conveniences. Rather, the goal
+is to show that traditional, "low-level" code with `Array`, `Int`,
+etc. can be written with opaque types without sacrificing any
+performance.
```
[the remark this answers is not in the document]
non: Agreed. done.

_sips/sips/2017-09-20-opaque-types.md
```diff
@@ -736,6 +753,179 @@ This example is somewhat similar to `Logarithm` above. Other
...
+In this pattern, the vast majority of time is spent in the third step,
+where the array's compactness and speed of iteration/access provide
+major wins over other data types.
+
+This pattern is currently only enforced by convention. However, opaque
+types create an opportunity to provide more safety without incurring
+any overhead:
+
+```scala
+package object ia {
```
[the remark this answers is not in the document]
non: 😎

_sips/sips/2017-09-20-opaque-types.md
non: I added something similar.

jvican:
I'm adding some information on value types vs opaque types here. At some point I'll put everything together in a better way in the document.
  
I've been having a look at all the Valhalla value types information. It seems they are still figuring out what the best way of representing value types is (which type descriptor hierarchy in the JVM is the most appropriate).
  
They want to push hard for getting value types extending interfaces (I-types), in a similar way to what we have in Scala with value classes + universal traits *but* without boxing. However, it's not even clear that is possible and is considered as a "potential" use case that could not be supported given the difficulty of the task.
  
One thing is clear in all the texts I've read about value types; they are identity-less, so there's no clear way they'll be able to outperform the approach that we take in opaque types (where, if you need identity and to override `hashCode` and `equals`, we recommend you use a plain class).

PR #1032, opened 2018-03-09
Fix typo in maps.md

crukedsh:


SethTisue:
nice catch — thanks!

I fixed it in the Japanese and Chinese translations as well in b85d0362bc6905b10398aeecbcbc240e192f481f

PR #1036, opened 2018-03-18
port compiler plugins guide from old site

SethTisue:
not not for merge yet


SethTisue:
merged since it's better than nothing (and better than the original http://www.scala-lang.org/old/node/140), but it definitely still needs more work. there is a bunch of content from my "Compiler Plugins 101" talk that I will merge in here after the Northeast Scala Symposium is done. at that point, it will be ready for community contributions.

PR #1041, opened 2018-03-21
Fixes minor typo in docs on Implicit Parameter page

angadgill:
"directly" was written twice on the page. This PR fixed that :)


angadgill:
👍

PR #1042, opened 2018-03-22
Elaborate on our policy for supporting JVM versions

adriaanm:
There are two axes regarding the required or supported JVM version: 
  - running Scala code, 
  - compiling Scala code,

and on both there is a lower and upper bound of the supported JVM.


_overviews/jdk-compatibility/overview.md
```diff
@@ -13,18 +13,25 @@ permalink: /overviews/jdk-compatibility/:title.html
...
 | 6           | 2.11.0, 2.10.0                                       |
 
+### Running versus compiling and required / supported JVM
+For most users, we recommend using Java 8 for *compiling* (and running) Scala code. Since the JVM is backward compatible, it is usually safe to use a newer JVM to *run* your code compiled by the Scala compiler for older JVM versions. There are notable exceptions with experimental/unsafe features, and the introduction of the module system in Java 9. The Scala compiler does usually need updates to run properly on newer versions of the JVM, so make sure to use the appropriate JVM when compiling your code.
+
+Issues with using the Scala compiler on *non-LTS* versions of Java will not necessarily be considered blockers for releases, but we will do our best to run CI on more versions of Java to catch bugs early, and to fix them as quickly as reasonably possible. If regressions do occur with non-LTS (or, generally, unsupported) versions of Java, we may bring the next minor release deadline a bit closer, so that these issues are generally resolved within a month or two. Lightbend does offer commercial support for faster resolution of issues like this.
```
retronym: We should define "LTS".

_overviews/jdk-compatibility/overview.md
```diff
@@ -13,18 +13,25 @@ permalink: /overviews/jdk-compatibility/:title.html
...
-| JDK version | First Scala release supporting JDK version per series |
+| JVM version | First Scala compiler release supported to run on this JVM |
 |:-----------:|:-----------------------------------------------------|
 | 9           | 2.12.4, 2.11.12, 2.10.7                              |
 | 8           | 2.12.0, 2.11.0, 2.10.0                               |
 | 7           | 2.11.0, 2.10.0                                       |
 | 6           | 2.11.0, 2.10.0                                       |
 
+### Running versus compiling and required / supported JVM
+For most users, we recommend using Java 8 for *compiling* (and running) Scala code. Since the JVM is backward compatible, it is usually safe to use a newer JVM to *run* your code compiled by the Scala compiler for older JVM versions. There are notable exceptions with experimental/unsafe features, and the introduction of the module system in Java 9. The Scala compiler does usually need updates to run properly on newer versions of the JVM, so make sure to use the appropriate JVM when compiling your code.
```
retronym: This limits the user from being able to call Java 9 APIs. Perhaps we should be explicit about that restriction.

retronym:
There are probably more steps in our support.

1. Code that was compiled against and with JDK8 may be executed with Java 9+
2. JDK9+ may be used as the development JDK, which will allow calls to `@since 9` APIs. However, the Scala compiler does not enforce JPMS module system.
3. The Scala compiler understands JPMS, and will flag errors when calling non-exported APIs, both for the `java.base` module and third party libraries on the module path.
4. The Scala standard library and scala-xml, etc are equipped with the metadata to be used as JPMS modules.

These might require a blog to articulate and relate to our roadmap, so the approximation you've started with could be enough for this document.

_overviews/jdk-compatibility/overview.md
SethTisue: done in dce41f9

SethTisue:
Jason has made some suggestions that aren't addressed, but I suggest we merge anyway since this is topical right now because of the regression in 2.12.5. We can always improve it more later.

_overviews/jdk-compatibility/overview.md
```diff
@@ -5,30 +5,39 @@ discourse: true
...
+
+And as Scala and the JVM improve independently over time, Scala may drop compatibility with older JVM versions, in order to better take advantage of new JVM features.
 
 ## Version compatibility table
 
-This table shows the first Scala release in each series that functions on each JVM release.
+This table shows the first Scala release in each series that works with each JVM release.
 
-| JVM version | First Scala compiler release supported to run on this JVM |
+| JVM version | Minimum Scala versions                               |
```
dwijnand: to me specifying "compiler" here was helpful..

retronym:
I'll loop back to update this to complement my forthcoming blog on the topic.

PR #1045, opened 2018-03-23
reorganize Overviews page

SethTisue:
* split "Core Scala" into "Language" and "Standard Library"
* make a new "Compatibility" section
* remove obsolete actors guides

### DIFF
--- a/_data/overviews.yml
+++ b/_data/overviews.yml
@@ -1,6 +1,27 @@
 
-- category: Core Scala
-  description: "Guides and overviews covering central libraries in the Scala standard library, core language features, and more."
+- category: Language
+  description: "Guides and overviews covering features in the Scala language."
+  overviews:
+    - title: String Interpolation
+      icon: usd
+      url: "core/string-interpolation.html"
+      description: >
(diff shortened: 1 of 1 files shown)


SethTisue:
the main thing I wanted to do here was collect the three compatibility guides into their own section.

SethTisue:
now live at http://docs.scala-lang.org/overviews/index.html

heathermiller:
I wouldn't delete the old Scala Actors guide. People are still using it, as far as I've heard... Maybe it makes more sense to put this in another category, like "Legacy" or something?

heathermiller:
Otherwise, it looks good!

One minor nitpick–would it make sense to shuffle the order of the sections? E.g., to have the standard library section come first? And the language section to come afterwards? I'd expect more people are interested in the collections guide than the value classes and universal traits docs, which come first on the page.

SethTisue:
> Maybe it makes more sense to put this in another category, like "Legacy" or something?

note that I didn't delete the guides, just the links. but yeah, on second thought actually I think you're right, a legacy section is a good idea.

> have the standard library section come first

good idea

I'll make a new PR with these two changes.

heathermiller:
❤️

PR #1046, opened 2018-03-23
improve structure & appearance of JDK compatibility page

SethTisue:
replace multi-page style with single-page style, since it is in fact a single page and can be expected to remain that way


SethTisue:
merging now since it's a clear bugfix, and to facilitate merging with #1042

SethTisue:
this is now live at http://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

PR #1050, opened 2018-04-05
Grammar fix: amount vs number

giftig:
Change "amount" to "number" in cases where amount is incorrect. One instance actually used number correctly in the next sentence :P

Amount refers to a continuous value as opposed to a discrete count, so number of processors / threads, amount of water or code.


SethTisue:
gratitude from a fellow grammarian

PR #1051, opened 2018-04-07
Removed references to deprecated onSuccess and onFailure methods.

iwalt:
I've tried to provide the equivalent examples and discussion; removed a couple of paragraphs that no longer seemed quite relevant (notably the little explanation of partial functions).

Hope this is okay.

### DIFF
--- a/_overviews/core/futures.md
+++ b/_overviews/core/futures.md
@@ -824,9 +824,8 @@ When asynchronous computations throw unhandled exceptions, futures
 associated with those computations fail. Failed futures store an
 instance of `Throwable` instead of the result value. `Future`s provide
 the `failed` projection method, which allows this `Throwable` to be
-treated as the success value of another `Future`.
-The following special exceptions are
-treated differently:
+treated as the success value of another `Future`. The following special
+exceptions are treated differently:
 
 1. `scala.runtime.NonLocalReturnControl[_]` -- this exception holds a value
 associated with the return. Typically, `return` constructs in method
--- a/_overviews/core/futures.md
+++ b/_overviews/core/futures.md
@@ -338,47 +338,35 @@ a `List[String]`-- a list of recent textual posts:
 
 The `onComplete` method is general in the sense that it allows the
 client to handle the result of both failed and successful future
-computations. To handle only successful results, the `onSuccess`
-callback is used (which takes a partial function):
+computations. In the case where only successful results need to be
+handled, the `foreach` callback can be used:
 
     val f: Future[List[String]] = Future {
       session.getRecentPosts
     }
 
-    f onSuccess {
-      case posts => for (post <- posts) println(post)
+    f foreach { posts =>
+      for (post <- posts) println(post)
     }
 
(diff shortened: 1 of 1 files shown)


_overviews/core/futures.md
```diff
@@ -338,47 +338,35 @@ a `List[String]`-- a list of recent textual posts:
...
-
-    val f = Future {
-      2 / 0
-    }
-
-    f onFailure {
-      case npe: NullPointerException =>
-        println("I'd be amazed if this printed out.")
-    }
+The `failed.foreach` callback is only executed if the future fails, that
```
giftig: That's technically true, but doesn't really explain what's going on. It'd be more accurate to say that `future.failed` converts the `Future[Failed[Throwable]]` to a `Future[Success[Throwable]]` to allow `foreach` to operate on the `Throwable`, and will produce a `Future[Failure[NoSuchElementException]]` if the original future succeeded, essentially swapping the position of the values.

I think probably the wording can just be a little simpler though, especially as there's a link to a projections anchor further down the page.

"To handle failed results, you can first use the `failed` projection to convert the `Failure[Throwable]` to a `Success[Throwable]` and then use `foreach` on the newly-successful `Future` instead."

I do note that the projections section clarifies that the `failed` projection is to enable handling exceptions with for comprehensions, though, so I wonder if there's actually an advantage to `future.failed.foreach` over simply

```scala
future onComplete {
  case Success(_) =>
  case Failure(t) => ... 
```
given that the former requires creating a new `Future`.

_overviews/core/futures.md
```diff
@@ -844,8 +823,9 @@ by the clients-- they can only be called by the execution context.
 When asynchronous computations throw unhandled exceptions, futures
 associated with those computations fail. Failed futures store an
 instance of `Throwable` instead of the result value. `Future`s provide
-the `onFailure` callback method, which accepts a `PartialFunction` to
-be applied to a `Throwable`. The following special exceptions are
+the `failed` projection method, which allows this `Throwable` to be
+treated as the success value of another `Future`.
+The following special exceptions are
```
giftig: I guess you were trying to avoid touching the next line unnecessarily but it's made the hard-wrapping a little untidy.

_overviews/core/futures.md
iwalt: I agree that `failed.foreach` doesn't really add very much (indeed if this was particularly useful then I expect `onFailure` wouldn't be deprecated).

So how about we just remove this failure-scenario discussion/example, replace it with your "to handle failed results" paragraph and a link to the 'Projections' section where it can be properly explained?

giftig: That makes sense to me.

iwalt: Okay so I've went a bit further to try and make things immediately obvious if the link is followed. See what you think..

SethTisue:
LGTM, perhaps @viktorklang would like to take a quick look

SethTisue:
thank you, Ian!

PR #1055, opened 2018-04-22
Add "build project" step and misc. improvements

nnutter:
For me, after `sbt` downloads ScalaTest the `CubeCalculatorTest` class
will still have several errors because ScalaTest has not been
compiled/indexed by IntelliJ IDEA so this explicitly adds a
"build project" step to trigger that.

Before I did that I wanted to fix a few other issues I noticed as I reviewed
the Markdown.


jvican:
Hmm, there’s something weird. Building the project should not have anything to do with IntelliJ indexing. For example, I never build with IntelliJ and still get indexing, autocompleions and all the juicy language experience.

nnutter:
@jvican You're suspicion was correct.  I jumped to a conclusion.  It turns out it was just that the `sbt` sync hadn't finished.  How do you feel about my correction?  Worth keeping it and the other changes?

PR #1056, opened 2018-04-24
Use println() rather than print()

XD-DENG:
Hi there,

Not sure if there is any specific reason to use `print()` here, but it doesn't print out anything in the console of **Scalafiddle**. This may be confusing/misleading to beginners like myself.

So I would like to suggest to use `println()` here in the sample code, which can print properly, rather than print().

Please let me know your thought.

### DIFF
--- a/_tour/for-comprehensions.md
+++ b/_tour/for-comprehensions.md
@@ -42,7 +42,7 @@ def foo(n: Int, v: Int) =
 
 foo(10, 10) foreach {
   case (i, j) =>
-    print(s"($i, $j) ")  // prints (1, 9) (2, 8) (3, 7) (4, 6) (5, 5)
+    println(s"($i, $j) ")  // prints (1, 9) (2, 8) (3, 7) (4, 6) (5, 5)
 }
 
 ```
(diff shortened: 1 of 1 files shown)


_tour/for-comprehensions.md
```diff
@@ -42,7 +42,7 @@ def foo(n: Int, v: Int) =
 
 foo(10, 10) foreach {
   case (i, j) =>
-    print(s"($i, $j) ")  // prints (1, 9) (2, 8) (3, 7) (4, 6) (5, 5)
+    println(s"($i, $j) ")  // prints (1, 9) (2, 8) (3, 7) (4, 6) (5, 5)
```
MasseGuillaume: the output would be 
```scala
(1, 9)
(2, 8)
(3, 7)
(4, 6)
(5, 5)
```

SethTisue: true, but not sure the difference is worth the vertical space

XD-DENG:
Hi @SethTisue @MasseGuillaume should I close this PR? Please advise. Thanks.

SethTisue:
thanks @XD-DENG!

XD-DENG:
Thank you @SethTisue

PR #1061, opened 2018-04-26
Bug in unapply

luiru72:
Split always returns a head, so None is never returned in the original code. The proposed change checks whether there is more than one element.


SethTisue:
good catch, thanks for the fix!

PR #1064, opened 2018-05-02
Update iterators.md

dlinov:
Other possible change is `res5: java.lang.String = of`. Checked against Scala 2.12.5 and 2.11.12.


SethTisue:
do you think the text following:

> Note again that `it` was changed by the call to `dropWhile`: it now points to the second word "number" in the list.

needs updating as well?

SethTisue:
merging since it's strictly an improvement. but further attention to following text would still be welcome

SethTisue:
thank you Dmitry!

PR #1066, opened 2018-05-04
FAQ rejiggered for implicit scope

som-snytt:
This was from late last night, sorry if it's not coherent or brief enough.

To clarify lexical scoping and name binding, split the description into the lexical side, which is familiar, and then implicit scope, which is unfamiliar.


SethTisue:
nice! this is great.

PR #1068, opened 2018-05-08
Calc example, add note on calc object

tbye-pfp:
The code for the Calc example will not compile as is.  It can be very frustrating to someone trying to learn a new language to figure out how to get the example code to compile.  I added a note about wrapping some of the code in a Calc object.  There is probably a better way to include that in the tutorial, but I think this is better than nothing.


SethTisue:
thank you Tony for the fix, and sorry for the frustration.

PR #1069, opened 2018-05-10
Updated page metadata to include in tour

bradfordlynch:
The Traits page wasn't being included in the tour TOC, but was being referenced by the tour pages before and after it, Classes and Class Composition with Mixins respectively. Updated the page metadata to include it in the tour, and fix the other tags.

### DIFF
--- a/_tour/abstract-types.md
+++ b/_tour/abstract-types.md
@@ -4,13 +4,14 @@ title: Abstract Types
 
 discourse: true
 
-tutorial: scala-tour
-categories: tour
+partof: scala-tour
 num: 23
 next-page: compound-types
 previous-page: inner-classes
+topics: abstract types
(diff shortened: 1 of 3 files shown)


Philippus:
the same goes for the 'Abstract Types' and 'Self-type' pages. Maybe fix those in the same PR if you have time?

bradfordlynch:
Thanks for spotting those as well. I've updated my PR to include those changes too.

bradfordlynch:
@Philippus Is there anything else you need me to do to get this merged?

Philippus:
Looks good to me! I don't have merge rights. Maybe @SethTisue can do the honours?
With the Scala Days conference happening right now it could take a little longer.

SethTisue:
thank you, Bradford! very good catch

PR #1070, opened 2018-05-11
Modify the first two paragraphs

Patchino:
The translation of the first two paragraphs is not accurate enough, thus it makes confusion.


SethTisue:
thank you!

PR #1071, opened 2018-05-13
Update futures.md

bartecheck:
erroneous 'method'

There is already 'methods' word, additional 'method' is an error.


SethTisue:
thank you!

PR #1076, opened 2018-05-27
Mini-SIP for case if

som-snytt:
Cf https://github.com/scala/scala/pull/6241

### DIFF
--- a/_sips/sips/2018-05-26-case-if.md
+++ b/_sips/sips/2018-05-26-case-if.md
@@ -1,0 +1,95 @@
+---
+layout: sips
+discourse: true
+title: SIP-NN - Uncluttering Abuse of Match
+
+vote-status: new
+vote-text:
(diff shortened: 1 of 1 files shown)


_sips/sips/2018-05-26-case-if.md
```diff
@@ -0,0 +1,95 @@
...
+title: SIP-NN - Uncluttering Abuse of Match
+
+vote-status: new
+vote-text:
+permalink: /sips/:title.html
+redirect_from: /sips/pending/case-if.html
+
+---
+
+**By: Som Snytt and A. P. Marki**
```
martijnhoekstra: Nice

jvican:
Thanks for the proposal. Let’s merge and I’ll try to see if we can have a quick discussion about the proposal in the next meeting.

PR #1077, opened 2018-05-27
Words about Predef in REPL

som-snytt:


SethTisue:
since `any2stringadd` has been deprecated for 2.13.0-M5, perhaps a different example would be better...?

som-snytt:
It's still the best example. How about if I follow up when 2.13 comes out?

Issue #995, opened 2018-01-17
Update the collections guide to the new collections framework


julienrf:
http://docs.scala-lang.org/overviews/collections/introduction.html

Most of the content should be still valid. Here are the actions needed:
- make a new guide for 2.13 so that it coexists with the current guide (maybe not! we should first try to keep a single guide with version-specific sections when necessary),
- remove the “Traversable” page and merge it with the “Iterable” page,
- remove the “migrating from 2.7” page (only if we do two guides),
- add a page about non-strict collections: views, lazy list, iterators.

SethTisue:
we should probably keep both the 2.11/2.12 and the 2.13 editions online, since people are still going to be using 2.11 and 2.12 for some years to come

mghildiy:
Actions needed:
1. Merge trait-traversable with trait-iterable.
2. Remove migrating-from-scala-27

Am I right here?

jvican:
I think so, @julienrf are you working on this or is this free to be picked up by contributors?

julienrf:
@mghildiy I have updated the issue description with bullet points for the actions needed. You can try to give them a shot if you want.

PR #1078, opened 2018-05-28
Port “The Architecture of Scala Collections” to the new architecture

julienrf:
Instead of replacing the existing documentation, it creates a new
page for 2.13’s collections.


julienrf:
I picked the contents of #960 and moved it to a new documentation page (without removing the existing documentation page):

index

I updated the contents so that it works with the last changes we applied to the collections framework.

There is still an area where we could improve the framework (and where we would have to update this document): scala/collection-strawman#325 should give us reusable builder-based operation implementations (in the current document we re-implement `map`, `concat` and a few others).

I addressed most of the feedback given in #960 by @lrytz and @SethTisue.

lrytz:
What do you think should be done about the collections user guide (https://docs.scala-lang.org/overviews/collections/introduction.html)? Most content would hopefully be the same, so there it's maybe better to have a single version with a version-specific sections where necessary.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+import scala.collection._
+import scala.collection.mutable.{ GrowableBuilder, Builder }
+
+class PrefixMap[A]
+  extends mutable.Map[String, A]
+    with mutable.MapOps[String, A, mutable.Map, PrefixMap[A]]
+    with StrictOptimizedIterableOps[(String, A), mutable.Iterable, PrefixMap[A]] {
+
+  var suffixes: immutable.Map[Char, PrefixMap[A]] = immutable.Map.empty
+  var value: Option[A] = None
```
ches: Access modifiers?

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+### Acknowledgement ###
+
+These pages contain material adapted from the 2nd edition of
+[Programming in Scala](http://www.artima.com/shop/programming_in_scala) by
+Odersky, Spoon and Venners. We thank Artima for graciously agreeing to its
+publication.
+
+## Appendix: Methods to overload to support the “same result type” principle ##
+
+You want to add overloads to specialize a transformation operations such that they return a more specific result. Examples are:
```
ches: "specialize transformations operations", minus "a". Also "return a more specific result _type_" seems more precise.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+- `map`, on `StringOps`, when the mapping function returns a `Char`, should return a `String` (instead of an `IndexedSeq`),
+- `map`, on `Map`, when the mapping function returns a pair, should return a `Map` (instead of an `Iterable`),
+- `map`, on `SortedSet`, when an implicit `Ordering` is available for the resulting element type, should return a
+`SortedSet` (instead of a `Set`).
+
+The following table lists transformation operations that might return a too wide type. You might want to overload
```
ches: Subtle, but "return a too wide type" is slightly awkward, "return too wide a type" more natural. Or "return an undesirably wide type" is perhaps an alternative.

lrytz:
Agree that we should explain more about how views are implemented, what role iterators play.

There should also be some overview on the differences between views, iterators and lazy lists, but that should go in the collections user guide, not here.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+scala> bases.to(RNA2)
+                ^
+       error: type mismatch;
+        found   : RNA2.type
+        required: scala.collection.Factory[Base,?]
+~~~
+
+### Final version of RNA strands class ###
+
+~~~ scala
```
lrytz: some missing imports

julienrf:
> What do you think should be done about the collections user guide (https://docs.scala-lang.org/overviews/collections/introduction.html)? Most content would hopefully be the same, so there it's maybe better to have a single version with a version-specific sections where necessary.

Good point, I’ve updated https://github.com/scala/docs.scala-lang/issues/995 to mention this idea.

julienrf:
> Agree that we should explain more about how views are implemented, what role iterators play.

OK, I’ll add a short section about views, but I think they should be covered more deeply in #995.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+However, in all these cases, to build the right kind of collection
+you need to start with an empty collection of that kind. This is
+provided by the `empty` method, which is the last method defined in
+`PrefixMap`. This method simply returns a fresh `PrefixMap`.
+
+We'll now turn to the companion object `PrefixMap`. In fact it is not
+strictly necessary to define this companion object, as class `PrefixMap`
+can stand well on its own. The main purpose of object `PrefixMap` is to
+define some convenience factory methods. It also defines an implicit
+`Factory` for a better interoperability with other collections.
```
lrytz: This should be explained in more detail. Why do we need it / are those in the std lib not enough? What does it do?

_overviews/core/architecture-of-scala-213-collections.md
julienrf: Thanks for the suggestions!

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+exactly one element `x`, if the option value is `Some(x)`.
+
+However, in all these cases, to build the right kind of collection
+you need to start with an empty collection of that kind. This is
+provided by the `empty` method, which is the last method defined in
+`PrefixMap`. This method simply returns a fresh `PrefixMap`.
+
+We'll now turn to the companion object `PrefixMap`. In fact it is not
+strictly necessary to define this companion object, as class `PrefixMap`
+can stand well on its own. The main purpose of object `PrefixMap` is to
```
lrytz: Maybe we should talk about when it's necessary to have a companion, what the `XFactory` traits are for etc.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+## Appendix: Methods to overload to support the “same result type” principle ##
+
+You want to add overloads to specialize transformation operations such that they return a more specific result type. Examples are:
+- `map`, on `StringOps`, when the mapping function returns a `Char`, should return a `String` (instead of an `IndexedSeq`),
+- `map`, on `Map`, when the mapping function returns a pair, should return a `Map` (instead of an `Iterable`),
+- `map`, on `SortedSet`, when an implicit `Ordering` is available for the resulting element type, should return a
+`SortedSet` (instead of a `Set`).
+
+The following table lists transformation operations that might return an undesirably wide type. You might want to overload
+these operations to return a more specific type.
```
lrytz: Maybe give some explanation when this is necessary - basically when the collection fixes some type parameter, like LongMap, BitSet, etc?

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+  CrossVersion.partialVersion(scalaVersion.value) match {
+    case Some((2, n)) if n >= 13 => sourceDir / "scala-2.13+"
+    case _                       => sourceDir / "scala-2.13-"
+  }
+}
+~~~
+
+And then you can define a Scala 2.13 compatible implementation of your collection
+in the `src/main/scala-2.13+` source directory, and an implementation for the
+previous Scala versions in the `src/main/scala-2.13-` source directory.
```
lrytz: We could include links to projects, like the scalacheck PR (https://github.com/rickynils/scalacheck/pull/411)

lrytz:
Where would we document how to write extension methods for collections? Here or in the user guide?

julienrf:
@lrytz That’s a good question. On one hand extension methods are a different topic than custom collections, so it would justify a separate documentation page, on the other hand, being familiar with the architecture definitely helps to understand how to write extension methods, so a lot of contents from this page is relevant.

I think that makes another argument to split this page in two: one part about the architecture of the collections, and another part about the implementation of custom collections. Then we could have a third page documenting how to write extension methods (and the first page would be a prerequisite for both implementing custom collections and extension methods). What do you think?

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+so:
+
+~~~ scala
+trait IterableOps[+A, +CC[_], +C] {
+  def filter(p: A => Boolean): C = …
+  def map[B](f: A => B): CC[B] = …
+}
+~~~
+
+And then leaf collection types appropriately instantiate the type
```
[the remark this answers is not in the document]
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+// from IterableOps
+def map[B](f: ((K, V)) => B): Iterable[B]
+~~~
+
+At use-site, if you call the `map` operation, the overloading resolution rules
+first try the definition that comes from `MapOps` (because `MapOps` is more
+specific than `IterableOps`), which returns a `Map`, but if it doesn’t type
+check (in case the return type of the function passed to the `map` call is not
+a pair), it fallbacks to the definition from `IterableOps`, which returns an
+`Iterable`. This is how we follow the “same-result-type” principle: wherever
```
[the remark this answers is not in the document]
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+  kind     |  not sorted   |  sorted
+-----------|---------------|----------------
+`CC[_]`    | `IterableOps` | `SortedSetOps`
+`CC[_, _]` | `MapOps`      | `SortedMapOps`
+
+Here is a diagram illustrating the architecture:
+
+![]({{ site.baseurl }}/resources/images/collections-architecture.svg)
+
+Template traits are in grey whereas collection types are in white.
```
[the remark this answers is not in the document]
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+bases and prefix maps implemented with Patricia tries.
+
+Say you want to create an immutable collection containing *at most* `n` elements:
+if more elements are added then the first elements are removed.
+
+The first task is to find the supertype of our collection: is it
+`Seq`, `Set`, `Map` or just `Iterable`? In our case, it is tempting
+to choose `Seq` because our collection can contain duplicates and
+iteration order is determined by insertion order. However, some
+properties of [`Seq`](/overviews/collections/seqs.html) are not satisfied:
```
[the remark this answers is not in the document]
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
```diff
@@ -0,0 +1,1414 @@
...
+scala> res4.take(3)
+res5: collection.immutable.Iterable[Int] = List(3, 4, 5)
+~~~
+
+You can see that if we try to grow the collection with more than four
+elements, the first elements are dropped (see `res4`). The operations
+behave as expected except for the last one: after calling `take` we
+get back a `List` instead of the expected `Capped1` collection. This
+is because all that was done in class
+[`Capped1`](#first-version-of-capped-class) was making `Capped1` extend
```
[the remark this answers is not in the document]
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
julienrf: Done.

_overviews/core/architecture-of-scala-213-collections.md
julienrf: Done.

julienrf:
I’ve done the splitting in the last commit I’ve pushed.

I’ve kept the Git history to make the review process easier (hopefully…).

Here is how the index looks like now:

overviews

PR #1086, opened 2018-06-17

bogamedia:
You left out the prompt that asks for a file name for the Scala worksheet.


SethTisue:
thank you for the correction!

PR #1089, opened 2018-06-20
Update for page for-comprehensions in Scala Tour

ajaysunilsharma:
Page mentions guard filters out users who are in their 20s but the guard is actually filtering out those users who are not in their 20s.


SethTisue:
thank you for the correction!

PR #1091, opened 2018-06-22
Indentation examples are basically the same.

bogamedia:
004_selection

The examples given for right and wrong indentation are exactly the same. Which one is which?


NthPortal:
The wrong indentation uses 4 spaces, and the right one uses 2 spaces

heathermiller:
Thanks @NthPortal! Indeed... In Scala, the convention is indenting with 2 spaces instead of 4.

PR #1092, opened 2018-06-25
Update favicon and add it for various platforms

Philippus:
see also https://github.com/scala/scala-lang/pull/917


jvican:
This fell through the cracks, I'm so sorry @Philippus. Thank you for the time invested in this. Merging.

Issue #1049, opened 2018-04-02
Add algolia search engine to the website


jvican:
I've submitted a request to crawl all the Scala official website to provide search capabilities to Algolia via https://community.algolia.com/docsearch/.
  
I have the same service in Bloop and I have to say it's excellent. https://scalacenter.github.io/bloop/
  
However, I have no idea about web, and I need the help of someone who would submit a PR trying to style and place the search box that Algolia will provide us. If you know how to do this, your help would be much appreciated.

franktominc:
Can I pick this up?

jvican:
Yes @franktominc! This is the installation instructions I got via email:
  
```

Congratulations, your search is now ready!
I've successfully configured the underlying crawler and it will now run every 24h. We will handle webpages that have the same layout than http://docs.scala-lang.org/tour/tour-of-scala.html

You're now a few steps away from having it working on your website:

- Copy the following CSS/JS snippets and add them to your page

<!-- at the end of the HEAD -->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />

<!-- at the end of the BODY -->
<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
<script type="text/javascript"> docsearch({
apiKey: 'fbc439670f5d4e3730cdcb715c359391',
indexName: 'scala-lang',
inputSelector: '### REPLACE ME ####',
algoliaOptions: { 'facetFilters': ["language:$LANGUAGE"] },
debug: false // Set debug to true if you want to inspect the dropdown
});
</script>

- Add a search input in your page if you don't have any yet. Then update the inputSelector value in JS snippet to a CSS selector that targets your search input field.

- Replace $LANGUAGE with the language you want to search on.
The list of possible language is hardcoded in the config.
So as of today you have: ba, en, es, ja, ko, pl, ru, zh-cn

For example if you want to refine the search to the language "ba" just specify: 'facetFilters': ["language:ba"]

- Optionally customize the look and feel by following the DocSearch documentation (https://community.algolia.com/docsearch/documentation/)
- You can also check your configuration in our github repo (https://github.com/algolia/docsearch-configs/blob/master/configs/scala-lang.json).

Please open a pull request if want to leverage your configuration!

Feel free to get back to us if you have any issues or questions regarding the integration.

[...]

PR #1096, opened 2018-06-29
Add Algolia search for documentation. Need test in prod.

kiviok:
i'm just trying.)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -95,6 +95,17 @@ $(document).ready(function() {
 });
 </script>
 {% endif %}
+
+<!-- Alogolia search for doc -->
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript"> docsearch({
+apiKey: 'fbc439670f5d4e3730cdcb715c359391',
+indexName: 'scala-lang',
+inputSelector: '#doc-search-bar',
+algoliaOptions: { 'facetFilters': ["language:en"] },
+debug: false // Set debug to true if you want to inspect the dropdown
+});
+</script>
 </body>
 
 </html>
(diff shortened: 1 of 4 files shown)


jvican:
Looks great! It seems that there are still no results coming from Algolia, some will at some point. In #1049 we discussed that the best place to add the search box is likely in the first menu bar (option 2 in the linked ticket). Could you move it there? I think it should be possible to add it there, but you may need to put it in https://github.com/scala/scala-lang as this repo only covers docs (tickets are spread out across both repos -- sorry for the confusion).

kiviok:
i think option2 is not good edition. because then we will break main site layout by width.
header contains enough menu items to bring them closer together.

s-pace:
The no result issue should be fixed. We needed an update of the configuration.

Attributes needed the "language" attribute to be set to "en" Solved by aae9cb6f81cde95af34b0d40d211359b4151e964

jvican:
As a first step, I think the current approach (option 3) is enough, but I'd love to see the search box be used not only for docs but for the whole website, and that's why I think option 1 or 2 need to be implemented at some point so that users coming to the website can quickly type things in the search box.
  
Do you have a suggestion on what's the best place to add the search box so that it's used for both docs and blog posts/guides/other data? @kiviok 
  
Also, what are your thoughts @heathermiller ?

kiviok:
First of all, all these pages need to be done in the same style. At least layout width. And then consider some option. Option2 then may be the better.

jvican:
I'm fine merging this soon, but before that @kiviok could you modify the PR to be responsive? Here's a screenshot of the current state:
  
screenshot-2018-7-7 documentation scala documentation

kiviok:
I have something like this now.

jvican:
Excellent, let's give this a try!

jvican:
By the way, thank you for this contribution @kiviok :smile:

SethTisue:
really happy to have this, think you @kiviok

kiviok:
@jvican @SethTisue you are welcome😉
this was my first PR) I really like scala, but at times there are difficulties. With jekyll I have a better situation😆
when ScalaDays arrive in Moscow? 🧐

SethTisue:
> when ScalaDays arrive in Moscow?

heh, that's a tough one, I don't know. that decision normally gets made by the business folks. probably a hard sell since https://www.meetup.com/Scala-Moscow/events/past/ doesn't look too active

PR #1098, opened 2018-07-03
Added contributors widget

### LINKED ISSUE #44: "Contributed by" widget, recognize contributors on each page
It'd be nice to have a list of people (as github user icons) under a subsection called "contributed by" on each page.

sake92:
Fixes #44 

Following [this answer](https://stackoverflow.com/a/19200303/4496364) I managed to get this via Github API:

I roughly followed Mozilla's [docs page](https://developer.mozilla.org/en-US/docs/Web/HTML) design and their contributors listing.
Some users don't have the "author" field. Probably deleted their profile?  
So, the fallback is `commit.commit.committer.name` which is, I suppose, always present because of Git...  

Also, if the author isn't present, image icon is generated with https://github.com/identicons. Example here is Ghildiyal user.

Currently, I added this to tour only. IDK did you guys intend to put these on each page?  
If so, there could be some problems with getting page URL from Github. IDK if all of them follow same pattern as "tours"..

### DIFF
--- a/resources/js/functions.js
+++ b/resources/js/functions.js
@@ -522,3 +522,57 @@ $(document).ready(function(){
         return false; 
     }); 
 });
+
+//Contributors widget
+// see https://stackoverflow.com/a/19200303/4496364
+$(document).ready(function () {
+  let githubApiUrl = 'https://api.github.com/repos/scala/docs.scala-lang/commits';
+  // transform "/tour/basics.html" to "_ba/tour/basics.md";
+  let thisPageUrl = window.location.pathname;
+  thisPageUrl = thisPageUrl.substring(1, thisPageUrl.lastIndexOf('.'));
+  thisPageUrl = '_' + thisPageUrl + '.md';
+  // e.g. https://api.github.com/repos/scala/docs.scala-lang/commits?path=README
+  let url = githubApiUrl + '?path=' + thisPageUrl;
+  $.get(url, function (data, status) {
+    
+    let contributorsUnique = [];
+    let res = data.forEach(commit => {
+      // add if not already in array
+      let addedToList = contributorsUnique.find(c => {
+        let matches = c.authorName == commit.commit.committer.name;
+        if (!matches && commit.author) {
+          matches = c.authorName == commit.author.login;
+        }
+        return matches;
+      });
+
+      if (!addedToList) {
+        // first set fallback properties
+        let authorName = commit.commit.committer.name;
+        let authorLink = 'mailto:' + commit.commit.committer.email;
+        let authorImageLink = 'https://github.com/identicons/' + commit.commit.committer.name + '.png';
+        // if author present, fill these preferably
+        if (commit.author) {
+          authorName = commit.author.login;
+          authorLink = commit.author.html_url;
+          authorImageLink = commit.author.avatar_url;
+        }
+        contributorsUnique.push({
+          'authorName': authorName,
+          'authorLink': authorLink,
+          'authorImageLink': authorImageLink
+        });
+      }
+    });
+
+    let contributorsHtml = '';
+    contributorsUnique.forEach(contributor => {
+      let contributorHtml = '<div>';
+      contributorHtml += '<img src="' + contributor.authorImageLink + '">';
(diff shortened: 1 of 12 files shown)


resources/js/functions.js
```diff
@@ -521,4 +521,58 @@ $(document).ready(function(){
...
+        if (!matches && commit.author) {
+          matches = c.authorName == commit.author.login;
+        }
+        return matches;
+      });
+
+      if (!addedToList) {
+        // first set fallback properties
+        let authorName = commit.commit.committer.name;
+        let authorLink = 'mailto:' + commit.commit.committer.email;
```
martijnhoekstra: Is it really a good idea to have commiter mailto links in a href? This feels unconfortable to me.

sake92: I agree that's sensitive information. We can put just "#" then?  
Although, users should be aware that their public commits must have an email always.  
You can have multiple emails associated with Github account.

martijnhoekstra: Alternatively, a link to the github account? Is that available at this point?

jvican: @SethTisue Do we have a github token that Sakib can use for the authentication?
  
@sake92 In the meanwhile, implementing conditional requests make sense. Even if we have authentication, we'll have a limit of 5000 requests per hour. I suspect we'll also hit that.
  
Ideally, we would like to persist the author information in the same repo. Is there a way to do that?

martijnhoekstra: Ah, now I see! If it's available, you take the author link, if not, the committer email of the commit - that's probably a much rarer problem.

If we do take an email, that should probably be the author data, not the committer data (commit.commit.author.name and commit.commit.author.email).

That said, I still don't think we should be providing mailto links. That's what I want in roughly 0% of situations, and I suspect that's the same for others.

My personal preferences, in decreasing order of preferentiality in case we only have an email address:

* A placeholder link: an `a` element with no `href` attribute
* A fragment link: An `a` element with a `#` href 
* No `a` element at all

@jvican wdyt?

jvican: sounds good to me!

sake92: @martijnhoekstra Fixed both issues:
- removed `mailto:`, now it's just `<a>authorName</a>`
- using `commit.commit.author.name` as fallback instead of `commit.commit.committer.name`

@jvican Regarding caching, I got tricked by Chrome devtools... xD I had `Disable cache` turned on, so it didn't do caching. Browser does a great job for us, sends headers automatically (you can check rate limiting stays same in `X-RateLimit-Remaining` header).

jvican: I'm a website newbie, so I have no idea how these headers work, but do they work across different clients or is the caching happening only per browser? I'm asking this because I would suspect we overpass the rate limit if we get, say, 5001 users visiting the docs.
  
Wouldn't there be an option to do this on the backend side instead of the frontend side? A tool we could use to persist this information in the same markdown, or even in the templates generated by Jekyll?

sake92: This is only browser cache, local. Github says that 60 req/hour is **per IP adress**, so every user would be able to **view 60 pages per hour**. Hmm.. we could probably live with this?  

If we want to make use of OAuth we must have a backend and a database to store these results.  
And then we'd just use jQuery or whatever to fetch these...

sake92:
@jvican I refactored JS logic to support more pages. The problem is that pages don't follow a consistent pattern. E.g. FAQ is rendered as `/tutorials/FAQ/context-bounds.html` but located in `_overviews/FAQ` and similar. It works on most of the pages (more than 95% where widget is included), it just has problem in those corner cases.
Script just removes the `div` if the request doesn't succeed (pages doesn't exist or something like that).

Also, I didn't put the widget on some general index pages and didn't include it in pages like books, SIPs, downloads, community etc.

jvican:
I just cloned the repo and check that the layout was correct for small devices. Thank you for this contribution @sake92, it's really good.

screenshot-2018-7-7 introduction scala documentation 1
screenshot-2018-7-7 introduction scala documentation

SethTisue:
this is really cool ❤️

PR #1102, opened 2018-07-16
Update documentation since deprecation now takes two arguments

jatinguptarally:


jatinguptarally:
thanks @SethTisue, let me know if this looks good

PR #1103, opened 2018-07-18
Use jQuery from CDN, fall back to local if needed

### LINKED ISSUE #1074: Chinese users can't get to jquery
Chinese users can't get to jquery。
 https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js
Please use the local server to store the jQuery file

Philippus:
Implemented the change suggested by @NthPortal.


NthPortal:
Thanks so much for doing this!!

_includes/footer.html
```diff
@@ -21,7 +21,8 @@
 </footer>
 
 <!-- jquery -->
-<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.1/jquery.min.js" integrity="sha384-3ceskX3iaEnIogmQchP8opvBy3Mi7Ce34nWjpBIwVTHfGYWQS9jwHDVRnpKKHJg7" crossorigin="anonymous"></script>
+<script>(window.jQuery) || document.write('<script src="/scripts/jquery.3.1.1.js"><\/script>');</script>
```
NthPortal: The only thing I'm wondering is if this needs to be `src="{{ site.baseurl }}/scripts/jquery.3.1.1.js"`, but I don't really know enough about jekyll to answer that for sure

Philippus: I've added it.

PR #1111, opened 2018-07-27
Updated the docs for Multiple Parameter Listing

technical-harsh90:
Here i have updated the documentation with the following,

- Another type of statement used to denote **foldLeft**,
- Another type of statement used to denote **foldRight**
- All generic statements used for fold family

### DIFF
--- a/_tour/multiple-parameter-lists.md
+++ b/_tour/multiple-parameter-lists.md
@@ -45,8 +45,22 @@ numbers.foldLeft(0, {(m: Int, n: Int) => m + n})
 ```
 numbers.foldLeft(0)(_ + _)
 ```
-    
-   Also, it allows us to fix the parameter `z` and pass around a partial function and reuse it as shown below:
+
+   Finally, `foldLeft` and `foldRight` can be used in any of the following terms,
+```tut
+val numbers = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+numbers.foldLeft(0)((sum, item) => sum + item) // Generic Form
+numbers.foldRight(0)((sum, item) => sum + item) // Generic Form
+
+numbers.foldLeft(0)(_+_) // Curried Form
+numbers.foldRight(0)(_+_) // Curried Form
+
+(0 /: numbers)(_+_) // Used in place of foldLeft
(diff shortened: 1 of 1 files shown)


_tour/multiple-parameter-lists.md
```diff
@@ -45,8 +45,22 @@ numbers.foldLeft(0, {(m: Int, n: Int) => m + n})
...
+numbers.foldRight(0)((sum, item) => sum + item) // Generic Form
+
+numbers.foldLeft(0)(_+_) // Curried Form
+numbers.foldRight(0)(_+_) // Curried Form
+
+(0 /: numbers)(_+_) // Used in place of foldLeft
+(numbers :\ 0)(_+_) // Used in place of foldRight
+```   
+
+   Above statement `numbers.foldLeft(0)(_ + _)` allows us to fix the parameter `z` and pass around a partial function and reuse it as shown below:
```
jvican: Don't you mean here "the statement below"?

technical-harsh90:
@jvican Thanks for the review on the PR, Yes it should have been Below statement, but to make it more clear to the users and readers, i have changed the order of my changes in the PR just a bit.

Have a look at it, it will suit now so that you can merge it.

SethTisue:
I'm not convinced these are good changes.

`numbers.foldLeft(0)(_ + _)` is an expression, not a statement.

We should not be showing new users `/:` and `:\` at all. They are not recommended for use.

Even if they were recommended, showing these alternate forms would not be relevant to the topic of this tour page, which is multiple parameter lists.

For the same reason, I don't see how showing both the underscore-based and non-underscore-based forms of calling the folds constitutes an improvement in this context. It has nothing to do with multiple parameter lists.

Even if we agree that both of those versions should be included, it is incorrect to describe underscore-based form as "curried". Using `_` or not is orthogonal to whether something is "curried".  (And anyway, it's almost always better to avoid the word "curried" in a Scala context, unless we're specifically talking about changing a signature from `(A, B) => C` to `A => B => C`... which is rare in Scala.)

(While I was reviewing this, I happened to notice that the use of "partial function" is incorrect. I think "partially applied function" was meant. The mistake wasn't introduced in this PR, though.)

SethTisue:
I've submitted my suggested changes at #1288

PR #1112, opened 2018-07-31
Fix typo in views.md

flash-gordon:
👋


SethTisue:
opened #1113 on the htmlproofer errors

Issue #833, opened 2017-07-25
Where to put extension methods


MasseGuillaume:
http://docs.scala-lang.org/overviews/core/value-classes.html#extension-methods

This section should show the two available options:

```scala
package object foo {
  implicit class RichInt(val self: Int) extends AnyVal {
    def toHexString: String = java.lang.Integer.toHexString(self)
  }
}
```

or

```scala
object Foo {
  implicit class RichInt(val self: Int) extends AnyVal {
    def toHexString: String = java.lang.Integer.toHexString(self)
  }
}
```

SethTisue:
pull request?

jan0sch:
PR created: https://github.com/scala/docs.scala-lang/pull/1117

PR #1117, opened 2018-08-01
#833 Where to put extension methods

jan0sch:
Add a short explanation and two examples.

### DIFF
--- a/_overviews/core/value-classes.md
+++ b/_overviews/core/value-classes.md
@@ -56,6 +56,46 @@ The following fragment of `RichInt` shows how it extends `Int` to allow the expr
 At runtime, this expression `3.toHexString` is optimised to the equivalent of a method call on a static object
 (`RichInt$.MODULE$.extension$toHexString(3)`), rather than a method call on a newly instantiated object.
 
+### Location of extension methods
+
+There are several options regarding where to place the implemented extension methods.
+They can be placed in a package object which makes them available within the package automatically.
+
+    package object mypackage {
+      implicit class RichInt(val self: Int) extends AnyVal {
+        def toHexString: String = java.lang.Integer.toHexString(self)
+      }
+    }
+
+If the extension methods shall be reused it might be feasible to implement them not implicitly and include them via traits.
+
+    // in file: src/main/scala/mypackage/package.scala
+    package object mypackage extends mypackage.Shared {
+    }
+
+    // in file: src/main/scala/mypackage/Shared.scala
+    trait Shared {
+      implicit def toRichInt(self: Int): RichInt = new RichInt(self)
+    }
+
+    class RichInt(val self: Int) extends AnyVal {
+      def toHexString: String = java.lang.Integer.toHexString(self)
+    }
+
+However automatically available implicits have their drawbacks so it might be better to place them within a regular object.
+
+    object implicits {
+      implicit class RichInt(val self: Int) extends AnyVal {
+        def toHexString: String = java.lang.Integer.toHexString(self)
+      }
+    }
+
+To make them available an import is necessary like this:
+
+    import org.example.mypackage.implicits._
+
+    val h = 3.toHexString
+
 ## Correctness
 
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
There is another trick I would add:

```scala
// in file: org/package.scala
package org
package object example extends example.PackageShared {
  // ...
}

// in file: org/example/PackageShared.scala
trait PackageShared {
  implicit def toRichInt(self: Int): RichInt = new RichInt(self)
}

class RichInt(val self: Int) extends AnyVal {
  def toHexString: String = java.lang.Integer.toHexString(self)
}
```

This allows to reuse all extension methods. (for example, you could add more in the package object of org.example.extra).

jan0sch:
Included.

_overviews/core/value-classes.md
```diff
@@ -56,6 +56,44 @@ The following fragment of `RichInt` shows how it extends `Int` to allow the expr
...
+        def toHexString: String = java.lang.Integer.toHexString(self)
+      }
+    }
+
+If the extension methods shall be reused it might be feasible to implement them not implicitly and include them via traits.
+
+    package object mypackage extends mypackage.Shared {
+    }
+
+    trait Shared {
```
MasseGuillaume: This should be on a separate file, and in `package mypackage`.

jan0sch:
Is it more clear now?

_overviews/core/value-classes.md
```diff
@@ -56,6 +56,46 @@ The following fragment of `RichInt` shows how it extends `Int` to allow the expr
...
+
+    package object mypackage {
+      implicit class RichInt(val self: Int) extends AnyVal {
+        def toHexString: String = java.lang.Integer.toHexString(self)
+      }
+    }
+
+If the extension methods shall be reused it might be feasible to implement them not implicitly and include them via traits.
+
+    // in file: src/main/scala/mypackage/mypackage.scala
```
MasseGuillaume: by convention this is `// in file: src/main/scala/mypackage/package.scala`

jan0sch: Ups. Correcting.

SethTisue:
don't extension methods most commonly go in the companion object?

MasseGuillaume:
@SethTisue yes, this is an advanced use case.

_overviews/core/value-classes.md
```diff
@@ -56,6 +56,46 @@ The following fragment of `RichInt` shows how it extends `Int` to allow the expr
...
+      implicit def toRichInt(self: Int): RichInt = new RichInt(self)
+    }
+
+    class RichInt(val self: Int) extends AnyVal {
+      def toHexString: String = java.lang.Integer.toHexString(self)
+    }
+
+However automatically available implicits have their drawbacks so it might be better to place them within a regular object.
+
+    object implicits {
```
jvican: The convention is to use `syntax` here instead of `implicits`, I think.

tpolecat:
@jvican my approach is probably a little over-engineered for the typical user. I put everything in a `syntax` package and have a [non-implicit] class for each hunk of syntax, a trait that defines the implicit conversion, an object that extends the trait (`import woozle.syntax.list._`), and an `all` object that extends all the conversions (`import woozle.syntax.all._`). The advantage is that it's very clear what's going on. Disadvantage is that it's probably needlessly elaborate for common cases.

Example syntax package [here](https://github.com/tpolecat/ocs3/tree/develop/modules/core/shared/src/main/scala/gem/syntax).

_overviews/core/value-classes.md
jan0sch: I've seen both but I think `syntax` might indeed make more sense here.

SethTisue: > One thing I don't understand is why this is placed in the value-classes.md guide

agree, this would be much better placed in the style guide

jan0sch: Well it started out as adding some stuff regarding value classes. ;-)

SethTisue:
closing, but a new PR in this general area, reflecting the feedback above and on Discourse, would be welcome

PR #1133, opened 2018-08-15
Update for-comprehensions.md

saadel:
Removed the unnecessary case class modifiers: 
`val` for the arguments and `new` when instantiating it


Philippus:
Could you make the same change in the `ba` language tour?

Issue #1137, opened 2018-08-27
documentation on ForkJoinPool should mention shutdown()


ktegan:
The threads in the ForkJoinPool are not released unless you call some flavor of shutdown(). 
 The code sample in https://docs.scala-lang.org/overviews/parallel-collections/configuration.html is essentially leaking resources.  This issue was discussed here as well: https://github.com/scala/bug/issues/11097 .

SethTisue:
the way this works on GitHub is, you fork the repo, you push the change you want to a branch in your fork, and then you open a pull request from that branch

(if you'd rather not bother with the usual process, I can apply your diffs for you this time, just let me know.)

PR #1138, opened 2018-08-27
Call ForkJoinPool.shutdown(), use new package location, PR for #1137

ktegan:
- If you want to repeatedly create new instances of ForkJoinPool
   you must call some flavor of shutdown() on the pool.  This
   adds the command to the example and gives a note after the
   example.

 - Use java.util.concurrent.ForkJoinPool as the package for
   ForkJoinPool as scala.concurrent.forkjoin.ForkJoinPool is
   deprecated


SethTisue:
review by @viktorklang

PR #1141, opened 2018-09-07
Update upper-type-bounds.md

f0ster:
Char with for html container is smaller than line length, can't see comment on website 
https://docs.scala-lang.org/tour/upper-type-bounds.html


SethTisue:
thank you!

PR #1142, opened 2018-09-11
Link directly to referenced document

vaer-k:
Link directly to referenced document, "Scala By Example".


SethTisue:
while this change is certainly a strict improvement and therefore mergeable, Scala by Example hasn't been updated since 2014 and therefore we probably shouldn't be mentioning it or linking to it anymore at all.

Perhaps the wording here predates the creation of the "Tour of Scala" (https://docs.scala-lang.org/tour/tour-of-scala.html), and it might be better to link to that instead...?

vaer-k:
@SethTisue That makes sense. I went ahead and updated PR to reflect your suggestion.

SethTisue:
thank you!

refined a bit further in 70926b24852f3d0a92c9845e6f7b2cf530f36f24

PR #1143, opened 2018-09-11
Improve JDK 9/10 compatibility notes visibility

Jacoby6000:
For somebody just glancing at this doc, it's way too easy to miss the note at the bottom. This adds the note in the table and makes it easier to see.

### DIFF
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -15,12 +15,14 @@ And as Scala and the JVM improve independently over time, Scala may drop compati
 
 This table shows the first Scala release in each series that works with each JVM release.
 
-| JVM version | Minimum Scala versions                               |
-|:-----------:|:-----------------------------------------------------|
-| 9, 10       | 2.12.4, 2.11.12, 2.10.7                              |
-| 8           | 2.12.0, 2.11.0, 2.10.2                               |
-| 7           | 2.11.0, 2.10.0                                       |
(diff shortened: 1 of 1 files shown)


Jacoby6000:
Also, this makes me wonder.  Does scala 2.10 not also have this problem?  Seems strange that compatibility would decrease between 2.10 and 2.11.

SethTisue:
> Does Scala 2.10 not also have this problem? Seems strange that compatibility would _decrease_ between 2.10 and 2.11

I agree that wording isn't as clear as it should be. There was no decrease in compatibility between 2.10 and 2.11.

(Not enforcing module constraints isn't so much a "problem" as simply an enhancement that hasn't happened yet, IMO.)

Anyway, as far as I can tell, the reason it says "2.11.12 and 2.12.6" is to attempt to imply that the situation may improve in the future in the 2.11.x and 2.12.x series, but it isn't likely to improve in the 2.10.x series.

But actually, it isn't likely to improve any further in the 2.11.x series, either, unless some new breakage on 9/10/11 is discovered. There are no definite plans for any further 2.11.x releases at all, and even if another 2.11.x release does end up happening for some reason, we'd be extremely conservative about what'd we accept for it.

SethTisue:
@Jacoby6000 does that make you want to tweak this any further, or should I merge as-is?

Jacoby6000:
Nah, go ahead and merge it. If more information is revealed to me I'll just
open another PR. No big deal

PR #1144, opened 2018-09-12
#850:Tuples page added to tour

mghildiy:


jvican:
I'd like to review this soon, but it looks like I'm all busy until the end of next week. @mghildiy Could you post to Scala Contributors asking for reviewers for this PR?

mghildiy:
Ok.

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+
+Tuple in Scala is a series of classes: Tuple2, Tuple3, etc., through Tuple22.
+So when we create a tuple with n elements(n lying between 2 and 22), Scala basically instantiates
+one of the corresponding class from the group, parametrized with types of constituent elements.
+For eg., ingredient is of type Tuple2[String, int].
+
+## Accessing the elements
+
+Tuple elements are accessed using underscore syntax.
+tuple._n gives nth element(given there are that many elements).
```
adriaanm: Quote as: `tuple._n`

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+
+```tut
+println(ingredient._1) // Sugar
+
+println(ingredient._2) // 25
+```
+
+## Iterating a tuple
```
adriaanm: Before you get to `productIterator`, which I think is a bit of a corner case, I'd show destructuring using patterns. Right now, you have shown how to do this in local vals, but the same applies to pattern matching (`x match { case (a,b) => ...`), function literals (`{ case (a, b) => a + b }`), and for comprehensions (`for ( (a, b) <- pairs ) ...`).

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+```tut
+ingredient.productIterator.foreach(println)
+```
+Tuple's toString method gives a String representation of its contents.
+
+```tut
+ingredient.toString // (Sugar,25)
+```
+
+Scala tuple also supports destructuring.
```
adriaanm: bump this to a `##` header.

PS: I think it's kind of neat to mention that `()` is conceptually `Tuple0`, or `Unit`. There can only be one value of this type since it has no elements.

Maybe one final comment: you should use case classes whenever the elements have more meaning.

_ko/tour/mixin-class-composition.md
```diff
@@ -10,7 +10,7 @@ num: 6
 language: ko
 
 next-page: higher-order-functions
-previous-page: traits
+previous-page: tuples
 ---
 
 _단일 상속_ 만을 지원하는 여러 언어와는 달리, 스칼라는 더욱 일반화된 시각에서 클래스를 재사용한다. 스칼라는 새로운 클래스를 정의할 때 _클래스의 새로운 멤버 정의_ (즉, 슈퍼클래스와 비교할 때 변경된 부분)를 재사용할 수 있다. 이를 _믹스인 클래스 컴포지션_ 이라고 부른다. 이터레이터를 추상화한 다음 예제를 살펴보자.
```
adriaanm: This should be replaced by the stub to indicate it hasn't been translated yet.

_pt-br/tour/mixin-class-composition.md
```diff
@@ -8,7 +8,7 @@ partof: scala-tour
 
 num: 5
 next-page: higher-order-functions
-previous-page: traits
+previous-page: tuples
 language: pt-br
 ---
 _Nota de tradução: A palavra `mixin` pode ser traduzida como mescla, porém é preferível utilizar a notação original_
```
adriaanm: This should be replaced by the stub to indicate it hasn't been translated yet.

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+
+num: 
+next-page: mixin-class-composition
+previous-page: traits
+topics: tuples
+
+redirect_from: "/tutorials/tour/tuples.html"
+---
+
+In Scala, a tuple is a class that can hold elements of different types.
```
Jasper-M: I'm never sure if you should say "is a class" or "is an object" in this case.

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+Tuples are immutable.
+
+Tuples come handy when we have to return multiple values from a function.
```
Jasper-M: come *in* handy

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+
+A tuple can be created as:
+
+```tut
+val ingredient = ("Sugar" , 25)
+
+println(ingredient.getClass) // class scala.Tuple2
```
Jasper-M: Not sure if we should expose beginners to `.getClass` here.

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+```tut
+val ingredient = ("Sugar" , 25)
+
+println(ingredient.getClass) // class scala.Tuple2 
+```
+This creates a tuple containing a String element and an Int element.
+
+Tuple in Scala is a series of classes: Tuple2, Tuple3, etc., through Tuple22.
+So when we create a tuple with n elements(n lying between 2 and 22), Scala basically instantiates
+one of the corresponding class from the group, parametrized with types of constituent elements.
```
Jasper-M: class -> classes
parametrized -> parameterized

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+For eg., ingredient is of type Tuple2[String, int].
```
Jasper-M: int -> Int
And I think it's better to show the `(String, Int)` syntax to beginners and then explain that it is syntactic sugar for `Tuple2[String, Int]`.

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+Tuple's toString method gives a String representation of its contents.
+
+```tut
+ingredient.toString // (Sugar,25)
+```
+
+Scala tuple also supports destructuring.
+
+```tut
+val(name, quantity) = ingredient
```
Jasper-M: I prefer a space between `val` and `(`.

_tour/tuples.md
adriaanm: It would be more correct to say "is a class whose instances hold elements". I think it's common enough to leave that implicit so as not to confuse people :-)

_tour/tuples.md
SethTisue: agree, the better way to demonstrate that something conforms to a certain type is to show that it still compiles if you add a type ascription such as `(String, Int)` or `Tuple2[String, Int]`

_tour/tuples.md
```diff
@@ -0,0 +1,68 @@
...
+
+```tut
+println(ingredient._1) // Sugar
+
+println(ingredient._2) // 25
+```
+
+## Iterating a tuple
+
+Tuple provides productIterator method to iterate it.
```
SethTisue: I would suggest omitting any mention of `productIterator` (and related methods such as `productElement`) from the tour entirely. They are rarely used, since they don't provide any type-safety.

SethTisue:
thanks for writing this, nice work! I don't expect this will need much more work in order to become mergeable.

of the various changes suggested, I think the single most important one is Adriaan's "I'd show destructuring using patterns..."

_pt-br/tour/mixin-class-composition.md
dcsobral: There _is_ an explanation, but it does not correspond to the English version of the page.

_ko/tour/mixin-class-composition.md
mghildiy: There is some translation there, not sure though if it corresponds to English version.

mghildiy:
This can be merged now.

SethTisue:
@jvican what do you make of the CI failure? I don't have rights to hit "rebuild" in Drone to see if it's transient.

jvican:
Thanks I just restarted the CI

mghildiy:
Please merge it.

_pt-br/tour/mixin-class-composition.md
adriaanm: Ah sorry, I got confused by the fact that this page is actually about mixins, and it's just the "next" pointer that's being updated.

_ko/tour/mixin-class-composition.md
adriaanm: My mistake,  I got confused by the fact that this page is actually about mixins, and it's just the "previous" pointer that's being updated.

PR #1153, opened 2018-09-24
(Minor edit) Curly brace was in the wrong spot in declarations.md

rkoeninger:
Curly brace was in the wrong spot. Double checked in correct spot now using repl.


sjrd:
Er, I suspect you might have misread the paragraph. The style there says that style (1) should be used. According to style (1), the brace was correctly placed.

rkoeninger:
You're right, my mistake.

Even after a page of explanation, I still didn't understand that was even valid code. I thought it was a typo.

PR #1154, opened 2018-09-24
Prevented TOC from capturing unwanted headings

### LINKED ISSUE #1148: Page Contributors shows as sub-heading in ToC
Just noticed something weird about how the table of contents is structured.

The generated "Contributors to this Page" section at the bottom of each page gets shown as a sub-heading in the ToC when it logically isn't:

(Contributors is not a sub-section of Structural Types)

image

Two quick ways of fixing it:

1. Adding `.content-contributors` in the `exclude` list when calling the toc plugin, so it doesn't show up in the ToC, see here:

https://github.com/scala/docs.scala-lang/blob/5ac9c209ec572f1be183ce8e6c55d2f2cd9d9f72/resources/js/functions.js#L298-L309

2. Or make the contributors heading an `h2` so it's on the same level as the previous heading and won't get nested under it:

https://github.com/scala/docs.scala-lang/blob/5ac9c209ec572f1be183ce8e6c55d2f2cd9d9f72/_includes/contributors-list.html#L1-L4

We can decide the preferred approach and I'll open a PR with the fix.

rkoeninger:
Fixes #1148 

Used `toc-context` class to identify where TOC should look for headings. Code was using `inner-box`, but `inner-box` divs often contain content not wanted on the TOC (e.g. the contributors list).

Nested `toc-context` div doesn't add padding or margin or appear to change the layout in any way.

Checked all the pages, comparing them to the currently live site to make sure everything still works.

### DIFF
--- a/_includes/inner-page-main-content.html
+++ b/_includes/inner-page-main-content.html
@@ -2,7 +2,9 @@
 	<div class="wrap">
 		<div class="content-primary">
 			<div class="inner-box">
-				{{content}}
+				<div class="toc-context">
+					{{content}}
+				</div>
 
 				{% include contributors-list.html %}
 			</div>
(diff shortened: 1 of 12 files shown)


SethTisue:
awesome, thank you!

PR #1161, opened 2018-10-01

emonmishra:
Remove line that were confusing for Java dev. 
Explanation:
Since Scala doesn't have primitive types and as it appears those lines compared 'Int' of Scala with 'Integer' wrapper class of Java then 'Integer' class in Java do extend to Object class. Hence 'Any' is not 'More' general version of 'Object' in this context. Pls let me know if I am missing something.


SethTisue:
I can see that the original wording is potentially confusing and could be improved, but I don't think the proposed change is actually an improvement.

`Any` is more general than `Object` because it includes types that `Object`/`AnyRef` doesn't include. It encompasses both reference types and primitive types. There needs to be *some* explanation of this here, since it's *the* reason why `Any` even exists at all.

PR #1163, opened 2018-10-01
Tweak wording in tour of extractors

som-snytt:
Someone tripped over this on stack overflow.

There is confusion because extractor is used in defining a value, not in an assignment. This feature is weird enough to warrant a couple of extra words, and maybe avoid a trip to SO.


SethTisue:
thanks Som!

PR #1165, opened 2018-10-05
Made Compound Types page refer to Class Composition with Mixins

### LINKED ISSUE #1128: Compound Types doc page should point to Class Composition Mixin page for refinement explanation
Currently the last line of the `Compound Types` page in the `Tour of Scala` reads:

> An example for the use of refinements is given on the page about [abstract types](https://docs.scala-lang.org/tour/abstract-types.html).

However refinements are not mentioned at all in the `Abstract Types` page. The concept is introduced, although the word `refinements` is not used, in the `Class Composition with Mixins` page. So maybe that last line should instead be:

> An example for the use of refinements is given on the page about [class composition with mixins](https://docs.scala-lang.org/tour/mixin-class-composition.html).

rkoeninger:
Fixes #1128

As suggested by @iserraller.

### DIFF
--- a/_tour/compound-types.md
+++ b/_tour/compound-types.md
@@ -51,4 +51,4 @@ def cloneAndReset(obj: Cloneable with Resetable): Cloneable = {
 Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
 The general form is: `A with B with C ... { refinement }`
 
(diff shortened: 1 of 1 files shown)


iserraller:
Looks good! My apologies for being too lazy to do this myself :-(

SethTisue:
thank you!

PR #1175, opened 2018-10-11
zh-cn for Scala Tour:Abstract-types.md

cunchen:
Simplified Chinese version of Scala Tour:Abstract Type

### DIFF
--- a/_zh-cn/tour/abstract-types.md
+++ b/_zh-cn/tour/abstract-types.md
@@ -29,8 +29,7 @@ abstract class SeqBuffer extends Buffer {
   def length = element.length
 }
 ```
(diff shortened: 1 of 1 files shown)


_zh-cn/tour/abstract-types.md
```diff
@@ -13,3 +13,66 @@ language: zh-cn
 next-page: compound-types
 previous-page: inner-classes
 ---
+特征(Trait)和抽象类(abstract class)可以包含抽象类型成员。这表示具体的实现可以在具体类型中。示例如下：
```
dongxuwang: > Traits and abstract classes can have an abstract type member. This means that the concrete implementations define the actual type.
特征和抽象类可以包含一个抽象类型成员，意味着实际类型可由具体实现来确定。

PR #1177, opened 2018-10-12
zh-cn for Scala Toor: Pattern matching

declan94:
Simplified Chinese version for Scala Toor: Pattern matching

### DIFF
--- a/_zh-cn/tour/pattern-matching.md
+++ b/_zh-cn/tour/pattern-matching.md
@@ -31,9 +31,9 @@ x match {
   case _ => "many"
 }
 ```
-The `val x` above is a random integer between 0 and 10. `x` becomes the left operand of the `match` operator and on the right is an expression with four cases. The last case `_` is a "catch all" case for any number greater than 2. Cases are also called _alternatives_.
+上述代码中的`val x`是一个0到10之间的随机整数，将它放在`match`运算符的左侧对其进行模式匹配，`match`的右侧是包含4条`case`的表达式，其中最后一个`case _`表示匹配其余所有情况，在这里即是`x`大于2的情况。
 
-Match expressions have a value.
+`match`表达式具有一个结果值
 ```tut
 def matchTest(x: Int): String = x match {
   case 1 => "one"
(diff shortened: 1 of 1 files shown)


_zh-cn/tour/pattern-matching.md
```diff
@@ -13,3 +13,140 @@ language: zh-cn
...
+val x: Int = Random.nextInt(10)
+
+x match {
+  case 0 => "zero"
+  case 1 => "one"
+  case 2 => "two"
+  case _ => "many"
+}
+```
+上述代码中的`val x`是一个0到10之间的随机整数，将它放在`match`运算符的左侧对其进行模式匹配，`match`的右侧是包含4条`case`的表达式，其中最后一个`case _`表示匹配其余所有情况，在这里即是`x`大于2的情况。
```
dongxuwang: > Cases are also called _alternatives_.
情形（cases）也叫备选（alternatives）

declan94: I think this is only about different English terms and doesn't make sense in Chinese. What do you think?

dongxuwang: Agree, so that's why you ignored it.

declan94: Yes indeed.

_zh-cn/tour/pattern-matching.md
```diff
@@ -13,3 +13,140 @@ language: zh-cn
...
+
+case class Email(sender: String, title: String, body: String) extends Notification
+
+case class SMS(caller: String, message: String) extends Notification
+
+case class VoiceRecording(contactName: String, link: String) extends Notification
+
+
+```
+`Notification` is an abstract super class which has three concrete Notification types implemented with case classes `Email`, `SMS`, and `VoiceRecording`. Now we can do pattern matching on these case classes:
```
dongxuwang: Need to be translated as well

SethTisue:
@declan94 want to try to finish this up...?

_zh-cn/tour/pattern-matching.md
declan94: Thank you. Translated now

declan94:
> @declan94 want to try to finish this up...?

Thank you. Have finished!

SethTisue:
thank you!

Issue #1109, opened 2018-07-22
Last example of Mixin class composition tour uses App trait without any explanation


jrnos:
The last example on page https://docs.scala-lang.org/tour/mixin-class-composition.html defines an object StringIteratorTest extending App i.e. creates an executable using App trait. However, that concept has not been mentioned earlier in the tour neither is explained here. Shouldn't this concept be explained shortly on the same page (or add a link to https://www.scala-lang.org/api/current/scala/App.html)?

jrnos:
The problematic example referring to App trait has been removed in commit dde4a8d32f0996f5e4c67bfc84f98f710f0905e4 so I'll close this issue as resolved.

PR #1186, opened 2018-10-18
1109 trait doc

jijosg:
#1109  Added example to extend App trait under usages of traits.


SethTisue:
I have doubts about whether this is a good addition. The job of the traits page should be to convey the essence of what a trait is. `App` seems outside of these bounds to me, plus `App` is a highly unusual trait, because it extends `DelayedInit`.

If it's important to cover `App` in the language tour, perhaps a new section would be appropriate? Not sure.

PR #1193, opened 2018-10-24
zh-cn for Scala Tour: implicit-parameters.md

realwunan:
Simplified Chinese version of Scala Tour: Implicit Parameters


liufengyun:
Thanks for your contribution @realwunan 🎉

PR #1200, opened 2018-10-31
Don't mention `scala.Predef.Pair` in the Tour

Duhemm:
`Pair` is deprecated since 2.11 and removed in 2.13.


martijnhoekstra:
Nice!

PR #1203, opened 2018-11-01
Add package objects docs [tutorial] {EN}

zafarella:
* sources:
 - https://www.scala-lang.org/docu/files/packageobjects/packageobjects.html
 - https://stackoverflow.com/questions/3400734/package-objects
 - https://github.com/scala/scala/blob/2.13.x/src/library/scala/package.scala


SethTisue:
thank you! this is a very nice addition to have.

SethTisue:
I suggested some edits at https://github.com/scala/docs.scala-lang/pull/1236

PR #1204, opened 2018-11-02
Simplify Chinese translation of Scala Tour: compound-types.md

realwunan:
Simplify Chinese translation of Scala Tour: Compound Types


realwunan:
Fix the build by rebase

PR #1215, opened 2018-11-08
Incorrect use of `if`, should be `filter` instead

satishpanduranga:
In the following sentence , `filter` should be used instead of `if`

This caused many problems, because people expected the `if` to be considered

### DIFF
--- a/_overviews/FAQ/yield.md
+++ b/_overviews/FAQ/yield.md
@@ -127,7 +127,7 @@ is the equivalent for-comprehension code:
     for (x <- Stream.range(1, 10); if x % 2 == 1 && !found)
       if (x == 5) found = true else println(x)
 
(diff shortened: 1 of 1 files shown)


SethTisue:
I don't understand why the old way was incorrect...?

satishpanduranga:
@SethTisue The previous paragraph that introduces this concept talks about the problem as being 
"The difference happens because **filter** is immediately applied with `List`,"
But the expectation of the people was that `filter` to be considered on-demand.

The change requested in this Pull requests is intended to change the language of the paragraph to be more clearer as to what the intention of introducing `withFilter` is about.

SethTisue:
I'm sorry, but I fail to see that this is an improvement.

Anyone else watching the repo have an opinion about it?

PR #1218, opened 2018-11-15
Update variances.md

hsmak:
Correcting how covariant/contravariant types relate to their subtype/supertype in the descriptions.

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -23,7 +23,7 @@ class Baz[A]  // An invariant class
 
 ### Covariance
 
-A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `A` is a subtype of `B`, then `List[A]` is a subtype of `List[B]`. This allows us to make very useful and intuitive subtyping relationships using generics.
+A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `A` is a supertype of `B`, then `List[A]` is a supertype of `List[B]`. This allows us to make very useful and intuitive subtyping relationships using generics.
 
 Consider this simple class structure:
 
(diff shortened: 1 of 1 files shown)


SethTisue:
the first of the two changes doesn't seem like an improvement to me. in the covariant case, it's more natural to think about subtypes rather than supertypes, I think. the new text isn't wrong, but the old one seems fine to me.

as for the second change, I think the original text was correct and the new one is not.

hsmak:
Seth,
Thank you for taking the time in reviewing the PR.

This has been mind-boggling to me for quite some time. I will assume that I haven't got it yet. So, please bear with me.
As far as the Covariance & Contravariance go, I've always looked at them as being analogous to UpperBound & LowerBound, respectively.

So, I understand that:

- **Covariance**: allows assigning an instance to a variable whose type is one of the instance's generic type; i.e. _supertype_.
- **Contravariance**: allows assigning an instance to a variable whose type is one of the instance's derived type; i.e. _subtype_.

The same rule applies when passing to a method parameter.

**For the 1st case**

`List[+A]` was used as an example. Replacing the type with a concrete example:

`List[Animal] // So here A = Animal`

The Covariant of `A` must be a **subtype** of `A` say `B`. `A` is an UpperBound.
So I understand that the following must hold true:

`List[Animal] :> List[Cat] \\ where A = Animal & B = Cat`

`List[Animal] :> List[Dog] \\ where A = Animal & B = Dog`

So, this brings me to the conclusion that `A` is a supertype of `B`, or in other words `B` is a subtype of `A`. Therefore,

`val animals:List[Animal] = List[Cat]() \\ holds true!`

I get stuck when I read the description:

"making `A` covariant implies that for two types `A` and `B` where `A` is a subtype of `B`..." <=  
If `A` is a covariance, why would it be assumed as a subtype of `B`? Shouldn't it be the way around? 
And if the wording of "subtype" is more natural, Would saying `B` is a subtype of `A` make sense?

**For the 2nd case**

I'm not sure if it's the verbiage that's confusing me, but looking at it for a while I see why you insisted my correction was wrong.

I wonder if having a verbiage like the following would make it less confusing:

Let's take `List[-A]` as an example. Replacing the type with a concrete example:

`List[Cat] // So here A = Cat`

The Contravariant of `A` must be a **supertype** of `A` say `B`. `A` is a LowerBound.
So I understand that the following must hold true:

`List[Cat] :< List[Animal] \\ where A = Cat & B = Animal`

`List[Dog] :< List[Animal] \\ where A = Dog & B = Animal`

So, this brings me to the conclusion that `A` is a subtype of `B`, or in other words, `B` is a supertype of `A`. Therefore (via printer example),

`val catPrinter: Printer[Cat] = new Printer[Animal]{...} \\ holds true!`

Do you see modifying the original content to something close to the the above verbiage would improve it?

Thanks
Husain

SethTisue:
I'd suggest taking your questions about how covariance and contravariance work to https://users.scala-lang.org, where more people are available to help than the relatively small number of people who watch this repo.

Also, any good Scala book will have a chapter on variance. The language tour is really just a quick tour and not a full introduction to the language, so it's not necessarily bad or unexpected that this tour section isn't enough to convey a full understanding all by itself. (Nonetheless, I'm sure it can be improved!)

Once you're confident you grasp covariance and contravariance correctly, then I think you'll be in a better position to suggest improved wordings for the language tour.

hsmak:
Thanks again, Seth, and my apologies if I'm bringing a lengthy discussion in the wrong place.
I will make sure to be concise next time.

Regards
Husain

PR #1233, opened 2018-11-30
Scala types are erased at Run Time.

mukesh210:
Scala types are erased at Run Time not Compile Time. Please update.


SethTisue:
I'm not sure whether this is an improvement or not. Both wordings are reasonable: type erasure happens at compile time (it's literally a compiler phase, and it affects the code that is generated), but then you could also say that the effects of type erasure are felt primarily at runtime.

I actually prefer the original wording, as I think it's closer to being literally true, but perhaps there's an opportunity here to find a different wording entirely that makes the whole thing more clear.

PR #1234, opened 2018-11-30
Acronyms should be camelCased like other words (fixes #665)

### LINKED ISSUE #665: Style guideline should tell how to deal with acronym
In http://docs.scala-lang.org/style/naming-conventions.html#classestraits, it said that:
> Classes should be named in upper camel case:

However, there is not a rule about how to deal with acronym.

In Scala Standard Library and other Lightbend libraries, some acronyms are named in all upper cases, and others are named in upper camel case.
For example:
 * upper camel cases:
   * `scala.xml.Xhtml`
   * `play.api.libs.json.Json`
   * `scala.Enumeration.maxId`
 * all upper cases:
   * `scala.xml.XML` 
   * `scala.util.parsing.json.JSON`
   * `scala.swing.Applet.UI`
   * `scala.sys.process.ProcessIO`
   * `scala.SerialVersionUID`

Atry:
A bad decision is better than no decision

### DIFF
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -17,6 +17,16 @@ each word is capitalized, except possibly the first word:
     UpperCamelCase
     lowerCamelCase
 
+Acronyms should be treated as normal words:
+
+    Xhtml
+    maxId
+
+instead of:
+
+    XHTML
+    maxID
+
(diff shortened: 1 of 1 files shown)


jvican:
@SethTisue I've just restarted Seth :smile: It looks like a resolution error when downloading Scala.

PR #1235, opened 2018-12-03
Overhaul structuring and examples

yawaramin:
Since for-comprehensions are an important syntax sugar for Scala, I felt they deserve a bit of an overhaul and simpler but still motivating examples.


som-snytt:
I'd caution that the tour is for "bite-sized introductions" and not detailed specs. I would show each example first, then describe the parts.

I don't know how folks teach this, but I'd suggest showing a for loop first, because everyone knows what that must mean. Add that it's really sugar for `foreach`, and everyone also knows what that means.

Then say a for comprehension collects the results computed by the yield. Then add that it's sugar for other method calls.

Then show `for ()` versus `for {}` on multiple lines using multiple generators. The difference is only nl inference, right?

Then throw in a filter.

Finally add the bit about how semantics depend on the underlying type. Add a `Future` example, which is sort of whiz-bang.

_tour/for-comprehensions.md
```diff
@@ -13,54 +13,78 @@ previous-page: extractor-objects
...
+val pythagoreanTriples = for {
+  a <- naturals
+  b <- naturals if b > a
+  c <- naturals if a * a + b * b == c * c
+} yield (a, b, c)
+// Vector((3,4,5), (5,12,13), (6,8,10), (8,15,17), (9,12,15), (12,16,20))
 ```
 
-Note that comprehensions are not restricted to lists. Every datatype that supports the operations `withFilter`, `map`, and `flatMap` (with the proper types) can be used in sequence comprehensions.
+The filter after the second generator ensures that we don't generate redundant triples like (3, 4, 5) and (4, 3, 5) by forcing the second number in the triple to be larger than the first. The filter after the third generator enforces that only valid Pythagorean triples will be generated, i.e. that they will follow [Pythagoras' theorem](https://en.wikipedia.org/wiki/Pythagorean_theorem).
```
som-snytt: That is a fun example, with not too much explanation.

SethTisue:
@yawaramin interested in pursuing this further...?

yawaramin:
@SethTisue yes, I'll follow up with some edits by this weekend, thanks for the reminder :-)

SethTisue:
@yawaramin ping :-)

SethTisue:
closing for inactivity. can be reopened if activity resumes

som-snytt:
After this last weekend, I have new-found respect for le weekend. I mean how it devastates all plans and hopes and dreams.

PR #1236, opened 2018-12-04
copyedit new package object tour section

SethTisue:
(followup to https://github.com/scala/docs.scala-lang/pull/1203)

### DIFF
--- a/_ba/tour/package-objects.md
+++ b/_ba/tour/package-objects.md
@@ -7,7 +7,7 @@ discourse: true
 
 partof: scala-tour
 
-num: 35
+num: 36
 previous-page: packages-and-imports
 ---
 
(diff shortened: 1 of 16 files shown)


SethTisue:
review by @zafarella ? (and/or by anyone watching the repo)

SethTisue:
> lgtm. I think type annotation is not really required though

true in Scala 2.12, but procedure syntax will be deprecated in Scala 2.13 and removed in some version after that. 2.13 warns: "warning: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `foo`'s return type"

so we want to steer people away from it starting now.

SethTisue:
geez GitHub, do you have to show how the world how many force-pushes it's taking me to get this right? 🤨

Philippus:
Maybe it would be good to add a note to discourage defining classes/objects inside of package objects. (which will result in warnings when using `-Xlint`)

PR #1238, opened 2018-12-08

mavaji:


Philippus:
Judging by https://en.wikipedia.org/wiki/Currying#Contrast_with_partial_function_application you are right.

SethTisue:
agree, it's better not to talk about "currying" in Scala at all unless we're talking about signatures such as `A => (B => C)`:

```
scala 2.12.8> def add(x: Int, y: Int) = x + y
add: (x: Int, y: Int)Int

scala 2.12.8> (add _).curried
res0: Int => (Int => Int) = scala.Function2$$Lambda$1078/828312047@3a66e67e
```

this is something different than multiple parameter lists.

PR #1240, opened 2018-12-08
Polish translation of Tour of Scala: Nested methods

kamilduda:
Update Polish version of nested functions

Add Polish translation for higher - order functions


SethTisue:
awesome to have this, thanks!

PR #1242, opened 2018-12-10
Fixed parameters vs arguments use and wording regarding substitution

fabiopakk:
In the first paragraph of "Other Examples", 'argument' was used instead of 'parameter'. Parameters refer to functions' or methods' definition, whereas arguments refer to function calls or method calls.

In the last paragraph of "Other Examples", you can say either:

* ... substitute the former *by* the latter, ... <or>
* ... substitute the latter *for* (every occurrence) of the former, ...


SethTisue:
thank you!

PR #1243, opened 2018-12-10
More precise wording regarding 'pairs' example

fabiopakk:
The second example is stated as "computes all pair ..." however the pair (9, 1), among other valid pairs, is not withing the results. Although (1, 9)  is listed, (1, 9) is different than (9, 1). Just remembering that pairs are tuples of size 2 and are ordered, as opposed to a 'pair' in natural language. I changed the code accordingly and tested prior to submission.

### DIFF
--- a/_tour/for-comprehensions.md
+++ b/_tour/for-comprehensions.md
@@ -37,19 +37,19 @@ Here is a more complicated example using two generators. It computes all pairs o
 ```tut
 def foo(n: Int, v: Int) =
    for (i <- 0 until n;
-        j <- i until n if i + j == v)
+        j <- 0 until n if i + j == v)
    yield (i, j)
 
 foo(10, 10) foreach {
   case (i, j) =>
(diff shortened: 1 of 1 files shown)


SethTisue:
I'd suggest leaving the code alone and just changing the text as needed, partly because this change results in a very long line, too wide for the site layout.

fabiopakk:
Dear reviewers,

I appreciate your comments. Pondering between changing the code vs changing the text, I chose the former because in this way we have a better separation between two generators (each one introducing a new variable) and a filter involving both generators. Though it is not wrong to restrict the range of a generator with the value of another generator, I feel it is more didactical to introduce generators with two examples that are independent, allowing the filter to connect them.

SethTisue:
okay, I don't have a strong feeling either way, but something should be done about the very long line

SethTisue:
something could be done about the long lines in a separate PR.

PR #1244, opened 2018-12-12
Removed redundent import from the example to reinforce precision …

aloklal99:
…and promote better understanding


aloklal99:
@Philippus My bad, I should have caught that related change, too.  I can submit a consolidated merge once you had chance to review the rework diff.

PR #1247, opened 2018-12-19
Rename abstract type members page

### LINKED ISSUE #1214: Rename the "Abstract type" page?
The page https://docs.scala-lang.org/tour/abstract-types.html is actually about abstract type members (which require the enclosing type to be abstract) and not about abstract types in general (i.e https://en.wikipedia.org/wiki/Abstract_type).

What do you think about renaming it and tweaking the first sentence to explain?

dwijnand:
Fixes #1214


julienrf:
Is it possible to redirect `abstract-type.html` requests to `abstract-type-members.html`?

dwijnand:
@julienrf that's what the `redirect_from: "/tour/abstract-types.html"` is for.

_tour/abstract-type-members.md
```diff
@@ -1,20 +1,23 @@
...
 partof: scala-tour
 num: 23
 next-page: compound-types
 previous-page: inner-classes
-topics: abstract types
+topics: abstract type members
 prerequisite-knowledge: variance, upper-type-bound
 
 redirect_from: "/tutorials/tour/abstract-types.html"
+redirect_from: "/tour/abstract-types.html"
```
dwijnand: Redirecting the hold URL

julienrf:
Awesome, I missed that part :)

PR #1254, opened 2019-01-06
use correct package name in package object docs

benjaminran:
~~These 2 occurrences of 'gardening' were probably meant to be 'gardening.fruits'. Let me know if I'm missing something.~~

Edit: the code I changed in my first commit was actually correct, so I reverted that. But `planted` and `showFruit` are being added to the package object for `gardening.fruits`, not `gardening`, so I think the text change is correct.


SethTisue:
thank you!

PR #1256, opened 2019-01-08
Docker support

kotobotov:
while i try to install Ruby with gems and jekyll for windows and got so many inconsistense and mess with versions.
If you are not ruby guy it's really pain in *** so i find the docker is more pleasant way to start working with docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,14 @@ To build and view the site locally:
 
 For more details, read on.
 
+## Quickstart with Docker ##
+
+To build and view site with docker:
+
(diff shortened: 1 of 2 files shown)


SethTisue:
I don't know Docker, is anyone able to review this?

SethTisue:
optimistically merging. I hope this proves helpful

kotobotov:
i used it, but container not include "tut rotune", 
it's ok to start working, if it's important, maybe someone add in future if it's important.

PR #1266, opened 2019-01-19
type comes before value in val declarations

ePaul:
See https://www.scala-lang.org/files/archive/spec/2.12/04-basic-declarations-and-definitions.html#value-declarations-and-definitions

### DIFF
--- a/_tour/tuples.md
+++ b/_tour/tuples.md
@@ -22,7 +22,7 @@ Tuples come in handy when we have to return multiple values from a function.
 A tuple can be created as:
 
 ```tut
-val ingredient = ("Sugar" , 25):Tuple2[String, Int]
+val ingredient : Tuple2[String, Int] = ("Sugar" , 25)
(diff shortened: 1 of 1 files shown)


_tour/tuples.md
```diff
@@ -22,7 +22,7 @@ Tuples come in handy when we have to return multiple values from a function.
 A tuple can be created as:
 
 ```tut
-val ingredient = ("Sugar" , 25):Tuple2[String, Int]
+val ingredient : Tuple2[String, Int] = ("Sugar" , 25)
```
ePaul: I guess the old syntax is also valid, it is a [typed expression](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#typed-expressions).

SethTisue:
I've submitted an overhaul of this tour section here: https://github.com/scala/docs.scala-lang/pull/1268

PR #1267, opened 2019-01-20
Adjust the language of Chinese

yuchenghu:
Adjust the language to make it easier for first time language users.


yuchenghu:
You are very welcome.

If need me to do some task, please let me know.

Thanks.

PR #1268, opened 2019-01-21
language tour: overhaul tuples section

SethTisue:


_tour/tuples.md
```diff
@@ -14,82 +14,71 @@ topics: tuples
...
-
   println(a * b)
-  
 }
 ```
 
-The value () of type Unit is conceptually the same as the value () of type Tuple0. There can only be one value of this type since it has no elements.
+## Tuples and case classes
+
+Users may sometimes find hard to choose between tuples and case classes. Case classes have named elements. The names can improve the readability of some kinds of code. In the planet example above, we might define `case class Planet(name: String, distance: Double)` rather than using tuples.
```
Philippus: missing an `it` between `find` and `hard`

NthPortal:
```suggestion
Users may sometimes find it hard to choose between tuples and case classes. Case classes have named elements. The names can improve the readability of some kinds of code. In the planet example above, we might define `case class Planet(name: String, distance: Double)` rather than using tuples.
```

NthPortal:
```suggestion
Users may sometimes find it hard to choose between tuples and case classes. Case classes have named elements. The names can improve the readability of some kinds of code. In the planet example above, we might define `case class Planet(name: String, distance: Double)` rather than using tuples.
```

PR #1270, opened 2019-01-24
Fix a word from 'variable' to 'value'

hatemogi:
At 6th paragraph in [Tour of Scala - Unified Types](https://docs.scala-lang.org/tour/unified-types.html), it seems that the wording 'variable' is actually meant to be 'value'.

### DIFF
--- a/_tour/unified-types.md
+++ b/_tour/unified-types.md
@@ -40,7 +40,7 @@ val list: List[Any] = List(
 list.foreach(element => println(element))
 ```
 
-It defines a variable `list` of type `List[Any]`. The list is initialized with elements of various types, but they all are instance of `scala.Any`, so you can add them to the list.
+It defines a value `list` of type `List[Any]`. The list is initialized with elements of various types, but they all are instance of `scala.Any`, so you can add them to the list.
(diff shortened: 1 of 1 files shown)


_tour/unified-types.md
```diff
@@ -40,7 +40,7 @@ val list: List[Any] = List(
 list.foreach(element => println(element))
 ```
 
-It defines a variable `list` of type `List[Any]`. The list is initialized with elements of various types, but they all are instance of `scala.Any`, so you can add them to the list.
+It defines a value `list` of type `List[Any]`. The list is initialized with elements of various types, but they all are instance of `scala.Any`, so you can add them to the list.
```
propensive: There's a grammatical mistake on this line. How about, `s/they all are instance of/each is an instance of/`, or similar?

SethTisue:
"variable" is unfortunately ambiguous, but "value" is too; in e.g. `val x = 3` I would tend to call `x` a "val" and `3` a "value"

regardless, this PR constitutes an improvement, so I'll merge it. but maybe we could do even better? not sure.

PR #1272, opened 2019-01-29
fix mismatching explanation: it is about currying itself, not the …

hatemogi:
At [Multiple Prameter Lists](https://docs.scala-lang.org/tour/multiple-parameter-lists.html) in Tour Of Scala, it seems that the explanation is about "currying" itself, not about the code snippet above. It might have been mistakenly edited about 6 months ago.


SethTisue:
thank you @hatemogi !

PR #1279, opened 2019-02-04
remove Object extending App from example

cobbce:
App has not been introduced in the tutorial up to this point. This code does not execute as-is in ScalaFiddle. Removing to achieve consistency with tutorials up to this point and to work out of the box for those following along from the beginning.


SethTisue:
thanks!

PR #1282, opened 2019-02-06
Turn code example gender-agnostic

vrinek:
- it also conveys the fact that any piece of code is usually not the product of one coder but a collection of them


ashawley:
Yeah, this makes sense.  It's also good the example isn't an "us versus them".

PR #1283, opened 2019-02-07
add a missing word 'million'.

hatemogi:
In the example code in [TOUR OF SCALA - TUPLES](https://docs.scala-lang.org/tour/tuples.html), there is a missing word, 'million'.


hatemogi:
Thank you. 😄

PR #1287, opened 2019-02-12
More precise wording in abstract-type-members.md

fabiopakk:
Dear reviewers,

I summarize below the reason for my changes:
* Line 37: "Notice how we can use yet another abstract type U as an upper-type-bound." -> That suggests that T <: U, which is not the case.
* Line 56: Make it clear that the anonymous class is providing concrete types to abstract type members.
* Line 78: The word 'members' is a must, as 'abstract type' is a broader than intended term. Would also be nice to show and example where abstract type members can't be rewritten with type parameters as claimed.

### DIFF
--- a/_tour/abstract-type-members.md
+++ b/_tour/abstract-type-members.md
@@ -34,7 +34,7 @@ abstract class SeqBuffer extends Buffer {
   def length = element.length
 }
 ```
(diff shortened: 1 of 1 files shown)


SethTisue:
thanks!

PR #1290, opened 2019-02-18
Fix Promise creation

sergeyh:


SethTisue:
good catch, thank you!

PR #1299, opened 2019-03-14
Alt formatting for the impl of Ord.lt

dwijnand:
I didn't immediately understand the code with its previous formatting, so I'm proposing this, but I'm happy to tweak it.  What do you think?


dwijnand:
Hmm, no still not convinced.

SethTisue:
I liked the indentation part, but not the added-braces part

Issue #1295, opened 2019-03-12
Glossary: the link for "filter" does not point to the definition of filter


lorenzpahl:
Clicking on the "filter" entry (https://docs.scala-lang.org/glossary/index.html#filter) in the sidebar of the glossary jumps to the search bar on the top of the page; not to the actual definition of the term.

Both, the input element of the search bar and the heading element of the term have the same id attribute value. That's why the link jumps to the search bar.

PR #1300, opened 2019-03-14
Resolve duplicate id attribute value

lorenzpahl:
This PR changes the `id` attribute value of the `input` element on the glossary page to resolve a naming conflict with the `id` value of the heading element for the term "filter".

Resolves: #1295


SethTisue:
thank you for noticing and fixing, much appreciated

Issue #916, opened 2017-11-01
Long sidebars don't scroll until you reach the bottom of the page


Matt343:
If a page has a sidebar that is long enough to extend past the bottom of the screen (like https://docs.scala-lang.org/glossary/), you cannot see the bottom portion of the sidebar without scrolling to the bottom of the entire page. This makes using the sidebar for navigation clumsy, since you usually want to use the sidebar to avoid scrolling all the way down the page. This is particularly bad on lower resolution screens.

Here is what the sidebar looks like when halfway down the page: https://i.imgur.com/zk1auSv.png

Here is what it looks like at the bottom of the page: https://i.imgur.com/wEeQERD.png

Tested on Chrome on Windows and Ubuntu

heathermiller:
PR welcome!

PR #1301, opened 2019-03-14
Improve sidebar usability

lorenzpahl:
This is an attempt to improve the usability of the sidebars, especially if the sidebar is so long that it exceeds the viewport. It limits the maximum height of the TOC such that the sidebar will not exceed the height of the viewport. A scrollbar is shown if the TOC gets too long.

Additionally, the sticky behaviour of the sidebar is now implemented with CSS. This does not really add anything, but it would remove one JavaScript library dependency (which is only used for this feature) and browser support for `position: sticky` is [quite good](https://caniuse.com/#feat=css-sticky).

Feedback welcome!

Resolves: #916


SethTisue:
thank you!

PR #1303, opened 2019-03-19
Update files.md

igorepim:
I propose to remove this part, because it's not a real recommendation. Something in the style "do whatever you want if it is good for you". I think that putting multiple entities in a single file is certainly a bad practice. Especially when we speak about a large complex project.

"Another case is when multiple classes logically form a single, cohesive
group, sharing concepts to the point where maintenance is greatly served
by containing them within a single file. These situations are harder to
predict than the aforementioned sealed supertype exception. Generally
speaking, if it is *easier* to perform long-term maintenance and
development on several units in a single file rather than spread across
multiple, then such an organizational strategy should be preferred for
these classes. However, keep in mind that when multiple units are
contained within a single file, it is often more difficult to find
specific units when it comes time to make changes."

### DIFF
--- a/_style/files.md
+++ b/_style/files.md
@@ -50,17 +50,6 @@ Because of the nature of sealed superclasses (and traits), all subtypes
 qualifies as an instance where the preference for single-unit files
 should be ignored.
 
-Another case is when multiple classes logically form a single, cohesive
-group, sharing concepts to the point where maintenance is greatly served
-by containing them within a single file. These situations are harder to
-predict than the aforementioned sealed supertype exception. Generally
-speaking, if it is *easier* to perform long-term maintenance and
-development on several units in a single file rather than spread across
-multiple, then such an organizational strategy should be preferred for
-these classes. However, keep in mind that when multiple units are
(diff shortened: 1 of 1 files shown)


SethTisue:
hmm, interesting. I do agree with the text you want to remove, I often put a tight cluster of short, closely related class in a single source file. I think doing so is pretty common

I think I'd be more comfortable with softening the wording a bit to be clear that many shops do this, but there's not a firm consensus on it?

any else have an opinion on this?

SethTisue:
closing for inactivity. can always be reopened if activity resumes

Ichoran:
Aside: I also agree with the existing text.  Using single files for highly interrelated things makes maintaining related things way easier; the task of finding where something is defined is usually pretty easy (e.g. `grep "trait MyThing" *.scala` compared to the chore of correctly updating a dozen tiny files.  So I disagreed with the change.  In fact, I'd suggest discouraging circular dependencies to the point of including them in single files if it's practical to do so.  But, with no activity, no reason to debate it either way.

PR #1304, opened 2019-03-21
replace JavaConversions to JavaConverters

Xztty:
from scala 2.12.0 the JavaConversions is marked as deprecated class.
And it's recommended to use JavaConverters


SethTisue:
thank you!

PR #1306, opened 2019-03-26
Fixed Typo in Conversions Table for Parallel Collections

salim7:


salim7:
I just noticed that the same typo also exists in the translated versions of this document.

SethTisue:
thank you!

PR #1319, opened 2019-04-26
Update building-a-scala-project-with-intellij-and-sbt.md

YikSanChan:


dwijnand:
This is kind of the nuclear option.  Maybe there should be a section towards the bottom with some troubleshooting advice, and then mention that section at the beginning (so people already know about it).  I think that's better than an inline comment like this PR proposes.

SethTisue:
I agree that it's too distracting from the main flow to have this in-line.

SethTisue:
thank you, @YikSanChan, for the suggestion, we appreciate it.

PR #1323, opened 2019-05-02
Linkify the Oracle LTS support URL

dwijnand:


dwijnand:
This should fix the page, that currently looks like this:

image

PR #1326, opened 2019-05-10
Add guidance on how to migrate scala.Seq

eed3si9n:


_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
 
 The [scala-collection-compat](https://github.com/scala/scala-collection-compat) is a library released for 2.11, 2.12 and 2.13 that provides some of the new APIs from Scala 2.13 for the older versions. This simplifies cross-building projects.
 
 The module also provides [migratrion rules](https://github.com/scala/scala-collection-compat#migration-tool) for [scalafix](https://scalacenter.github.io/scalafix/docs/users/installation.html) that can update a project's source code to work with the 2.13 collections library.
 
+## scala.Seq migration
+
+In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]`, instead of `scala.collection.Seq[A]`. This change requires some planning depending on how your code is going to be used.
+
+If you're making an application, and simply migrating Scala 2.12 code base to 2.13, it might be ok to keep using `scala.Seq` in your code.
```
dwijnand:
```suggestion
If you're making an application, and simply migrating a Scala 2.12 code base to 2.13, it might be ok to keep using `scala.Seq` in your code.
```

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
+
+If you're making a library intended to be used by other programmers, then using `scala.Seq` or vararg is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
```
dwijnand:
```suggestion
If you're making a library intended to be used by other programmers, then using `scala.Seq` or varargs is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
```

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
+import scala.collection.Seq
+~~~
+
+In the future when your API is able to break the source compatibility, it might also make sense to migrate towards the `scala.collection.immutable.Seq` ("ISeq") for both Scala 2.12 and Scala 2.13.
+
+~~~ scala
+import scala.collection.immutable.Seq
+~~~
+
+Note that in Scala 2.13 the sequence passed into a vararg as `orderFood(xs: _*)` must also be immutable. This is because the sequence passed into a vararg must conform to`scala.Seq` acoording to [Specification](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#function-applications). Thus if your API exposes varargs, it will be an unavoidable breaking change. This might affect Java interoperability.
```
dwijnand:
```suggestion
Note that in Scala 2.13 the sequence passed into as a varargs as `orderFood(xs: _*)` must also be immutable. This is because the sequence passed in as a varargs must conform to `scala.Seq` according to [SLS 6.6](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#function-applications). Thus, if your API exposes varargs it will be an unavoidable breaking change. This might affect Java interoperability.
```

dwijnand:
Not being able to pass an array I don't think is "a breaking change in the API semantics".  It's just source-incompatible.

I actually think continuing to use `scala.Seq` is what should be used for libraries.  2.12 code will continue to use `scala.collection.Seq` and 2.13 code will use `scala.collection.immutable.Seq`.  That might mean some arrays need to `.toSeq` before `: _*`.

eed3si9n:
> Not being able to pass an array I don't think is "a breaking change in the API semantics". It's just source-incompatible.

The meaning of your API changed, so I think it's a breaking change. For most libraries it might be an acceptable change, but for others it might not work. We don't know.

The vararg you have no option, but you do have an option to use CSeq for your own API.

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
+- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
+- if your library users frequently uses mutable collections such as `Array`
+
+you can import `scala.collection.Seq` ("CSeq") explicitly in your code.
+
+~~~ scala
+import scala.collection.Seq
+~~~
+
+In the future when your API is able to break the source compatibility, it might also make sense to migrate towards the `scala.collection.immutable.Seq` ("ISeq") for both Scala 2.12 and Scala 2.13.
```
NthPortal: it doesn't seem like you refer to "ISeq" or "CSeq" anywhere, so the parentheticals can probably be removed

eed3si9n: ok.

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
+In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]`, instead of `scala.collection.Seq[A]`. This change requires some planning depending on how your code is going to be used.
+
+If you're making an application, and simply migrating a Scala 2.12 code base to 2.13, it might be ok to keep using `scala.Seq` in your code.
+
+If you're making a library intended to be used by other programmers, then using `scala.Seq` or varargs is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
+
+- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
+- if your library users frequently uses mutable collections such as `Array`
+
+you can import `scala.collection.Seq` ("CSeq") explicitly in your code.
```
lrytz: It might be worth mentioning that this is a potentially source-breaking change. If a user in 2.12 has:

```
// `Seq` is the default `scala.Seq`
val a: Seq[Food] = orderFood(myOrder)
```

Then that will no longer compile in 2.13

lrytz: Also, it might be a good strategy for APIs to only move to `immutable.Seq` in the result type, but keep accepting `collection.Seq` arguments, if the method implementation can work fine with mutable collections,.

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
 
 The [scala-collection-compat](https://github.com/scala/scala-collection-compat) is a library released for 2.11, 2.12 and 2.13 that provides some of the new APIs from Scala 2.13 for the older versions. This simplifies cross-building projects.
 
 The module also provides [migratrion rules](https://github.com/scala/scala-collection-compat#migration-tool) for [scalafix](https://scalacenter.github.io/scalafix/docs/users/installation.html) that can update a project's source code to work with the 2.13 collections library.
 
+## scala.Seq migration
+
+In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]`, instead of `scala.collection.Seq[A]`. This change requires some planning depending on how your code is going to be used.
+
+If you're making an application, and simply migrating a Scala 2.12 code base to 2.13, it might be ok to keep using `scala.Seq` in your code.
```
szeiger: This sounds like "keep using `scala.Seq`" should be the safe and simple option. I'd argue it's the opposite (unless your code happens to just work out of the box). In a complex codebase the simplest way forward is to change all references from `scala.Seq` to `scala.collection.Seq` and possibly insert `toSeq` calls where necessary. This is easy to do and will cross-compile on 2.12 without any changes in semantics.

Switching your own codebase from `scala.collection.Seq` to `scala.immutable.Seq` is the more advanced refactoring. In many cases it may have been the intention all along, so it makes sense to do it. It also avoids unnecessary copying when varargs are involved, but it's much harder to get right consistently in a large codebase.

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +23,38 @@ The most important changes in the Scala 2.13 collections library are:
...
+
+If you're making a library intended to be used by other programmers, then using `scala.Seq` or varargs is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
+
+- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
+- if your library users frequently uses mutable collections such as `Array`
```
szeiger: If code has been optimized to the point that you build an array instead of a regular collection, you're probably okay calling `unsafeWrapArray`, so it will still work with immutable varargs.

eed3si9n: Arrays might come up simply as Java API that you provide as a subset of API.

_overviews/core/collections-migration-213.md
eed3si9n: Interesting. I'll mention recommendation to use scala.collection.Seq first then.

_overviews/core/collections-migration-213.md
eed3si9n: Added commits that ended up adding references to ISeq and CSeq.

_overviews/core/collections-migration-213.md
eed3si9n: I've added this in 2b4de2d.

_overviews/core/collections-migration-213.md
eed3si9n: I've added this in 2b4de2d.

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +24,103 @@ The most important changes in the Scala 2.13 collections library are:
...
+}
+~~~
+
+Similarly, if you're making an end-user application, unifying to CSeq might be the easier and safer initial path especially for a larger and complex code base. Switching to ISeq will be a more advanced refactoring.
+
+Note that in Scala 2.13 the sequence passed into as a varargs as `orderFood(xs: _*)` must also be immutable. This is because the sequence passed in as a varargs must conform to `scala.Seq` according to [SLS 6.6](https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#function-applications). Thus, if your API exposes varargs it will be an unavoidable breaking change. This might affect Java interoperability.
+
+### Masking scala.Seq
+
+To use the compiler to bad the use of plain `Seq`, you can declare your own `Seq` to mask `scala.Seq`.
```
dwijnand: s/bad/ban/

_overviews/core/collections-migration-213.md
```diff
@@ -22,14 +24,103 @@ The most important changes in the Scala 2.13 collections library are:
...
+## scala.Seq and scala.IndexedSeq migration
+
+In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]` ("ISeq"), instead of `scala.collection.Seq[A]` ("CSeq"). Similarly, `scala.IndexedSeq[+A]` is an alias for `scala.collection.immutable.IndexedSeq[A]`. These changes require some planning depending on how your code is going to be used.
+
+If you're making a library intended to be used by other programmers, then using `scala.Seq`, `scala.IndexedSeq`, or vararg is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
+
+- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
+- if your library users frequently uses mutable collections such as `Array`
+
+you can import collection Seq explicitly in your code.
```
dwijnand: "import collection Seq" looks weird and kind of ambiguous.

PR #1330, opened 2019-05-17
Refine pattern-matching.md

asakaev:
number is a better name than many here.

### DIFF
--- a/_tour/pattern-matching.md
+++ b/_tour/pattern-matching.md
@@ -28,19 +28,19 @@ x match {
   case 0 => "zero"
   case 1 => "one"
   case 2 => "two"
-  case _ => "many"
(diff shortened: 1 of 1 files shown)


dwijnand:
I think the "many" is short for "more than 2? woah, that's too many to count".

Why do you prefer "number"?

asakaev:
I like "many" a lot. It's very clear, but zero and negative numbers do not match "many" definition.

dwijnand:
Ah, I see your point.  Perhaps "other", then.

And maybe even mention the fact that the int returned by `Random.nextInt(10)` is known to be between 0 and 10, but that's not in the type?

asakaev:
Exactly what is needed!

PR #1331, opened 2019-05-18

suyash1:
updating doc for adding scala framework support addition which was missing earlier

### DIFF
--- a/getting-started-intellij-track/getting-started-with-scala-in-intellij.md
+++ b/getting-started-intellij-track/getting-started-with-scala-in-intellij.md
@@ -36,7 +36,7 @@ take a few minutes but subsequent projects can use the same SDK.
 ## Writing code
 
 1. On the **Project** pane on the left, right-click `src` and select
-**New** => **Scala class**. If you don't see **Scala class**, right-click on **HelloWorld** and click on **Add Framework Support...**, select **Scala** and proceed. If you see **Error: library is not specified**, you can either click donload button, or select the library path manually.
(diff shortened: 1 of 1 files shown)


getting-started-intellij-track/getting-started-with-scala-in-intellij.md
```diff
@@ -36,7 +36,7 @@ take a few minutes but subsequent projects can use the same SDK.
 ## Writing code
 
 1. On the **Project** pane on the left, right-click `src` and select
-**New** => **Scala class**.
+**New** => **Scala class**. If you don't see **Scala class**, right-click on **HelloWorld** and click on **Add Framework Support...**, select **Scala** and proceed. If you see **Error: library is not specified**, you can either click donload button, or select the library path manually.
```
dwijnand: There's a typo in "download".

NthPortal:
```suggestion
**New** => **Scala class**. If you don't see **Scala class**, right-click on **HelloWorld** and click on **Add Framework Support...**, select **Scala** and proceed. If you see **Error: library is not specified**, you can either click download button, or select the library path manually.
```

suyash1: Fixed typo.
@dwijnand IMHO putting this change in prerequisite section will be better because any new installation will need this step.

SethTisue:
> But, in alternative to giving the information in-line, should we try and have it in a preferring prerequisite section or a later troubleshooting section?

maybe if any _more_ troubleshooting information ends up being added, we'll need to do that, but I think it's okay for now

SethTisue:
thank you @suyash1 and reviewers!

PR #1332, opened 2019-05-21
RC2 is here

SethTisue:


SethTisue:
live at https://docs.scala-lang.org/api/all.html

PR #1333, opened 2019-05-21
Tweak Seq migration guide

dwijnand:


_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,142 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
-In the future when your API is able to break the source compatibility, it might also make sense to migrate towards ISeq for both Scala 2.12 and Scala 2.13.
+### Option 3: use immutable sequences
+
+The third migration strategy is to change your API to use immutable sequences for both parameter and result
+types.  When cross-building your library for Scala 2.12 and 2.13 this could either mean:
+
+- continuing to use `scala.Seq` which means it stays source and binary-compatible in 2.12, but would have to
+    have immutable sequence semantics (but that might already be the case).
+- switch to explicitly using immutable Seq in both Scala 2.12 and 2.13, which means breaking source, binary and
+    semantic compatibility in 2.12:
```
dwijnand:
```suggestion
    (possibly) semantic compatibility in 2.12:
```

dwijnand:
Polling for what names we should use for Seqs: https://www.strawpoll.me/18027945

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,142 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
 
-- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
-- if your library users frequently uses mutable collections such as `Array`
+[SLS 6.6]: https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#function-applications
 
-you can import collection Seq explicitly in your code.
+Therefore any method signature in Scala 2.13 which includes `scala.Seq`, varargs, or `scala.IndexedSeq` is going
+to have a breaking change in API semantics (as the immutable sequence types require more, immutability, than the
+not-immutable types).  For example, with a method like `def orderFood(order: Seq[Order]): Seq[Food]`, previously
+the method user would have been able to pass in an array of `Order`, but that won't work for 2.13.
```
szeiger: `Array` will still work (albeit with a deprecation warning). I'd use a different example here like `ArrayBuffer`.

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,142 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
+### Option 1: migrate back to scala.collection.Seq
+
+The first, in some ways simplest, migration strategy for all non-varargs usages of `scala.Seq` is to replace
+them with `scala.collection.Seq` (and require users to call `.toSeq` or `unsafeWrapArray` when passing such
+sequences to varargs methods).  We recommend, in order of preference, doing such a replacement with one of:
+
+- `import scala.{ collection => sc }` and using `sc.Seq`
+- `import scala.collection.{ Seq => CSeq }` and using `CSeq`
+- replacing `Seq[...]` usage with the fully-qualified `scala.collection.Seq[...]` (e.g. in code gen or macros)
+- `import scala.collection.Seq`, which shadows `scala.Seq` and is a 1-line change, but causes name confusion
```
szeiger: I don't think we should have too many different recommendation here. And I wouldn't put `import scala.collection.Seq` into this list, unless "in order of preference" means "from what you should do to what you definitely should not do".

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,142 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
+  * import scala.{ collection => sc }
+  * import scala.collection.{ immutable => sci }
+  * }}}
   *
-  * This Seq trait is a dummy type to prevent the use of `Seq`.
+  * This `Seq` trait is a dummy type to prevent the use of `Seq`.
   */
-@compileTimeOnly("Use ISeq or CSeq") private[example] trait Seq[A1, F1[A2], A3]
+@compileTimeOnly("Use sci.Seq or sc.Seq")
+private[example] trait Seq[A1, F1[A2], A3]
```
szeiger: Why all the type parameters?

_overviews/core/collections-migration-213.md
dwijnand:
```suggestion
the method user would have been able to pass in an `ArrayBuffer` of `Order`, but that won't work for 2.13.
```

_overviews/core/collections-migration-213.md
dwijnand: Yeah, let's move it out and mention it as the bad option.  And use the winner of the strawpoll in place of the first two options.  And then maybe move the code-gen/macro option separate too.

_overviews/core/collections-migration-213.md
dwijnand: ¯\\\_(ツ)\_/¯

dwijnand:
LOL

```
jekyll 3.8.3 | Error:  Liquid syntax error (line 139): Variable '{{{
  * import scala.{ collection => sc }' was not properly terminated with regexp: /\}\}/
```

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,140 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
-## scala.Seq and scala.IndexedSeq migration
+## scala.Seq, varargs and scala.IndexedSeq migration
 
-In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]` ("ISeq"), instead of `scala.collection.Seq[A]` ("CSeq"). Similarly, `scala.IndexedSeq[+A]` is an alias for `scala.collection.immutable.IndexedSeq[A]`. These changes require some planning depending on how your code is going to be used.
+In Scala 2.13 `scala.Seq[+A]` is an alias for `scala.collection.immutable.Seq[A]`, instead of `scala.collection.Seq[A]`, and `scala.IndexedSeq[+A]` is an alias for `scala.collection.immutable.IndexedSeq[A]`. These changes require some planning depending on how your code is going to be used.
 
-If you're making a library intended to be used by other programmers, then using `scala.Seq`, `scala.IndexedSeq`, or vararg is going to be a breaking change in the API semantics. For example, if there was a function `def orderFood(order: Seq[Order]): Seq[Food]`, previously the library user would have been able to pass in an array of `Order`, but it won't work for 2.13.
+This also has the effect of making the type of varargs parameters immutable sequences, due to [SLS 6.6][], so in
```
NthPortal: a few issues:

- change from plural to singular ("These changes require" => "This [...] has")
- no previous effects were mentioned, so the word "also" is confusing
- unqualified "this" - "this change" would be a little better (this one is rather nitpicky and mostly channeling my high school english teacher)

NthPortal: is there meant to be a link for `SLS 6.6`?

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,140 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
+a method such as `orderFood(xs: _*)` the varargs parameter `xs` must be an immutable sequence.
 
-- if you cross build with Scala 2.12 and want to maintain the API semantics for 2.13 version of your library, or
-- if your library users frequently uses mutable collections such as `Array`
+[SLS 6.6]: https://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#function-applications
 
-you can import collection Seq explicitly in your code.
+Therefore any method signature in Scala 2.13 which includes `scala.Seq`, varargs, or `scala.IndexedSeq` is going
+to have a breaking change in API semantics (as the immutable sequence types require more, immutability, than the
```
NthPortal: this parenthetical reads a bit awkwardly to me. replacing the commas around "immutability" with em dashes *might* help a little, but it still feels odd to me

_overviews/core/collections-migration-213.md
```diff
@@ -31,95 +31,140 @@ The [scala-collection-compat](https://github.com/scala/scala-collection-compat)
...
+not-immutable types).  For example, with a method like `def orderFood(order: Seq[Order]): Seq[Food]`, previously
```
NthPortal: the commas in this sentence don't flow well for me. would the sentence be a bit smoother rearranged? perhaps the following:

> For example, users of a method like `def orderFood(order: Seq[Order]): Seq[Food]` would previously have been able to pass in an `ArrayBuffer` of `Order`, but cannot in 2.13.

_overviews/core/collections-migration-213.md
dwijnand: The only change I didn't apply is the "also" part.  Do you think it's not obvious to the reader the immediate impact of changing the alias?  We can fix this up still.

SethTisue:
this is super. (and, I like the decision to not introduce `CSeq`/`ISeq`)

PR #1343, opened 2019-06-02
Fix Chinese Simple(zh_CN) version of issue#1339

dongxuwang:
based on commit https://github.com/scala/docs.scala-lang/commit/55ed64e205d80298f59a6724bd7c44a26ec6e0e7

(Fix: should extend NotUniversal trait)


Philippus:
thank you!

PR #1347, opened 2019-06-04
Changed the version in collections-2.13 introduction.md from 2.8 to 2.13

matomer:
Are the versions supposed to be 2.13 instead of 2.8 or am I missing something?


matomer:
Are the versions supposed to be 2.13 instead of 2.8 or am I missing something?

PR #1348, opened 2019-06-04
Fix the version in the 2.13 collections introduction.md

matomer:
Is the version supposed to be 2.13 or am I missing something?


dwijnand:
This is still copy-pasted from the 2.8 collections guide, so probably most of the introduction needs a rewrite.

julienrf:
I would just remove the whole paragraph, instead.

dwijnand:
Fixed in #1359.

PR #1349, opened 2019-06-05
Typo: "seemless" to "seamless"

nfd9001:


som-snytt:
What if they meant to say seemly.

NthPortal:
then it would seem that their grammar is slightly unseemly ;)

PR #1350, opened 2019-06-06
Fix buggy Regex definition

changlinli:
The original Regex definition doesn't fully match the `background-image` value.


SethTisue:
thank you!

PR #1351, opened 2019-06-06

cclaudiu81:
enhanced doc on methods that take fns as args

### DIFF
--- a/_tour/higher-order-functions.md
+++ b/_tour/higher-order-functions.md
@@ -19,6 +19,8 @@ The terminology can get a bit confusing at this point, and we use the phrase
 "higher order function" for both methods and functions that take functions as parameters
 or that return a function.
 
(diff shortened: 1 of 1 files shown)


cclaudiu81:
you can do it in any place you want :) i just wanted to emphasize the idea that overusing closures when objects are used to represent the domain could produce some dangerous things... they can easily become antipatterns just like everything which is overused...In this context "leak internal state" i see it as a synonym for "escape internal state"...but again if you consider "leak internal state" is more appropriate then go ahead and use it.

SethTisue:
@cclaudiu81 there isn't anyone who's job it is to take pull requests and put them into final form. if you're interested in responding to the review feedback yourself, great. if not, we'll close the PR, while inviting other volunteers to submit their own versions if they like.

cclaudiu81:
@SethTisue will add this information as PR changes. thanks!

PR #1352, opened 2019-06-06

cclaudiu81:
quick-fix & maybe-useful info.


cclaudiu81:
glad you like the actual correction :)
In regards to my last comment, in my experience whenever i used HOFs was when i simply avoided to "lock down" (hardcode) the mechanism/operation inside the function itself, and for building more abstract constructions this was a good place to start...This approach helped me successfully to avoid inheritance when one can override & provide it's own version of impl, by simply using the concept of passing functions and use composition instead. From this the mouthful idea of "abstract mechanisms".. Again you can change it to whatever you consider it's worth adding value to this great tutorial!

SethTisue:
@cclaudiu81 there isn't anyone who's job it is to take pull requests and put them into final form. if you're interested in responding to the review feedback yourself, great. if not, we'll close the PR, while inviting other volunteers to submit their own versions if they like.

PR #1354, opened 2019-06-08
Fix translations of class for zh-cn

llinjupt:
Fix some wrong translations of class for zh-cn.


changlinli:
FWIW I can vouch that the original translation was incorrect and the new translation fixes it. The first line originally meant "the name of the class should be written in capitals" (i.e. all of the letters) which is now "the first letter of the class should be capitalized." The second line originally said "Using a 'private' modifier allows for hiding them outside the _function_" and is now "Using a 'private' modifier allows for hiding the outside the _class_."

liufengyun:
Thanks for your contribution @llinjupt 🎉 and thanks for review @changlinli 👍

PR #1355, opened 2019-06-08
Fix typos

asakaev:


SethTisue:
LGTM except possibly for the apostrophe change. in typography, it's standard practice to prefer a "curly" apostrophe and never use the "straight" one (see e.g. https://typographyforlawyers.com/straight-and-curly-quotes.html), so that part seems like a step backwards to me.

dwijnand:
Similarly, the change from em-dash to hyphen-minus is a regression in my eyes.

asakaev:
@SethTisue Thanks for the English typography. I did not know about apostrophes, a very interesting book.
@dwijnand It was not hyphen-minus but em-dash symbol, but I will do it in the next PR.

ashawley:
Looking at the HTML output from Jekyll, it seems it gets the typography folks desire here when you use the ASCII straight apostrophe, double quotes, or double hyphens for em dash.

PR #1357, opened 2019-06-10
Scala 2.13 release

SethTisue:


SethTisue:
(intended for merge sometime Tuesday)

SethTisue:
I got rid of `scala-213-version`, but it occurs to me that I probably need to add `scala-212-version` now, so the API links in the 2.12 collection guides remain correct

SethTisue:
missing piece, volunteer? https://github.com/scala/docs.scala-lang/issues/1365

PR #1388, opened 2019-06-21
Update breakOut.md fix type error

mumutu66:
the function should be (Int, String) => String
fix the B type error


dwijnand:
Thanks!

PR #1397, opened 2019-06-24
Add 2.13.0 to the JDK compatibility table

dwijnand:


Philippus:
Probably the "JDK 11 compatibility notes" require an update as well.

dwijnand:
I had a look at it but didn't see anything.  What did you have in mind?

Philippus:
The "As of" part doesn't mention 2.13.0 which could be misinterpreted I think.

dwijnand:
Done.

PR #1407, opened 2019-06-28
Translate "Partial application" part into russian

a-ostretsova:
Hello team, I don't know if any additional details needed for this PR. I just noticed that this part is missing from Russian translation. Let me know if you need anything else from me.

Thanks,
Ana.


kotobotov:
this paragraph was moved to the very top
but look's like lost the exemples

a-ostretsova:
Ah yeah, it is a part of the beginning. Would you like me to close the PR or do some changes? @kotobotov

kotobotov:
> Ah yeah, it is a part of the beginning. Would you like me to close the PR or do some changes? @kotobotov
I think it's a good idea to duplicate the example anyway.

PR #1418, opened 2019-07-04
Tweak literal types spec part about asInstanceOf

dwijnand:
Feedback welcome.


sjrd:
Should we *add* some dedicated note about `asInstanceOf`, maybe? In addition to removing the existing incorrect wording.

PR #1426, opened 2019-07-11
[ja] Let match print messages to output comments

shuuji3:
Comments below the code seem to be the output of the corresponding codes.
This PR fixes this the mismatch between the codes and comments


shuuji3:
@eed3si9n Thanks for merging!

PR #1430, opened 2019-07-11
Update type-inference.md

parambirs:
Remove semicolon


Philippus:
the same change could be done for the translated tours

PR #1432, opened 2019-07-12
Remove wrong information related to changes in float tokenisation.

johnkjellberg:
`1.` is not a valid token, therefore the parentheses is not needed.


Philippus:
Nice find! Would you mind making this change for the translated tours as well to keep them more in sync?

johnkjellberg:
> Nice find! Would you mind making this change for the translated tours as well to keep them more in sync?

I did the changes to all languages now. It seemed to have failed for some unknown reason though. A glitch?

SethTisue:
thank you!

PR #1433, opened 2019-07-14
add space in function call in pattern matching page

puorc:


Philippus:
Thanks! Would you mind applying the same change to the translated tours?

puorc:
Hi I have changed pages for other langs

Issue #1421, opened 2019-07-07
PR validation failing with new round of 404s


SethTisue:
seen recently at e.g. https://ci.scala-lang.org/scala/docs.scala-lang/189/3

```
- ./_site/ko/tour/annotations.html
366s
937
  *  External link https://www.scala-lang.org/api/current/scala/remote.html failed: 404 No error
366s
938
- ./_site/overviews/collections/trait-iterable.html
366s
939
  *  External link https://www.scala-lang.org/api/current/scala/collection/GenSetLike.html failed: 404 No error
366s
940
- ./_site/overviews/collections/trait-traversable.html
366s
941
  *  External link http://www.scala-lang.org/api/current/scala/collection/TraversableOnce.html failed: 404 No error
366s
942
- ./_site/sips/inline-meta.html
366s
943
  *  External link http://gitter.im/scalameta/sips failed: 404 No error
```

dongxuwang:
The link is 502 now.

Philippus:
```
- ./_site/overviews/scaladoc/generate.html
  *  External link https://www.scala-lang.org/files/archive/nightly/docs/manual/html/scaladoc.html failed: 404 No error
```

dwijnand:
I don't understand why these keep coming up...

exoego:
https://www.scala-lang.org/files/archive/nightly/docs/ is empty

SethTisue:
it became empty because https://github.com/scala/scala-lang/issues/453#issuecomment-510458550

PR #1436, opened 2019-07-16
Make the landing page translatable and add its Japanese translation

### LINKED ISSUE #1324: Localized landing pages and UX
The Scala Documentation site contains many localized contents that various language community have put in efforts over the years. For example, [the entire collections docs](https://docs.scala-lang.org/ja/overviews/collections/introduction.html) are translated into Japanese, Tour of Scala lists nine languages including Spanish, Standard Chinese, Korean, etc.

It would be nice if we can create a landing page for each language, similar to what Wikipedia has https://es.wikipedia.org/wiki/Wikipedia:Portada, and link to the available localized contents.

Prior to 2017 redesign, there used to be a link to localized overview page from the English overview page:
- https://web.archive.org/web/20170611072332/http://docs.scala-lang.org/ja/overviews

I think having top-level localized landing pages would motivate the various language communities to fill the gaps, and over time improve the usability by non-English-speaking learners of Scala.

exoego:
Closes #1324
Cc: @eed3si9n 

* Add a language dropdown to [the landing page](https://docs.scala-lang.org/) 
* Add Japanese translation of landing page.  Some links guides reader to Japanese translation if available, otherwise guide to English pages.
japanese-landing-page

Unfortunately, current template is not considered to have the seach bar and language dropdown at same time.
I gave a try but it is not easy to add language dropdown (both from design POV and coding POV).
image

Honestly, I don't want to contribute my time into *redesign*-ish issues.
Because I am afraid that PRs get rejected [because those who responsible for docs prohibit community to get involved in redesign](https://github.com/scala/scala-lang/pull/891#issuecomment-386555832).

However, because I beleive #1324 is a key to improve UX for non-English learners, I propose this change.
Even though a design incosistency, I think this is a good start.
Allowing community to translate langing page will improves UX very much.
If needed, those who are resnponsible for docs may consider hiring a professional to redesign in later.

### DIFF
--- a/_includes/masthead-documentation.html
+++ b/_includes/masthead-documentation.html
@@ -1,6 +1,34 @@
 <section class="table-of-content">
 	<div class="wrap">
 		<div class="inner-box">
+
+			<div class="language-dropdown inverted">
+				<div id="dd" class="wrapper-dropdown" tabindex="1">
+					<span>Language</span>
+					<ul class="dropdown"></ul>
+				</div>
+			</div>
+
+			<ul id="available-languages" style="display: none;">
+				{% if page.languages %}
+					<li><a href="{{ site.baseurl }}{{ page.url }}">English</a></li>
+					{% for l in page.languages %}
+						{% capture intermediate %}{{ page.url }}{% endcapture %}
+						{% capture rootTutorialURL %}{{ intermediate | remove_first: '/' }}{% endcapture %}
+						{% assign lang = site.data.languages[l] %}
+						<li><a href="{{ site.baseurl }}/{{ l }}/{{ rootTutorialURL }}" class="lang">{{ lang.name }}</a></li>
+					{% endfor %}
+				{% elsif page.language %}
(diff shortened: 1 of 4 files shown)


exoego:
404 check fils due to 
https://github.com/scala/docs.scala-lang/issues/1421#issuecomment-511481361

eed3si9n:
Thanks for this!

My personal opinion is that once these landing pages become popular, the language bar would become less important since we can navigate within the language. So propose we place the language bar below the main contents.

On the English page (https://docs.scala-lang.org/), we could make language navigation an actual text, similar to how it currently shows " More Resources: Online Courses, Exercises, & Blogs Books".

exoego:
> 
> 
> I have a suggestion for the placement of the language bar (to be below the contents), but if the build passes, I'd be happy to merge this as-is, and work on more details later.

Thanks for suggestion!!
I think the language dropdown should be shown in the front of content, a.k.a *first view* or *above-the-fold*.
So that non-English-native users find the dropdown without scrolling.

PR #1440, opened 2019-07-18
Support “next” and “previous” links between pages of overviews

julienrf:
Pages of a multipage-overview can now have a `previous-page` property
or a `next-page` property, which generate a “previous” and a “next”
link, respectively, like it is currently done in the Tour.

Also, update the collections-2.13 overview to make use of this feature.

Example:
image


SethTisue:
CI failure is unrelated

does this need review, or should we just go ahead and merge it?

julienrf:
I think we can just merge it.

PR #1435, opened 2019-07-15
Update compiler-options for Scala 2.13.x

exoego:
Closes #1381

```bash
git clone https://github.com/scala/scala.git
cd scala/
sbt generateDocsData
```

PR #1442, opened 2019-07-18
Fix compiler options

exoego:
Scala compiler options for 2.13 added in https://github.com/scala/docs.scala-lang/pull/1435 is broken.
Unfortunately, CI (Drone) did not run properly in the PR somehow, so gave false-success and accidentally merged.

This PR fixes
* Wrong indentation in `choices` properties.
* Some unescaped chars `\u` and `"` in string

I then fix `generateDocsData` command in `scala/scala` repository to reflect these changes.


exoego:
Opened a bugfix PR in https://github.com/scala/scala/pull/8247

PR #1445, opened 2019-07-19
Naming convention exception generating/deriving formats

### LINKED ISSUE #1439: Style guide naming conventions doesn't cover formats / protocolls
The Naming Conventions on the Style guide does not cover formats or protocols. It says:
> Underscores in names `(_)` are not actually forbidden by the compiler, but are strongly discouraged as they have special meaning within the Scala syntax. (But see below for exceptions.)

What I mean by that are field names in classes that have the sole purpose of describing a format and are used to dervice/generate encoders and decoders and being seperate from the case class that is actually used within the rest of the code (using camelCase fields). These format-classes are usually used to separate domain classes from multiple possible formats or ways of serializations, especially when there is no control over the format. The naming conventions should cover this.

A concrete example using circe would be a something like:
```
case class UserData(
   user_id: Long,
   name: String,
   lastUpdatedTimestamp: Long
)
object UserData {
   implicit val userDataDecoder: Decoder[UserData] = io.circe.generic.semiauto.deriveDecoder
   def convert(data: UserData): domain.User = ???
}
```

valenterry:
Added the exceptions for types that have the sole purpose of describing and/or generating formats, typeclass instances etc.

Hope this is a good location for that addition, please feel free to improve. :)

Closes #1439

### DIFF
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -40,6 +40,12 @@ Classes should be named in upper camel case:
 
 This mimics the Java naming convention for classes.
 
+Sometimes traits and classes as well as their members are used to describe
+formats, documentation or protocols and generate/derive them. 
+In these cases it is desirable to be close to a 1:1 relation to the output format
+and the naming conventions don't apply. In this case, they should only be used
+for that specific purpose and not throughout the rest of the code.
+
 ## Objects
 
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM thanks!

PR #1450, opened 2019-07-20
Remove the redundant 'val' and 'new' keywords

### LINKED ISSUE #1443: Redundant `new` and `val` keywords usage in the 'Operators' section of the Scala Tour
From the [Operators](https://docs.scala-lang.org/tour/operators.html):

```scala
case class Vec(val x: Double, val y: Double) {
  def +(that: Vec) = new Vec(this.x + that.x, this.y + that.y)
}
```

Since case classes do not require the `val` keyword for parameters of case class primary constructor and do not require the `new` keyword to create an instance of a case class, the example above could be simplified as follows:

```scala
case class Vec(x: Double, y: Double) {
  def +(that: Vec) = Vec(this.x + that.x, this.y + that.y)
}
```

maseev:
Fixes #1443


SethTisue:
great, thanks for doing the translations too!

PR #1451, opened 2019-07-20
Update Style Guide "Invocation" section and TOC

### LINKED ISSUE #1431: Style Docs Missing Arity-1 and Higher-order Functions sections
On the style guide, the Table of contents includes an entry for Arity-1 and Higher-Order Functions under "Method invocation": https://docs.scala-lang.org/style/

However the linked page (https://docs.scala-lang.org/style/method-invocation.html) doesn't actually contain these sections.

Were they forgotten, or no longer relevant?

flakeparadigm:
Fixes #1431 

This also makes sure that the TOC and the index page are in the matching order, and all of the section links in the index page should now also deep link to their section properly.

The new content is a slight reorganization of the existing text as well as adding the missing entries from one of the sources of the text for the style guide. The added content seems like it might have been accidentally left out when it was originally created (the page only has 2 commits in the history).

http://www.codecommit.com/scala-style-guide.pdf


SethTisue:
awesome, thank you!

PR #1456, opened 2019-07-22
Add a newer version notice in Collections (Legacy)

### LINKED ISSUE #1121: The architecture of scala collections should have a link to the 2.13 sister page
This page: https://docs.scala-lang.org/overviews/core/architecture-of-scala-collections.html
Should link to that page: https://docs.scala-lang.org/overviews/core/architecture-of-scala-213-collections.html#inner-main

Something like a warning on the top of the old documentation that you should be looking at the 2.13.x page.

exoego:
Fixes #1121 

There is 2 version of  **"Overviews => Collections"**: one is for Scala 2.13, and the other is for legacy (Scala 2.12 and older).
I think it requires a lot attention to distinguish both of them, since page header are identical.

This PR adds 

- `(Scala 2.8 - 2.12)` to sub title
- A link to newer version of corresponding page

so readers (and search crawlers) know it is old and there is new version.

link-to-newer

### DIFF
--- a/_layouts/inner-page-parent-dropdown.html
+++ b/_layouts/inner-page-parent-dropdown.html
@@ -1,5 +1,6 @@
 {% include headertop.html %}
 {% include headerbottom.html %}
+{% if page.new-version %}<a class="new-version-notice" href="{{ page.new-version }}">This page has a new version.</a>{% endif %}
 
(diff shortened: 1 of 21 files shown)


SethTisue:
fixes #1121

PR #1461, opened 2019-07-24
Small cheatsheet fixes

toy:
A lot of added and removed whitespace, added dots, capitalisation + few slightly changed and one fixed example.

### DIFF
--- a/_cheatsheets/index.md
+++ b/_cheatsheets/index.md
@@ -481,8 +481,9 @@ println(upper.getOrElse(""))</code></pre></td>
       <pre class="highlight"><code>option match {
   case Some(x) =&gt; f(x)
   case None    =&gt; ()
-}</code></pre></td>
-      <td>Apply a procedure on optional value</td>
+}
+()</code></pre></td>
+      <td>Apply a procedure on optional value. Returns <code>Unit</code></td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>option.fold(y)(f(_))</code></pre>
--- a/_cheatsheets/index.md
+++ b/_cheatsheets/index.md
@@ -470,8 +470,8 @@ println(upper.getOrElse(""))</code></pre></td>
       <td><pre class="highlight"><code>optionOfOption.flatten</code></pre>
       <em><strong>same as</strong></em>
       <pre class="highlight"><code>optionOfOption match {
-  case Some(Some(x)) =&gt; Some(x)
-  case _             =&gt; None
+  case Some(x: Option[_]) =&gt; x
+  case None               =&gt; None
 }</code></pre></td>
       <td>Extract nested option.</td>
     </tr>
@@ -502,8 +502,7 @@ println(upper.getOrElse(""))</code></pre></td>
       <pre class="highlight"><code>option match {
   case Some(x)
     if f.isDefinedAt(x) =&gt; ...
-  case Some(_)          =&gt; None
-  case None             =&gt; None
+  case _                =&gt; None
 }</code></pre></td>
       <td>Apply partial pattern match on optional value</td>
     </tr>
--- a/_cheatsheets/index.md
+++ b/_cheatsheets/index.md
@@ -25,16 +25,16 @@ languages: [ba, fr, ja, pl, pt-br, zh-cn, th, ru]
       <td> </td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>var x = 5</code></pre><br /> <span class="label success">Good</span><br><pre class="highlight"><code>x=6</code></pre></td>
-      <td>variable</td>
+      <td><pre class="highlight"><code>var x = 5</code></pre><br /> <span class="label success">Good</span><br><pre class="highlight"><code>x = 6</code></pre></td>
+      <td>variable.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>val x = 5</code></pre><br /> <span class="label important">Bad</span><br><pre class="highlight"><code>x=6</code></pre></td>
-      <td>constant</td>
+      <td><pre class="highlight"><code>val x = 5</code></pre><br /> <span class="label important">Bad</span><br><pre class="highlight"><code>x = 6</code></pre></td>
+      <td>constant.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>var x: Double = 5</code></pre></td>
-      <td>explicit type</td>
+      <td>explicit type.</td>
     </tr>
     <tr>
       <td><span id="functions" class="h2">functions</span></td>
@@ -57,15 +57,15 @@ languages: [ba, fr, ja, pl, pt-br, zh-cn, th, ru]
       <td>call-by-value <br /> call-by-name (lazy parameters)</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>(x:R) =&gt; x * x</code></pre></td>
+      <td><pre class="highlight"><code>(x: R) =&gt; x * x</code></pre></td>
       <td>anonymous function</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>(1 to 5).map(_ * 2)</code></pre> vs.<br /> <pre class="highlight"><code>(1 to 5).reduceLeft( _ + _ )</code></pre></td>
+      <td><pre class="highlight"><code>(1 to 5).map(_ * 2)</code></pre> vs.<br /> <pre class="highlight"><code>(1 to 5).reduceLeft(_ + _)</code></pre></td>
       <td>anonymous function: underscore is positionally matched arg.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>(1 to 5).map( x =&gt; x * x )</code></pre></td>
+      <td><pre class="highlight"><code>(1 to 5).map(x =&gt; x * x)</code></pre></td>
       <td>anonymous function: to use an arg twice, have to name it.</td>
     </tr>
     <tr>
@@ -94,15 +94,15 @@ languages: [ba, fr, ja, pl, pt-br, zh-cn, th, ru]
       <td>anonymous functions: to pass in multiple blocks, need outer parens.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>val zscore = 
+      <td><pre class="highlight"><code>val zscore =
   (mean: R, sd: R) =&gt;
-    (x:R) =&gt; 
+    (x:R) =&gt;
       (x - mean) / sd</code></pre></td>
       <td>currying, obvious syntax.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>def zscore(mean:R, sd:R) =
-  (x:R) =&gt; 
+  (x:R) =&gt;
     (x - mean) / sd</code></pre></td>
       <td>currying, obvious syntax</td>
     </tr>
@@ -126,7 +126,7 @@ languages: [ba, fr, ja, pl, pt-br, zh-cn, th, ru]
       <td>infix sugar.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>def sum(args: Int*) = 
+      <td><pre class="highlight"><code>def sum(args: Int*) =
   args.reduceLeft(_+_)</code></pre></td>
       <td>varargs.</td>
     </tr>
@@ -209,7 +209,7 @@ languages: [ba, fr, ja, pl, pt-br, zh-cn, th, ru]
       <td>conditional sugar.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>while (x &lt; 5) { 
+      <td><pre class="highlight"><code>while (x &lt; 5) {
   println(x)
   x += 1
 }</code></pre></td>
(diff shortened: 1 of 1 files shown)


ashawley:
Thanks for doing this.  I fixed a lot of formatting in #1280, but knew it still needed more work.

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,271 @@ permalink: /cheatsheets/index.html
...
+      <td>Explicit type.</td>
     </tr>
     <tr>
       <td><span id="functions" class="h2">functions</span></td>
       <td> </td>
     </tr>
     <tr>
       <td><span class="label success">Good</span><br> <pre class="highlight"><code>def f(x: Int) = { x * x }</code></pre><br /> <span class="label important">Bad</span><br> <pre class="highlight"><code>def f(x: Int)   { x * x }</code></pre></td>
-      <td>define function <br /> hidden error: without = it’s a Unit-returning procedure; causes havoc</td>
+      <td>Define function.<br />Hidden error: without <code>=</code> it’s a <code>Unit</code>-returning procedure; causes havoc.</td>
```
ashawley: The site's CSS for `<code>` is adding weird horizontal whitespace so it looks strange in some of these places you added it, e.g. "`Unit `-recording".

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,271 @@ permalink: /cheatsheets/index.html
...
     <tr>
-      <td><pre class="highlight"><code>(1 to 5).map( x =&gt; x * x )</code></pre></td>
-      <td>anonymous function: to use an arg twice, have to name it.</td>
+      <td><pre class="highlight"><code>(1 to 5).map(x =&gt; x * x)</code></pre></td>
+      <td>Anonymous function: to use an arg twice, have to name it.</td>
     </tr>
     <tr>
       <td><span class="label success">Good</span><br> <pre class="highlight"><code>(1 to 5).map(2 *)</code></pre><br /> <span class="label important">Bad</span><br> <pre class="highlight"><code>(1 to 5).map(* 2)</code></pre></td>
-      <td>anonymous function: bound infix method.<br /> Use <code class="highlighter-rouge">2 * _</code> for sanity’s sake instead.</td>
+      <td>Anonymous function: bound infix method.<br /> Use <code class="highlighter-rouge">2 * _</code> for sanity’s sake instead.<br />Will require <code>import scala.language.postfixOps</code> in the future.</td>
```
ashawley: In light of https://github.com/scala/scala/pull/6559 maybe we just delete this row from the page?  It doesn't seem relevant for a cheatsheet, and post fix ops are on their way out.

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,271 @@ permalink: /cheatsheets/index.html
...
   (x: R) =&gt; g(h(x))</code></pre> <br /> <pre class="highlight"><code>val f = compose(_ * 2, _ - 1)</code></pre></td>
-      <td>anonymous functions: to pass in multiple blocks, need outer parens.</td>
+      <td>Anonymous functions: to pass in multiple blocks, need outer parens.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>val zscore = 
+      <td><pre class="highlight"><code>val zscore =
   (mean: R, sd: R) =&gt;
-    (x:R) =&gt; 
+    (x:R) =&gt;
```
ashawley: Like elsewhere, add a space after the colon in the type annotation?

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,271 @@ permalink: /cheatsheets/index.html
...
     <tr>
-      <td><pre class="highlight"><code>for (x &lt;- xs if x%2 == 0)
-yield x * 10</code></pre>
+      <td><pre class="highlight"><code>for (x &lt;- xs if x % 2 == 0)
+  yield x * 10</code></pre>
       <br><em><strong>same as</strong></em><br>
-      <pre class="highlight"><code>xs.filter(_%2 == 0).map( _ * 10)</code></pre></td>
-      <td>for comprehension: filter/map</td>
+      <pre class="highlight"><code>xs.filter(_ % 2 == 0).map(_ * 10)</code></pre></td>
+      <td>For comprehension: filter/map.</td>
```
ashawley: Maybe "For-comprehension" to be consistent elsewhere?

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,271 @@ permalink: /cheatsheets/index.html
...
+      <td>For-comprehension: cross product.</td>
     </tr>
     <tr>
-      <td><pre class="highlight"><code>for (x <- xs; y <- ys) {
+      <td><pre class="highlight"><code>for (x &lt;- xs; y &lt;- ys) {
   val div = x / y.toFloat
   println("%d/%d = %.1f".format(x, y, div))
 }</code></pre></td>
-      <td>for comprehension: imperative-ish<br /><a href="http://java.sun.com/javase/6/docs/api/java/util/Formatter.html#syntax">sprintf-style</a></td>
+      <td>For-comprehension: imperative-ish.<br /><a href="http://java.sun.com/javase/6/docs/api/java/util/Formatter.html#syntax">Sprintf-style</a>.</td>
```
ashawley: This looks strange.  Maybe keep `sprintf` as lowercase, since it's a function name?

_overviews/cheatsheets/index.md
toy: Agree, this one looked bad, fixed it and overall the whitespace makes it a bit weird, but I still think it is better to use `<code>`.

_overviews/cheatsheets/index.md
toy: Then with `<code>`?

toy:
@ashawley Please have a look once more
Seems I can't re-request review :)

_overviews/cheatsheets/index.md
ashawley: I agree.  It should be possible to use `<code>`.  At some point we need to fix the CSS for `<code>`.

_overviews/cheatsheets/index.md
```diff
@@ -405,16 +402,18 @@ yield x * y</code></pre>
...
+      <td>The singleton empty optional value.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>Option(null) == None
-Option(obj.unsafeMethod)</code></pre></td>
-      <td>Null-safe optional value factory</td>
+Option(obj.unsafeMethod)</code></pre>
+      <em><strong>but</strong></em>
+      <pre class="highlight"><code>Some(null) != None</code></pre></td>
+      <td>Null-safe optional value factory.</td>
```
ashawley: Is `Some(null)` useful to mention in a cheatsheet?  Seems like a corner case.  For most types of `Option`, it's not possible to encode `Some(null)`.  The one exception is `Option[String]`, since it is nullable.  With a type signature for any other `Option` (function parameter, return type, class member, variable annotation, etc.), Scala won't compile if a user tries to assign or pass `Some(null)` to it.

_overviews/cheatsheets/index.md
```diff
@@ -464,25 +460,26 @@ println(upper.getOrElse(""))</code></pre></td>
...
+      <td>Same as map but function must return an optional value.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>optionOfOption.flatten</code></pre>
       <em><strong>same as</strong></em>
       <pre class="highlight"><code>optionOfOption match {
-  case Some(Some(x)) =&gt; Some(x)
-  case _             =&gt; None
+  case Some(x: Option[_]) =&gt; x
+  case None               =&gt; None
```
ashawley: This alternate pattern match for `flatten` is also true, but I predict the original was better at showing what `flatten` does.

_overviews/cheatsheets/index.md
```diff
@@ -464,25 +460,26 @@ println(upper.getOrElse(""))</code></pre></td>
...
     <tr>
       <td><pre class="highlight"><code>option.foreach(f(_))</code></pre>
       <em><strong>same as</strong></em>
       <pre class="highlight"><code>option match {
   case Some(x) =&gt; f(x)
   case None    =&gt; ()
-}</code></pre></td>
-      <td>Apply a procedure on optional value</td>
+}
+()</code></pre></td>
```
ashawley: Although accurate, putting `()` unit literal after the pattern match seems clumsy.  An annotation to the first pattern match case could be added:

```scala
  case Some(x) => f(x): Unit
```

However, I'm not sure if such precision is necessary on a cheatsheet.

_overviews/cheatsheets/index.md
```diff
@@ -501,10 +498,9 @@ println(upper.getOrElse(""))</code></pre></td>
       <pre class="highlight"><code>option match {
   case Some(x)
     if f.isDefinedAt(x) =&gt; ...
-  case Some(_)          =&gt; None
-  case None             =&gt; None
+  case _                =&gt; None
```
ashawley: Yes, it could be minimized with the wildcard pattern match, but the extra pattern match cases helped explain `collect`, IMO.

_overviews/cheatsheets/index.md
```diff
@@ -594,25 +590,25 @@ println(upper.getOrElse(""))</code></pre></td>
...
+      <td>Optional value doesn't satisfy predicate.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>option.exists(f(_))</code></pre>
       <em><strong>same as</strong></em>
       <pre class="highlight"><code>option match {
-  case Some(x) if f(x) =&gt; true
-  case _               =&gt; false
+  case Some(x) =&gt; f(x)
+  case None    =&gt; false
```
ashawley: This is also a true implementation of `exists`.  There's no reason to have the boolean guard condition.  However, I think the longer version helps with readability.

_overviews/cheatsheets/index.md
```diff
@@ -594,25 +590,25 @@ println(upper.getOrElse(""))</code></pre></td>
...
+      <td>Apply predicate on optional value or <code>false</code> if empty.</td>
     </tr>
     <tr>
       <td><pre class="highlight"><code>option.forall(f(_))</code></pre>
       <em><strong>same as</strong></em>
       <pre class="highlight"><code>option match {
-  case Some(x) if f(x) =&gt; true
-  case None            =&gt; false
+  case Some(x) =&gt; f(x)
+  case None    =&gt; true
```
ashawley: Good catch on the bugs with `forall` here!

Again, the code golf version is shorter, but I'm not convinced removing the guard helps.

_overviews/cheatsheets/index.md
toy: My point was to show that one should be careful not to use `Some(obj.unsafeMethod)` and use `Option(obj.unsafeMethod)` instead. Option of any ref type can have `Some(null)` assigned to it.
Do you think it will be more confusing to include it?

_overviews/cheatsheets/index.md
toy: I also don't like the added line in the code, but I think it is probably better to leave a note in description that return value is `Unit`, so one can't chain calls like `o.foreach(…).foreach(…)`

_overviews/cheatsheets/index.md
```diff
@@ -26,270 +26,267 @@ permalink: /cheatsheets/index.html
...
+      <td>Explicit type.</td>
     </tr>
     <tr>
       <td><span id="functions" class="h2">functions</span></td>
       <td> </td>
     </tr>
     <tr>
       <td><span class="label success">Good</span><br> <pre class="highlight"><code>def f(x: Int) = { x * x }</code></pre><br /> <span class="label important">Bad</span><br> <pre class="highlight"><code>def f(x: Int)   { x * x }</code></pre></td>
-      <td>define function <br /> hidden error: without = it’s a Unit-returning procedure; causes havoc</td>
+      <td>Define function.<br />Hidden error: without <code>=</code> it’s a procedure returning <code>Unit</code>; causes havoc.</td>
```
julienrf: Side note: procedure syntax has been deprecated in 2.13.0.

_overviews/cheatsheets/index.md
ashawley: Ok, good point about `AnyRef`.   I guess it is common.

_overviews/cheatsheets/index.md
ashawley: Sure, but fortunately the compiler will complain if someone tries to chain the calls.

toy: Good point. So you think I should remove the note from description?

_overviews/cheatsheets/index.md
toy: I've added a note

toy:
@ashawley @julienrf Please have a look once more

_overviews/cheatsheets/index.md
ashawley: No, I don't think it's necessary.  The term "procedure" is meant to connote that it returns `Unit`.

toy: True, removed the note

ashawley:
Looks good to me.  Thanks again for doing this.

SethTisue:
needs rebase

toy:
@SethTisue Done

SethTisue:
awesome, thanks for this!

PR #1465, opened 2019-07-29
Update tuples.md

Deepanshu8391:
Redundant 's' in line #67


julienrf:
Hello, thanks for your effort but actually this `s` is *required*. See https://docs.scala-lang.org/overviews/core/string-interpolation.html for more information. Consequently, I’m closing the PR.

PR #1467, opened 2019-07-30
Make index page of "Overviews" translatable again and add its Japanese translation

### LINKED ISSUE #938: Inconsistent style of doc sites in different language
Now our style of https://docs.scala-lang.org/zh-cn/overviews/ is different from that of https://docs.scala-lang.org/overviews/ and of https://docs.scala-lang.org/ja/overviews/ , which two are in the same.

I wonder the reason why it arrives here, and whether we're supposed to unify the style?

@heathermiller (wish i don't \at a wrong member)

exoego:
Closes #938 
cc: @eed3si9n @dongxuwang 

The [index page of "Overviews"](https://docs.scala-lang.org/overviews/index.html) had been redesigned several years ago, without consideration for translation.

This PR makes the redesigned `overviews.html` template works for other languages.

All actions contributors need to do are
1. Add their `lang` code in `languages` property of `/_overviews/index.md`
2. Add `/_data/overview-{lang}.yml` based on `/_data/overview.yml`
     * Just keep `url`s in the file as same as English version. `overviews.html` will generate automatically a link: if the page has `{lang}` translated version, the generated link will be a translated version. Otherwise English version.
    

See how it works in Japanese.
japanese-translation-of-overviews-index


eed3si9n:
In the screen capture, I see the chapter names of Collection overview in English, like Arrays. Is that correct?

exoego:
> 
> 
> In the screen capture, I see the chapter names of Collection overview in English, like Arrays. Is that correct?

Forgotten to translte.
Fixed in https://github.com/scala/docs.scala-lang/pull/1467/commits/b00eb9f6ccbf98a4db4795fbe9a813c4e1ba2bae

PR #1468, opened 2019-07-30
Add Twitter card/OGProtocol for sharing pages on SNS

exoego:
Add Twitter card & OG Protocol (for Facebook or other SNS) to docs.scala-lang.org.
Those are a rich preview of pages that may attract readers than ordinal text link, and easy to click on touch screen since bigger area.
Those are aledy included in [scala-lang repo](https://github.com/scala/scala-lang/blob/c63ea9892628a3e37ddcea6fa999132b8d42fff3/_includes/headertop.html)

image

### DIFF
--- a/_includes/headertop.html
+++ b/_includes/headertop.html
@@ -12,6 +12,19 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
+    <meta property="og:site_name" content="{{ site.title }}"/>
+    <meta property="og:type" content="{% if page.url == '/' %}website{% else %}article{% endif %}"/>
+    <meta property="og:url" content="{{ site.url }}{{ page.url }}"/>
+    <meta property="og:image" content="{{ site.url }}/resources/img/scala-spiral-3d-2-toned-down.png"/>
+    <meta property="og:title" content="{% if page.title %}{{ page.title }}{% else %}{{site.title}}{% endif %}"/>
+    {% if page.description %}<meta property="og:description" content="{{ page.description }}"/>{% endif %}
+
+    <meta name="twitter:card" content="summary"/>
+    <meta name="twitter:site" content="@scala_lang"/>
+    <meta name="twitter:creator" content="@scala_lang"/>
+    {% if page.title %}<meta property="twitter:title" content="{{ page.title }}"/>{% endif %}
+    {% if page.description %}<meta property="twitter:description" content="{{ page.description }}"/>{% endif %}
+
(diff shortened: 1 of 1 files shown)


jvican:
Thanks for this! @exoego LGTM! :smile:

_includes/headertop.html
```diff
@@ -12,6 +12,16 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
+    <meta property="og:site_name" content="{{ site.title }}"/>
+    {% if page.title %}<meta property="og:title" content="{{ page.title }}"/>{% endif %}
```
julienrf: The `og:title` property is **required** by the protocol, but we provide it only if the page has a title. This raises two questions:
- Do we have pages without title?
- What about using `site.title` as a fallback in such a case?

_includes/headertop.html
```diff
@@ -12,6 +12,16 @@
...
+    {% if page.description %}<meta property="og:description" content="{{ page.description }}"{% endif %}
```
julienrf: Should we also provide the `og:type`, `og:image`, and `og:url` properties? They are documented as **required**.

https://ogp.me/

_includes/headertop.html
exoego: `page.title` is Jekyll's.
Not sure it is required from https://jekyllrb.com/docs/variables/
Added fallback in https://github.com/scala/docs.scala-lang/pull/1468/commits/62ef7f0c76fad3917a0645585c87a776edf4a1d7

exoego:
Squashed commits.

_includes/headertop.html
```diff
@@ -12,6 +12,19 @@
...
+    <meta property="og:url" content="{{ site.url }}{{ page.url }}"/>
+    <meta property="og:image" content="{{ site.url }}/resources/img/scala-spiral-3d-2-toned-down.png"/>
+    <meta property="og:title" content="{% if page.title %}{{ page.title }}{% else %}{{site.title}}{% endif %}"/>
+    {% if page.description %}<meta property="og:description" content="{{ page.description }}"/>{% endif %}
+
+    <meta name="twitter:card" content="summary"/>
+    <meta name="twitter:site" content="@scala_lang"/>
+    <meta name="twitter:creator" content="@scala_lang"/>
+    {% if page.title %}<meta property="twitter:title" content="{{ page.title }}"/>{% endif %}
+    {% if page.description %}<meta property="twitter:description" content="{{ page.description }}"/>{% endif %}
```
julienrf: Actually, I think twitter markup needs to use a `name` attribute instead of `property`.

julienrf: Fixed by #1470

PR #1472, opened 2019-07-31
Update to links for version 2.13

komainu8:
These links point to old version contents.
Therefore, we updated to links for the latest version.


Philippus:
thank you @komainu8 !

PR #1474, opened 2019-08-01
Small note to describe Nil

manishpatelUK:
This is the first time the user comes across Nil in the tutorial, so a small explanation of what it is.

### DIFF
--- a/_tour/generic-classes.md
+++ b/_tour/generic-classes.md
@@ -31,7 +31,7 @@ class Stack[A] {
 ```
 This implementation of a `Stack` class takes any type `A` as a parameter. This means the underlying list, `var elements: List[A] = Nil`, can only store elements of type `A`. The procedure `def push` only accepts objects of type `A` (note: `elements = x :: elements` reassigns `elements` to a new list created by prepending `x` to the current `elements`).
 
-`Nil` here is a special case of `List` and is not to be confused with `Null`. It is simply an empty list.
+`Nil` here is an empty `List` and is not to be confused with `Null`.
 
 ## Usage
 
--- a/_tour/generic-classes.md
+++ b/_tour/generic-classes.md
@@ -31,7 +31,7 @@ class Stack[A] {
 ```
 This implementation of a `Stack` class takes any type `A` as a parameter. This means the underlying list, `var elements: List[A] = Nil`, can only store elements of type `A`. The procedure `def push` only accepts objects of type `A` (note: `elements = x :: elements` reassigns `elements` to a new list created by prepending `x` to the current `elements`).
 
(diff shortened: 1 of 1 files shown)


_tour/generic-classes.md
```diff
@@ -31,6 +31,8 @@ class Stack[A] {
 ```
 This implementation of a `Stack` class takes any type `A` as a parameter. This means the underlying list, `var elements: List[A] = Nil`, can only store elements of type `A`. The procedure `def push` only accepts objects of type `A` (note: `elements = x :: elements` reassigns `elements` to a new list created by prepending `x` to the current `elements`).
 
+`Nil` here is a special case of `List` and is not to be confused with `Null`. It is a `List` of zero length.
```
Philippus: Let's mention the term "the empty list".

manishpatelUK: Done

_tour/generic-classes.md
```diff
@@ -31,7 +31,7 @@ class Stack[A] {
 ```
 This implementation of a `Stack` class takes any type `A` as a parameter. This means the underlying list, `var elements: List[A] = Nil`, can only store elements of type `A`. The procedure `def push` only accepts objects of type `A` (note: `elements = x :: elements` reassigns `elements` to a new list created by prepending `x` to the current `elements`).
 
-`Nil` here is a special case of `List` and is not to be confused with `Null`. It is a `List` of zero length.
+`Nil` here is a special case of `List` and is not to be confused with `Null`. It is simply an empty list.
```
martijnhoekstra: Maybe just "`Nil` here is the empty `List` (not to be confused with `null`)." is enough here.

manishpatelUK: Same difference, IMO, but now changed

_tour/generic-classes.md
```diff
@@ -31,6 +31,8 @@ class Stack[A] {
 ```
 This implementation of a `Stack` class takes any type `A` as a parameter. This means the underlying list, `var elements: List[A] = Nil`, can only store elements of type `A`. The procedure `def push` only accepts objects of type `A` (note: `elements = x :: elements` reassigns `elements` to a new list created by prepending `x` to the current `elements`).
 
+`Nil` here is an empty `List` and is not to be confused with `Null`.
```
SethTisue: `Nil` is a term, not a type, so I think the term `null` would make more sense here, `Null` is the type.

Philippus:
```suggestion
`Nil` here is an empty `List` and is not to be confused with `null`.
```

SethTisue:
final fixup in bae05563ec3b36db888b63edee0d3c165e9f3d65

PR #1475, opened 2019-08-01
Use [LABEL][LINK-ID] style

komainu8:
It might be a good idea to put links in one by [LABEL][LINK-ID] style.
In my opinion, this style is a little bit easier to maintain.

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -7,9 +7,9 @@ title: Contribute
 
 ## A Place to Build Documentation Together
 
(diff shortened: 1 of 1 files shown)


SethTisue:
I suggest we not merge this, I think things are fine the way they are.

when someone writes a whole new page, or does a heavy rewrite of an existing page, I don't mind if they use either style, as they prefer.

I don't see value in flipping the style existing pages, just to do it.

komainu8:
@SethTisue Thank you for your comment.
I agree with you. So, I close this PR.

Issue #1459, opened 2019-07-23
all links to scala-lang.org should be https not http


SethTisue:
as noticed by @Philippus at https://github.com/scala/docs.scala-lang/pull/1458#pullrequestreview-265694532

SethTisue:
fixed by #1459

SethTisue:
#1471 fixed it repo-wide

PR #1471, opened 2019-07-31
Use https instead of http

komainu8:
Dear, developers.
I've fixed an issue #1459.

PR #1476, opened 2019-08-01
Use https instead of http

komainu8:
Dear, developers.
I've fixed an issue #1459.

### DIFF
--- a/learn.md
+++ b/learn.md
@@ -37,7 +37,7 @@ the paid version delivers a certificate of completion. Learn more about
 
 ## Dr.Mark C Lewis's Lectures from Trinity University
 
- * [Dr.Mark C Lewis](http://www.cs.trinity.edu/~mlewis/) from Trinity University, San Antonio, TX, teaches programming courses using the Scala language. Course videos are available in YouTube for free. Some of the courses below.
+ * [Dr.Mark C Lewis](https://www.cs.trinity.edu/~mlewis/) from Trinity University, San Antonio, TX, teaches programming courses using the Scala language. Course videos are available in YouTube for free. Some of the courses below.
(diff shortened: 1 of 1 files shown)


exoego:
There is already your pull request https://github.com/scala/docs.scala-lang/pull/1471 addressing the issue #1459 
What is a reason to separate PRs?
I think an ordinal issue (not meta issue) should be addressed in single PR for most cases, so that reviewers close the issue after merge.

komainu8:
@exoego Thank you for your comment.
I agree with you. I think I should put those pull request in one.

However, files that should modify the same as this pull request is many.

I think that is difficult to review all modify at once.
Therefore, I split one pull request by the one file.

If we should put all modify in one is better, I'll do so.

I'm sorry, I'm bad at English.
Is what I'm saying getting through to you properly?

exoego:
Don't mind, it is clear what you are saying.

So, I guess you are afraid of broken links (404) that reviewers should take care, if replace all `http://` into `http://` at once ?
If so, please do not mind !!
There is a link checker in this repo, and verifies that all links is valid (with some exception for oracle.com) on pull requests.
If 404 found, you can find easily which links need some tweaks (e.g.:  there is no https://... for the domain).

So I suggest to
1. Replace all http:// to https:// and add all into single commit.
2. Push and watch link checker result.
3. If there are 404, fix them and create a new commit. So reviewers can spot only URLs that carefull review are required.

komainu8:
Oh! That's a great mechanism.
Yes. I was afraid of broken links (404) that reviewers should take care of.

Thank you for your suggestion.
I'll accept your suggestion.

First, I'll close this PR.
Second, I replace all http:// to https:// and add all modify into #1471.
Third, I check links and repair broken links.

PR #1477, opened 2019-08-01
Mention `Array.deep` has been removed

### LINKED ISSUE #1400: `.deep` method is gone in 2.13 but not mentioned in migration guide
missing from https://docs.scala-lang.org/overviews/core/collections-migration-213.html

reported by "Rogach" on #scala IRC

Philippus:
fixes #1400. I'm struggling with describing a good alternative though, but maybe that is not needed?


julienrf:
Maybe an alternative would be `.flatten.toSeq`?

SethTisue:
that works sometimes, but not always:

```
scala 2.12.8> Array(Array(1), Array("foo")).deep
res4: IndexedSeq[Any] = Array(Array(1), Array(foo))
```

```
scala 2.13.0> Array(Array(1), Array("foo")).flatten.toSeq
                                            ^
              error: No implicit view available from Array[_ >: String with Int] => scala.collection.IterableOnce[B].
```

SethTisue:
> Maybe an alternative would be `.flatten.toSeq`?

no, `.deep` doesn't flatten

I no longer remember what the original reporter of #1400 was using `deep` for

I think this whole topic is so obscure, it's fine to just merge this as-is

smarter:
I miss .deep too, there's a reimplementation in partest that I wish was part of the stdlib: https://github.com/scala/scala/blob/466990276fa361006a90a744616aac95321265a2/src/partest/scala/tools/partest/Util.scala#L65-L78

SethTisue:
@smarter can you `runtime.ScalaRunTime.stringOf` it, or do you actually need to stop before going all the way to `String`?

smarter:
I wasn't aware of that, that works too but afaik `scala.runtime` is considered an internal API and should not be relied on.

SethTisue:
yeah. perhaps a PR putting them stuff somewhere in stdlib, without adding to the number of methods available on `Array` and other types (which is the main reason I was glad to see `.deep` go), would be accepted.

PR #1487, opened 2019-08-11
SIP-NN - Curried varargs

Atry:
See discussion at https://contributors.scala-lang.org/t/pre-sip-curried-varargs/3608

### DIFF
--- a/_sips/sips/2019-08-11-curried-varargs.md
+++ b/_sips/sips/2019-08-11-curried-varargs.md
@@ -11,6 +11,7 @@ title: SIP-NN - Curried varargs
 | Date          | Version       |
 |---------------|---------------|
 | Aug 11th 2019 | Initial Draft |
+| Aug 12th 2019 | Translating sequence arguments to `applyNextSeq` calls |
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


_sips/sips/2019-08-11-curried-varargs.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Motivating Examples
+
+### Examples
+
+Recently I was working on the implementation of [Pre SIP: name based XML literals](https://contributors.scala-lang.org/t/pre-sip-name-based-xml-literals/2175). During implementing that proposal, I found that the proposal is inefficiency due to repeated parameters, and it could be improved dramatically with the help of curried functions.
+
+For example, according to the proposal the XML literal `<div title="my-title">line1<br/>line2</div>` will result the following code:
+
+{% highlight scala %}
```
NthPortal: any reason not to do the following?

```
`‌`‌`scala
```

Atry: According to https://github.com/scala/docs.scala-lang/blob/master/_sips/sip-template.md

NthPortal: oh, huh.

SethTisue: I don't think that applies anymore

lihaoyi:
I like this a lot:

- It is a natural solution to the “should varargs be mutable or inmutable or readonly Seq” question: with a builder, you can make your vararg take exactly the type you need, and have it be constructed directly without overhead

- It provides more information to the compiler and optimizer: where the number of args at a callsite is fixed, simple inlining produces a fixed set of instructions to run on a builder, which makes further optimization possible. The current Seq-based model obfuscates everything by forcing the params into a variable length sequence, which any optimizer would have to deobfuscate to figure out is fixed length

- The fact that the builder’s method calls can be overloaded can remove yet another layer of boxing, where previously we have had to take a supertype/trait and have the possible parameter types boxed into compliance with the since vararg type

- It has plenty of precedence: it’s just the Visitor Pattern or Finally Tagless. Like how uPickle uses visitors to abstract over multiple ASTs and JSON libraries with a performance *boost* instead of a penalty, this would let varargs abstract over multiple possible collection types, and even non collection types (e.g. Scalatags builders), also with a performance boost

- The fact that it could allow a heterogenous set of types to be inferred from a variadic argument list opens up a lot of possibilities, e.g. it makes translation of `Tuple.apply(x, y, z)` calls into HList-esque tuples trivial to implement in userland, a nice `Rec(“foo” -> bar)` syntax for record-types also entirely in userland, and probably other things that I can’t come up with off the top of my head

It’s not common that a new technique allows both better abstraction, better flexibility, *and* better performance. This definitely seems like one of those cases.

valenterry:
> * The fact that it could allow a heterogenous set of types to be inferred from a variadic argument list opens up a lot of possibilities, e.g. it makes translation of `Tuple.apply(x, y, z)` calls into HList-esque tuples trivial to implement in userland, a nice `Rec(“foo” -> bar)` syntax for record-types also entirely in userland, and probably other things that I can’t come up with off the top of my head

I agree! This would be huge for making dependently typed apis much more user friendy.

PR #1490, opened 2019-08-15
Fix #1485 (en), add : Unit = for the deprecated procedure snippet.

### LINKED ISSUE #1485: scala-for-java-programmers.md uses and explains deprecated procedure method syntax

dongxuwang:
Fix #1485 , English.


leventov:
This PR doesn't reword this part, which now is straightaway misleading:

> The main method does not return a value (it is a procedure method). Therefore, it is not necessary to declare a return type.

dongxuwang:
@leventov good find

PR #1498, opened 2019-08-18
Add link to compiler plugins so readers can reach them

exoego:
It helps readers who interested in using/developing compiler plugins to reach plugins.

Updates this page.
https://docs.scala-lang.org/overviews/plugins/index.html


SethTisue:
nice

PR #1499, opened 2019-08-18

verkhovin:
I believe "http" should be returned in case when `ssl` is `false` :)


Philippus:
Good catch! Would you mind making this change in the translated tours as well?

verkhovin:
oh, sure. I'll run through the translations and make changes there too.

verkhovin:
@Philippus done.

Philippus:
Thank you!

PR #1508, opened 2019-08-23
Added missing argument when calling extension method equals

juanavelez:
Examples of how classes are translated to method extensions are missing an argument to the equals extension method


SethTisue:
thank you!

PR #1513, opened 2019-08-26
Add a tiny example for an explanation of an anonymous function

komainu8:
Dear, developers.

This is trivial modify.

We added a tiny example into "Tour of Scala - Basics" page.
Because in my opinion a little difficult to understand how to give a value to the anonymous function.

### DIFF
--- a/_tour/basics.md
+++ b/_tour/basics.md
@@ -110,6 +110,14 @@ You can define an anonymous function (i.e. no name) that returns a given integer
 
 On the left of `=>` is a list of parameters. On the right is an expression involving the parameters.
 
+You can give value to the anonymous function as below.
+
+{% scalafiddle %}
+```tut
+println(((x: Int) => x + 1)(5)) //6
+```
+{% endscalafiddle %}
+
 You can also name functions.
 
 {% scalafiddle %}


_tour/basics.md
```diff
@@ -110,6 +110,14 @@ You can define an anonymous function (i.e. no name) that returns a given integer
 
 On the left of `=>` is a list of parameters. On the right is an expression involving the parameters.
 
+You can give value to the anonymous function as below.
+
+{% scalafiddle %}
+```tut
+println(((x: Int) => x + 1)(5)) //6
```
exoego: I do not think it is not a common practice to pass arguments to anonymous function diretly.
I suggest a below.

```scala
val func = (x: Int) => x + 1
println(func(5)) // 6
println(func(10)) // 11
```

kotobotov: i think it's good to have both example

komainu8: Thank you very much for your comments.
I understood my example is strange.

The example that @exoego suggested has been already exist. Therefore, I just close this PR.

komainu8: Thank you very much for your comments.
I understood my example is strange.

The example that @exoego suggested has already existed. Therefore, I just close this PR.

PR #1517, opened 2019-08-29

ativ97:
In line 333, shouldn't the call to method "tostring" include parenthesis since the redefinition of "tostring" method in Class "Complex" includes parenthesis? This is in regard to be consistent with lines 294-305.


SethTisue:
in Scala, it's idiomatic to always omit `()` on `toString`, so I suggest we move in that direction, not this direction

PR #1518, opened 2019-08-29
Edit sentence for clarity

gmal1:


SethTisue:
hmm, it's not clear to me that this is an improvement. we want to say that the enforcement is taking place at compile-time. placing "at compile time" farther from the word "enforcement" obscures that rather than clarifies it, IMO. in the new version, it sounds like "at compile time" modifies "are used", which isn't the intended meaning.

is there some other way of improving the wording to address the clarity problem you're seeing?

_tour/tour-of-scala.md
```diff
@@ -32,7 +32,7 @@ Scala is also a functional language in the sense that [every function is a value
 Furthermore, Scala's notion of pattern matching naturally extends to the [processing of XML data](https://github.com/scala/scala-xml/wiki/XML-Processing) with the help of [right-ignoring sequence patterns](regular-expression-patterns.html), by way of general extension via [extractor objects](extractor-objects.html). In this context, [for comprehensions](for-comprehensions.html) are useful for formulating queries. These features make Scala ideal for developing applications like web services.
 
 ## Scala is statically typed ##
-Scala is equipped with an expressive type system that enforces at compile-time that abstractions are used in a safe and coherent manner. In particular, the type system supports:
+Scala is equipped with an expressive type system that enforces that abstractions are used in a safe and coherent manner at compile-time. In particular, the type system supports:
```
Philippus: Maybe some commas would help to make the sentence 'flow' better:
```suggestion
Scala is equipped with an expressive type system that enforces, at compile-time, that abstractions are used in a safe and coherent manner. In particular, the type system supports:
```

gmal1: @Philippus I like that. Or, perhaps something like `Scala's expressive type system enforces, at compile-time, that abstractions are used in a safe and coherent manner.`?

SethTisue: yeah, I agree that "is equipped with" doesn't really add meaning and might as well be dropped

SethTisue:
thank you!

PR #1522, opened 2019-09-05
Remove 'new' keyword when creating instances of case class

Zukkari:
Remove 'new' keyword when creating instances of 'Circle' class in 'Singleton Objects' section.

Class 'Circle' is defined as case class in given code segment.
Using 'new' keyword is redundant when creating instances of case class
as explained in section 'Case Classes'.


SethTisue:
thank you!

PR #1527, opened 2019-09-19
Fix invalid code example

coreywoodfield:


_style/declarations.md
```diff
@@ -220,7 +220,7 @@ There are three main reasons you should do this:
         // If, instead:
         def foldLeft[B](z: B, op: (B, A) => B): B
         // above won't work, you must specify types
-        List("").foldLeft(0, (b: Int, a: String) => a + b.length)
+        List("").foldLeft(0, (b: Int, a: String) => b + a.length)
```
kotobotov: i think it would be better  to use : `List("").foldLeft(0)((a: Int, b: String) => a + b.length)`

coreywoodfield: The code is an example of how a single parameter list can be more unwieldy than multiple parameter lists so I think that would defeat the point

kotobotov: 👌

PR #1530, opened 2019-09-24
Update string-interpolation.md

supaggregator:
For me maybe the two lines are inverted, cause we want to generate a json string, so we'd like to firstly append "StringContext" ?
Thanks.


SethTisue:
the existing code seems correct to me

Philippus:
@supaggregator `strings.next` is first called before entering the while-loop, in `var buf = new StringBuffer(strings.next)`, could it be that you missed that?

supaggregator:
> @supaggregator `strings.next` is first called before entering the while-loop, in `var buf = new StringBuffer(strings.next)`, could it be that you missed that?

Yes totally true...

supaggregator:
> the existing code seems correct to me

My bad...thanks

PR #1531, opened 2019-09-25
Double dollar syntax in string interpolation docs

bxt:
Hey, I realized the feature of Scala with the double dollar signs to produce a single dollar sign in an interpolated string is not documented anywhere as far as I can tell. So I think it should be mentioned in the docs here to give developers that stumble upon it in the wild a change to decipher it.


SethTisue:
thank you!

PR #1534, opened 2019-10-01
Fix(Algolia): guard the algolia search script execution if div does not exist

YoshinoriN:
## Desctiption

Many pages has JavaScript error `No input element in the page matches #doc-search-bar` caused by not found `<div class="search-container">`.

js-error

Because current all pages include JavaScipt for `Algolia search`. But, many pages hasn't `<div class="search-container">`.

That script is no need if a page hasn't following the search field.

search-input

## Solution

Divide JavaScipt and CSS for Algolia into other `_includes` template.

## Question

I created `_scripts` and `_styles` directory under the `_includes` directory and I put the Algolia's JavaScript and CSS to these directories. Is it no problem? Should I put them to under the `_includes` directly? Which is better?

---

Thanks :)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -100,13 +100,16 @@ $(document).ready(function() {
 
 <!-- Alogolia search for doc -->
 <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
-<script type="text/javascript"> docsearch({
-apiKey: 'fbc439670f5d4e3730cdcb715c359391',
-indexName: 'scala-lang',
-inputSelector: '#doc-search-bar',
-algoliaOptions: { 'facetFilters': ["language:en"] },
-debug: false // Set debug to true if you want to inspect the dropdown
-});
+<script type="text/javascript">
+if ($("#doc-search-bar").length) {
+  docsearch({
+    apiKey: 'fbc439670f5d4e3730cdcb715c359391',
+    indexName: 'scala-lang',
+    inputSelector: '#doc-search-bar',
+    algoliaOptions: { 'facetFilters': ["language:en"] },
+    debug: false // Set debug to true if you want to inspect the dropdown
+  });
+}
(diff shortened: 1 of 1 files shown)


SethTisue:
is there a reviewer who can evaluate this...?

Philippus:
I tried this, the console errors are still there. Adding a guard before `docsearch` is called would work, something like:
```
<script type="text/javascript"> if ($("#doc-search-bar").length) docsearch({
  apiKey: 'fbc439670f5d4e3730cdcb715c359391',
  indexName: 'scala-lang',
  inputSelector: '#doc-search-bar',
  algoliaOptions: { 'facetFilters': ["language:en"] },
  debug: false // Set debug to true if you want to inspect the dropdown
});
</script>
```

_includes/_scripts/algolia.html
```diff
@@ -0,0 +1,10 @@
+<!-- Algolia search for doc -->
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript"> docsearch({
```
Philippus:
```suggestion
<script type="text/javascript"> if ($("#doc-search-bar").length) docsearch({
```

YoshinoriN: @Philippus 
Thank you for your review. Sorry, maybe I made mistakes when I test this locally.
I updated it. 

One more question if we will choose this approach, no-need to divide algolia's assets into other files. So, should I revert directory and file structure change?

YoshinoriN:
Sorry for confusing you about the wrong change and comment.

I updated and force-pushed.  The new change is just checked div existence before execute the algolia script. This change is based on @Philippus's suggestion.

CI failure is another reason. It caused by broken external links. Some external links return 404.

> https://ci.scala-lang.org/scala/docs.scala-lang/566/3

Thank you.

Philippus:
Thank you!

PR #1546, opened 2019-10-10
update sip template

martijnhoekstra:
and also update SIPs to use the template format so that they are shown where they belong

Ideas blatantly stolen from https://github.com/scala/docs.scala-lang/pull/1541/commits/193d8be264fac76c7a5b9ebbe68dbe1598538612

Pinging those whose SIPs I've edited: @som-snytt @Atry

### DIFF
--- a/_sips/sip-template.md
+++ b/_sips/sip-template.md
@@ -1,6 +1,9 @@
 ---
-layout: sips
+layout: sip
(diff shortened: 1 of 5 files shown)


som-snytt:
Thanks for the heads up. I should either delete my contribution or adjust the date to April 1.

PR #1549, opened 2019-10-15
Updated pattern-matching.md

mehuled:
1. Added print statements for clarity.
2. corrected someSms to contain the non-important number.


mehuled:
Thanks. Kindly merge the PR.

SethTisue:
the [Tommy Tutone](https://en.wikipedia.org/wiki/Tommy_Tutone) reference was cute, but is culturally specific so I guess we're better off without it

PR #1553, opened 2019-10-17
Update scala-for-java-programmers

maksym-k:
As I understood, for current example truth table looks like (sure if == defined correctly):

```
this/that  <  <= >  >=
1/2        t  t  f  f
2/2        f  t  f* t*
3/2        f  f  t  t

where t - true,
      f - false
      * - incorrect value
```


ashawley:
I'm not seeing anything wrong with the current implementation of `Ord` trait on that page:

https://docs.scala-lang.org/tutorials/scala-for-java-programmers.html#traits

`2 > 2` should be false, and `2 >= 2` should be true.

Philippus:
I agree, the current implementation is correct.

PR #1558, opened 2019-10-24
grammar and consistency fix

vegerot:
line 39: `- project (sbt uses this to install manage plugins and dependencies)` should be:
              `- project (sbt uses this to install and manage plugins and dependencies)`
still doesn't sound quite good (too many "and"s), but is accurate.

Lines 44-45: Each file/function starts with a `- ` followed by a space.  Line 44 has a `-` but no space, line 45 doesn't have any `- `


SethTisue:
thank you!

PR #1562, opened 2019-10-28
Make immutable point clearer

vegerot:
I think having `message1` be a mutable `var` makes the point clearer that case classes' elements are immutable `val`s by default.  If `message1` is a `val`, readers may think the immutability comes from the larger instance `message1` being immutable.

P.S. You could also make a point that the opposite is true too.  If you define `Message` like
```
case class Message(var sender: String, recipient: String, Body: String)
```
and instantiate `message1` to be a `val`, then `message1.sender` is still a mutable `var`


SethTisue:
By itself, I don't think this is an improvement. The thinking behind it makes sense, but I don't think the change actually communicates the thinking to the reader.

PR #1570, opened 2019-11-05
Fix definition of narrow in SIP-23

valencik:
In SIP-23 there is a function defined:
```scala
def narrow[T <: Singleton](t: T): T = t
```
Using it to form a tuple appears to still return a wide type:
```
scala> val xt = (narrow(23), narrow(42))
xt: (Int, Int) = (23,42)
```
But if one changes the return type from `T` to `T {}` in, say, a `narrower` function:
```scala
def narrower[T <: Singleton](t: T): T {} = t
```
A tuple of literal types appears:
```
scala> val yt = (narrower(23), narrower(42))
yt: (23, 42) = (23,42)
```

Additionally the `T {}` return type is used in [sip23-narrow.scala](https://github.com/scala/scala/blob/4c6c6768e1511e1b62ef5302589f4aca8654015a/test/files/pos/sip23-narrow.scala#L6).

### DIFF
--- a/_sips/sips/2014-06-27-42.type.md
+++ b/_sips/sips/2014-06-27-42.type.md
@@ -99,7 +99,7 @@ Lightbend Scala compiler.
   ```
   def wide[T](t: T): T = t
   wide(13)                           // result is 13: Int
-  def narrow[T <: Singleton](t: T): T = t
+  def narrow[T <: Singleton](t: T): T {} = t
   narrow(23)                         // result is 23: 23
   ```
 
@@ -471,7 +471,7 @@ applications which work with large datasets.
   ```
   def wide[T](t: T): T = t
   wide(13)                           // result is 13: Int
-  def narrow[T <: Singleton](t: T): T = t
+  def narrow[T <: Singleton](t: T): T {} = t
   narrow(23)                         // result is 23: 23
   ```
 


julienrf:
Indeed, the empty refinement seems to be needed by Scala 2.13: https://scastie.scala-lang.org/45N4U3xfTkmgEuReMoq2cA

But I believe this is a problem in the implementation. This behavior is not specified in the description:

> The presence of an upper bound of `Singleton` on a formal type parameter indicates that
  singleton types should be inferred for type parameters at call sites.

So, the presence of the `Singleton` upper bound should be enough.

I think we should open an issue in scala/bug.

By the way, Dotty does not implement what is in the proposal, and the empty type refinement does not work: https://scastie.scala-lang.org/qkmxdI6GQyG9IFk8cLTzpw (so, you can not have an *inferred* narrow type, you have to write it explicitly).

Any comment @smarter or @milessabin ?

milessabin:
The following should work in Scala 2 (and Dotty) without relying on the semantics of empty refinements,

```scala
type Id[T] = T
def narrow[T <: Singleton](t: T): Id[T] = t
```
Demo ...
```
miles@tarski:~% scala                                                                                                                      
Welcome to Scala 2.13.0-20190613-143643-unknown (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_231).

scala> type Id[T] = T
defined type alias Id

scala> def narrow[T <: Singleton](t: T): Id[T] = t
narrow: [T <: Singleton](t: T)Id[T]

scala> val xt = (narrow(23), narrow(42))
xt: (Id[23], Id[42]) = (23,42)
```

julienrf:
According to the specification, the following should work as well:

~~~ scala
def narrow[T <: Singleton](t: T): T = t
val x = narrow(42)
x: 42
~~~

Or am I missing something?

BTW, I find it confusing that the introduction of the `Id` type constructor changes the inferred type.

milessabin:
> Or am I missing something?

Yes. In the example you give the compiler will widen the result type of the call of `narrow`,

```scala
scala> def narrow[T <: Singleton](t: T): T = t
narrow: [T <: Singleton](t: T)T

scala> val x = narrow(42)
x: Int = 42
```

> BTW, I find it confusing that the introduction of the Id type constructor changes the inferred type.

It's consistent with the behaviour of other type constructors, eg.,

```scala
scala> def narrow[T <: Singleton](t: T): Option[T] = Some(t)
narrow: [T <: Singleton](t: T)Option[T]

scala> val x = narrow(42)
x: Option[42] = Some(42)
```

ie. widening doesn't reach under the type constructor.

julienrf:
> Yes. In the example you give the compiler will widen the result type of the call of `narrow`,

In that case, I suggest that we fix the specification, then. Do you agree?

milessabin:
> In that case, I suggest that we fix the specification, then. Do you agree?

Yes, we should. I suggest using the `Id` variant.

SethTisue:
would someone like to PR the suggested spec change?

valencik:
I'd love to open a PR changing the spec. :)

SethTisue:
great! it's over in the scala/scala repo, https://github.com/scala/scala/tree/2.13.x/spec

PR #1572, opened 2019-11-06

jclerman:
Change print to println in the Partial Application subsection, so results appear immediately in e.g. ScalaFiddle


Philippus:
thank you!

PR #1573, opened 2019-11-06
Add information about unapply method

vegerot:
The documentation currently just states "we're calling the unapply method", without giving any indication to *how* it's being called initially.  Saying something like "we're calling the unapply method with `CustomerID.unapply(customer1ID)`" or "we're calling the unapply method with `name=CustomerID.unapply(customer1ID)`" earlier, while maybe not strictly accurate, helps give readers a better understanding of what's going on.  I think we should move (or copy) the section that starts with "This is equivalent to..." up above, and reiterate the point below.  I know for me personally I found this very confusing, and a line or two about this earlier would have been very helpful.

### DIFF
--- a/_tour/extractor-objects.md
+++ b/_tour/extractor-objects.md
@@ -32,7 +32,7 @@ customer1ID match {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


SethTisue:
thank you!

PR #1578, opened 2019-11-13
according to english documention, change chinese translation

GLBB:
I read chinese translation, I don't kown what's meaning until I see english version.


SethTisue:
thank you!

PR #1579, opened 2019-11-13
Grammar and wording changes to 'Basics' page.

Tazzle:


_tour/basics.md
```diff
@@ -54,13 +51,13 @@ println(x) // 2
...
 
-Values cannot be re-assigned.
+Values cannot be re-assigned:
 
 ```tut:fail
 x = 3 // This does not compile.
 ```
 
-Types of values can be inferred, but you can also explicitly state the type, like this:
+The type of a value can be omitted and inferred by [type inference](https://docs.scala-lang.org/tour/type-inference.html), or it can be explicitly stated:
```
Ichoran: Being inferred by inference is kind of redundant.  Maybe say something like, "When no type is specified, [the type is inferred](...).  The type can also be explicitly specifed:"

_tour/basics.md
```diff
@@ -74,11 +71,11 @@ Variables are like values, except you can re-assign them. You can define a varia
 
 ```tut
 var x = 1 + 1
-x = 3 // This compiles because "x" is declared with the "var" keyword.
+x = 3 // This compiles because "x" is declared with the "var" keyword:
```
Ichoran: This change is incorrect!  The comment refers to _this_ line, not the _next_ line which is what the colon suggests!  No colon--the original was correct.

_tour/basics.md
```diff
@@ -74,11 +71,11 @@ Variables are like values, except you can re-assign them. You can define a varia
...
 println(x * x) // 9
 ```
 
-As with values, you can explicitly state the type if you want:
+As with values, the type of a variable can be omitted and inferred by [type inference](https://docs.scala-lang.org/tour/type-inference.html), or it can be explicitly stated:
```
Ichoran: This should also be rewritten to avoid inferred-by-inference.

_tour/basics.md
```diff
@@ -100,17 +97,17 @@ println({
 
 ## Functions
 
-Functions are expressions that take parameters.
+Functions are expressions that have parameters, and take arguments.
```
Ichoran: This is redundant.  Having parameters is equivalent to taking arguments.  The original is better.  (Also, functions can take zero arguments, so the statement isn't even correct either way, but the new one is worse than the old one.)

Also, although "arguments" is more standard terminology, the parameters are called parameters in Scala, so it's probably better to leave it alone.

_tour/basics.md
```diff
@@ -119,7 +116,7 @@ println(addOne(1)) // 2
 ```
 {% endscalafiddle %}
 
-Functions may take multiple parameters.
+A function can take multiple arguments:
```
Ichoran: I am not convinced that _parameters_ should be replaced by _arguments_ given that the nomenclature universally uses the former (e.g. they are "parameter blocks" not "argument blocks").

_tour/basics.md
```diff
@@ -128,7 +125,7 @@ println(add(1, 2)) // 3
 ```
 {% endscalafiddle %}
 
-Or it can take no parameters.
+Or no arguments at all:
```
Ichoran: Again, not convinced of the change

_tour/basics.md
```diff
@@ -205,21 +202,23 @@ We will cover classes in depth [later](classes.html).
 
 ## Case Classes
 
-Scala has a special type of class called a "case" class.  By default, case classes are immutable and compared by value. You can define case classes with the `case class` keywords.
+Scala has a special type of class called a "case" class. By default, case classes are immutable, and they are compared by value (unlike classes, which are compared by reference). This makes them particularly useful for [pattern matching](https://docs.scala-lang.org/tour/pattern-matching.html#matching-on-case-classes).
```
Ichoran: This isn't what makes them useful for pattern matching.  They are _additionally_ useful for pattern matching because the compiler automatically generates the matcher.

_tour/basics.md
```diff
@@ -266,17 +265,17 @@ We will cover objects in depth [later](singleton-objects.html).
 
 ## Traits
 
-Traits are types containing certain fields and methods.  Multiple traits can be combined.
+Traits are abstract data types containing certain fields and methods. In Scala inheritance, a class can not only inherit from a super class, but also mix in one or more traits.
```
Ichoran: This is phrased somewhat awkwardly, though the point is good.  Maybe something like, "A class can only extend a single other class, but it can extend multiple traits."  (Caveat: I'm not sure "extend" is the right word to use.)

_tour/tour-of-scala.md
```diff
@@ -17,49 +17,49 @@ This tour contains bite-sized introductions to the most frequently used features
...
+Scala is also a functional language in the sense that [every function is a value](unified-types.html). Scala provides a [lightweight syntax](basics.html#functions) for defining anonymous functions, it supports [higher-order functions](higher-order-functions.html), it allows functions to be [nested](nested-functions.html), and it supports [currying](multiple-parameter-lists.html). Scala's [case classes](case-classes.html) and its built-in support for [pattern matching](pattern-matching.html) model algebraic types are used in many functional programming languages. [Singleton objects](singleton-objects.html) provide a convenient way to group functions that aren't members of a class.
```
Ichoran: "model algebraic types" is pretty abstruse.  While you're making changes, maybe you want to say something like, "...provide the functionality of algebraic types, which are used in many functional programming languages."

_tour/basics.md
Tazzle: I take your point but I was aiming this at newcomers to Scala who might want to know what it is inferred by.

_tour/basics.md
Tazzle: I take your point about the functions having zero arguments, but the original statement saying a function can "take" parameters doesn't make sense either.

 A function has parameters. When it says functions "take" parameters it sounds like "parameters" can be passed in. Arguments can be passed in, parameters can't.

_tour/basics.md
Tazzle: I can change this to "have" multiple parameters.

_tour/basics.md
Tazzle: You are correct. I will change this.

_tour/basics.md
Tazzle: I think I'm reading this as "for pattern matching, they are particularly useful", and you are reading this as "they are useful specifically for pattern matching", so I will change this due to ambiguity, thanks!

_tour/tour-of-scala.md
Tazzle: The change here is that I added in an extra "it" for "and it supports currying" but this change seems to only show up in the rich diff, unfortunately.

Ichoran: I'm aware--I just think that the "model algebraic types" language (using model as a verb) is so unclear that it ought to be fixed as long as you're making broad improvements here.

_tour/basics.md
Ichoran: I am reading it as, "The qualities I described above _is what makes them useful for matching."  This is what I object to, because it mostly isn't the reason they're useful for matching.

_tour/basics.md
Ichoran: Fair enough, but I don't think the clarity of the text should rely upon the reader understanding that arguments are the passed-in version of parameters.

_tour/basics.md
Ichoran: They can figure it out easily enough with a link and a suggestive phrase like "type is inferred", no?

Tazzle: Sorry, I was on mobile when I read that so I didn't see the hyperlink you add to "the type is inferred". Sure, that makes sense.

_tour/basics.md
Tazzle: If a reader sees "functions can _take_ parameters", they might wonder about the quality, and accuracy of the Scala documentation as a whole.

The information should be made as accurate as possible, regardless of the reader's level of understanding. 

I think this a (living) language thing. Some languages call them two different types of parameters; English calls them parameters and arguments.

_tour/basics.md
Tazzle: I don't know whether extend or mix in is more accurate.

In the [Traits](https://docs.scala-lang.org/tour/traits.html) page it says "Classes and objects can **extend** traits", and in the [mixins](https://docs.scala-lang.org/tour/mixin-class-composition.html) page it says "**Mixins** are traits which are used to compose a class.", so do you mix in a mixin, or do you extend a mixin?

SethTisue:
@Tazzle interested in getting this over the finish line...?

Tazzle:
> @Tazzle interested in getting this over the finish line...?

Tazzle:
> @Tazzle interested in getting this over the finish line...?
@SethTisue  sure! I didn't realise my first PR had been merged. Will get onto this today.

_tour/basics.md
SethTisue: I'm going to merge the PR regardless, but I lean more towards Rex's side here. I invite anyone interested to submit a followup PR with a different wording that we all like?

SethTisue:
Thanks Tazzle for the good work here! And thanks Rex for the useful review feedback.

PR #1580, opened 2019-11-16
fix in case class link description

dirceusemighini:
don't need to translate case class to portuguese, also clase is not a word


SethTisue:
thank you!

PR #1583, opened 2019-11-20
Fix a little bug in the example code

qiuyangnie:
The variable (a, b, c) in each case should be matched the variable printed out.

### DIFF
--- a/_overviews/scala-book/match-expressions.md
+++ b/_overviews/scala-book/match-expressions.md
@@ -226,8 +226,8 @@ Here are a few other examples of how you can use `if` expressions in `case` stat
 ```scala
 i match {
   case a if 0 to 9 contains a => println("0-9 range: " + a)
-  case b if 10 to 19 contains b => println("10-19 range: " + a)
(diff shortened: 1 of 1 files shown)


SethTisue:
@qiuyangnie I agree this is an improvement. I wonder if it might actually be better more idiomatic to write `case a` all three times. the current code kind of gives the impression that you need to use a different name each time, but you don't.

wdyt?

qiuyangnie:
I do agree with you! The earlier example in this chapter, however, should have demonstrated the 'variable name' in each `case` can be same. This example could provide an alternative way to demonstrate the 'variable name' in each `case` can be different.

SethTisue:
I think that's a subtlety that would be lost on readers unless it was explicitly called out, and I don't think it's actually worth calling out, because using a different name each time doesn't actually make the code better; I think it's actually worse that way.

So I still suggest that the right fix here is simply:

> write `case a` all three times

SethTisue:
(regardless: if you disagree or aren't interested in pursuing this farther, I'll hit "merge" anyway)

PR #1586, opened 2019-11-23
Replace the deprecated code

qiuyangnie:
There was one deprecation warning as procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `speak`'s return type or implicitly using `=`.


SethTisue:
good catch, thank you!

PR #1589, opened 2019-11-24
Fix a little bug in the example code

qiuyangnie:
```error: too many arguments (3) for method insert: (index: Int, elem: Int)Unit```

The fix is based on avoiding unnecessary repetition, and matching to comments.


SethTisue:
This is a 2.12/2.13 difference: in 2.12, `insert` took varargs. Without going digging, I don't know why it changed.

Thanks for the fix!

PR #1599, opened 2019-12-10
WIP Fix SIP-23 (Literal Types)

valencik:
This PR cleans up some incorrect code examples in the SIP-23 document.

> a PR bringing the SIP in line with the implementation would be welcome; so would a PR merging the whole SIP into the spec - @SethTisue (via [gitter](https://gitter.im/scala/contributors?at=5dee8b9832df1245cb112b45))

- [x] update comments showing return types to match repl session
- [x] confirm all examples in proposal details section
- [ ] figure out if we need to update the history section to reflect these changes

## Additional commentary
This PR better addresses the issues discussed in https://github.com/scala/docs.scala-lang/pull/1570
I do not think this addresses all the confusion one could run into when playing with literal types.
I think we need to improve the spec and write some documentation elsewhere, perhaps an overview or section in the tour.
However, hopefully this PR serves as an incremental improvement over the current state.

### DIFF
--- a/_sips/sips/2014-06-27-42.type.md
+++ b/_sips/sips/2014-06-27-42.type.md
@@ -113,7 +113,7 @@ Lightbend Scala compiler.
   added yielding the unique value of types with a single inhabitant.
   ```
   def foo[T](implicit v: ValueOf[T]): T = v.value
-  foo[13]                            // result is 13: 13
+  foo[13]                            // result is Int = 13
   ```
 
 
--- a/_sips/sips/2014-06-27-42.type.md
+++ b/_sips/sips/2014-06-27-42.type.md
@@ -89,9 +89,10 @@ Lightbend Scala compiler.
   ```
 
 + The `.type` singleton type forming operator can be applied to values of all subtypes of `Any`.
+  To prevent the compiler from widening our return type we assign to a final val.
   ```
   def foo[T](t: T): t.type = t
-  foo(23)                            // result is 23: 23
+  final val bar = foo(23)            // result is bar: 23 = 23
   ```
 
 + The presence of an upper bound of `Singleton` on a formal type parameter indicates that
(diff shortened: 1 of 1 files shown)


milessabin:
I think the example use of `.type` with subtypes of `AnyVal` should be tweaked, but should stay. We can prevent the widening with a singleton expected type, or by assigning to a final val, eg.,

```scala
scala> def foo[T](t: T): t.type = t
foo: [T](t: T)t.type

scala> final val bar = foo(23)
bar: 23 = 23
```

Otherwise LGTM.

valencik:
Hi @milessabin, thanks for the comment.
Unfortunately I am not seeing that behavior in scala 2.13.1:

```
Welcome to Scala 2.13.1 (OpenJDK 64-Bit Server VM, Java 1.8.0_222).

scala> def foo[T](t: T): t.type = t
foo: [T](t: T)t.type

scala> foo(23)
res0: Int = 23

scala> final var bar = foo(23)
bar: Int = 23
```
In each case I am seeing the widened type.

valencik:
Oh how embarrassing, look how that dreaded `var` snuck in and ruined the fun.
Using a `val` not a `var` we see the expected behavior return:

```
Welcome to Scala 2.13.0 (OpenJDK 64-Bit Server VM, Java 1.8.0_222).

scala> def foo[T](t: T): t.type = t
foo: [T](t: T)t.type

scala> foo(23)
res0: Int = 23

scala> final val bar = foo(23)
bar: 23 = 23
```

Although it is perhaps interesting to note that the type gets widened for a `var`...

I'll restore the example with `.type`

valencik:
A remaining question is whether we should update the history section of the SIP or not.

SethTisue:
I don't think the history section needs updating; that section's purpose is to alert the SIP committee to substantial changes to the proposal.

thanks @valencik for tackling this!

PR #1601, opened 2019-12-11
Update package-objects.md

liming199323:
方法重载在包对象中是起作用的，我测试过

### DIFF
--- a/_zh-cn/tour/package-objects.md
+++ b/_zh-cn/tour/package-objects.md
@@ -65,5 +65,3 @@ package object fruits extends FruitAliases with FruitHelpers {
   // helpers and variables follows here
 }
 ```
-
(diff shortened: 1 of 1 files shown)


liufengyun:
@SethTisue : it's about removing the last sentence in  `package-objects.md`: 

> Note that method overloading doesn't work in package objects.

If the restriction is lifted, maybe we need to remove the sentence for all language versions.

liufengyun:
@liming199323 Thanks for the PR. I believe the restriction is already lifted in the latest version of Scala. Could you please help to remove the sentence from other language versions as well?

Philippus:
Thank you! Changes for japanese and russian have been made in https://github.com/scala/docs.scala-lang/pull/1623.

PR #1602, opened 2019-12-13
Update variances.md

emonmishra:
a parameter is a declaration while an argument is an actual value that is supplied during invocation. In my opinion, calling it a parameter makes more sense.

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -135,7 +135,7 @@ Fortunately, the compiler stops us long before we could get this far.
 
 ### Other Examples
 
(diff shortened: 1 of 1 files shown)


emonmishra:
do I need to do anything else about it? Or is it going to be merged by someone else?

SethTisue:
we usually wait for a second reviewer. but if no second reviewer appears and the PR seems uncontroversial, we go ahead and merge anyway. so that's what I'll do now

SethTisue:
thanks for the improvement!

emonmishra:
thank you!

PR #1607, opened 2019-12-22
Add redirects for older files in /tutorials directories

### LINKED ISSUE #836: inconsistent html suffix
The top google result for `scala symbols` is http://docs.scala-lang.org/tutorials/FAQ/finding-symbols which is 404. It looks like the correct url might be http://docs.scala-lang.org/tutorials/FAQ/finding-symbols.html. 

There are several other previously-indexed pages as well: https://www.google.com/search?q=-filetype:html+site:docs.scala-lang.org (starting on page 2).

Another thing I noticed is that if you somehow get a link like this: http://docs.scala-lang.org/tutorials/tour/singleton-objects.html.html the links on the sidebar also contain `.html.html`, but not all pages' navigation have this behavior.

Because both suffixed and non-suffixed pages are currently indexed, I think it probably makes sense, for discoverability, to support both but perform some sort of redirect and navigation normalization to whichever is desired.

### LINKED ISSUE #1606: https://docs.scala-lang.org/tutorials deprecated?
Hey, I've noticed that the deprecated website [https://docs.scala-lang.org/tutorials](https://docs.scala-lang.org/tutorials) is still up, containing an older version of _A Tour of Scala_. (The latest version can be found at [https://docs.scala-lang.org/tour/tour-of-scala.html](https://docs.scala-lang.org/tour/tour-of-scala.html).) Is there a reason for not removing the old version?

Philippus:
Applying this PR will add redirects for all the older files that are still present in the /tour and /tutorials directories for all languages that were part of that version of the tour.
Fixes #1606 and I think it fixes #836 as well.


SethTisue:
there may have been a race condition between you and Fabien: https://github.com/scala/docs.scala-lang/issues/1606#issuecomment-568444771

is this still needed, after whatever Fabien did?

Philippus:
yes, I saw it. I think this is no longer needed and the solution by @fsalvi  is probably cleaner. :)

PR #1613, opened 2020-01-05
Correct typo in built-in-types.md

perfumister:
Leading minus symbol missing from negative end of `Short` and `Int` table descriptions


SethTisue:
thank you!

PR #1619, opened 2020-01-11
uu tube

som-snytt:
As in vacuum tube.

My French friend, who is a diacritical critic, says don't retain the accent in the short form:

> Séb: yeah, maybe

Maybe @sjrd will clarify for us barbarians.

### DIFF
--- a/_sips/minutes/2019-12-18-sip-minutes.md
+++ b/_sips/minutes/2019-12-18-sip-minutes.md
@@ -55,7 +55,7 @@ committee and/or which they will spend time investigating and trying to poke hol
 2. provide a detailed review of the feature, especially for the contentious ones; and, finally
 3. straw poll will be taken on each feature, as a smoke test for the real vote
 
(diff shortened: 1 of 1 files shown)


dwijnand:
Thanks!

sjrd:
Indeed, in French the short form of Sébastien is Seb, without accent. This matches the prononciation, which becomes more Sèb than Séb. That change of prononciation has something to do with closed and open syllables, but I don't recall the precise rules right now.

som-snytt:
Thanks. I asked my source of all things French to say the two phonemes, but I couldn't really distinguish them; or maybe my worsening tinnitus prevents me; but I still derive some pleasure from orthography.

SethTisue:
> Indeed, in French the short form of Sébastien is Seb, without accent

I always wondered & now I know.

som-snytt:
The other question is why Seb instead of Bas.

The great wonder is how Saint Seb survived the `=>`s.

PR #1620, opened 2020-01-11
Update string interpolation docs with further information on escaping special chars

### LINKED ISSUE #1314: Mention / Adding a section on escaping (double quotes, dollar sign, etc) in String Interporlation
First and foremost, thank you for this awesome project 😄 !
This [article](https://docs.scala-lang.org/overviews/core/string-interpolation.html) doesn't say much escaping special character with string interpolations. IMHO, it would be helpful to highlight this works for special characters such as `double quotes`, `dollar sign` etc
It might be worth as well mentioning string interpolation with triple quotes as well

chloecollins:
Resolves  #1314

Added ways in which Interpolation with double quotes can be achieved. Dollar signs are already documented.

Are there are further characters which would be necessary?

### DIFF
--- a/_overviews/core/string-interpolation.md
+++ b/_overviews/core/string-interpolation.md
@@ -42,11 +42,22 @@ String interpolators can also take arbitrary expressions.  For example:
 
 will print the string `1 + 1 = 2`.  Any arbitrary expression can be embedded in `${}`.
 
+For some special characters, it is necessary to escape them when embedded within a string.
 To represent an actual dollar sign you can double it `$$`, like here:
 
     println(s"New offers starting at $$14.99")
     
 which will print the string `New offers starting at $14.99`.
+
+Double quotes also need to be escaped. This can be done on an individual basis using a backslash before the double quote, as follows:
+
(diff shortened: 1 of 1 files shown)


martijnhoekstra:
Including this in this particular section on the interpolations is somewhat misleading, in that you *can't* use `\` to escape double quotes in interpolations.

SethTisue:
@chloecollins thanks for tackling this — do you think your changes ought to be adjusted to reflect Martijn's feedback?

chloecollins:
Thanks, @martijnhoekstra - I see that you are correct in pointing out that the backslash symbol cannot be used in this way. Reading your documentation at https://docs.scala-lang.org/sips/interpolation-quote-escape.html has clarified this for me. I have removed that section and now only refer to triple quotes as a solution.

SethTisue:
@martijnhoekstra wdyt of the latest version?

martijnhoekstra:
Looks like a net positive, and mergeable to me. Making it an s interpolation rather than a string literal would be better still.

SethTisue:
thank you @chloecollins! welcome to the repo

PR #1626, opened 2020-01-21
Update metadata

Philippus:
Applying this PR will make the SIP show up in the list and will give it a permalink. @dwijnand is the status of 'pending' correct?


dwijnand:
Reading https://docs.scala-lang.org/sips/sip-submission.html#proposal-states my guess is it's "Validated" given it was merged?  But perhaps for "vote status" pending applies.

Thanks for the fix!

PR #1629, opened 2020-01-26
Terminology fix: case class instances

int-index:
Before this patch, the text said things like "case classes are compared by value". This caused me some confusion: is it the classes themselves that are compared, or is it the instances of those classes?


SethTisue:
thanks!

PR #1632, opened 2020-02-02
Add "number" to "complex"

RohanTalip:
The use of "complex" here is as an adjective, not a noun, so it needs to be used with "number".


SethTisue:
CI failure is unrelated (and will be fixed by #1637)

SethTisue:
thank you!

PR #1633, opened 2020-02-02
Improved the grammar around type inference

RohanTalip:
Also removed the negative part of the statement as it seems redundant.


SethTisue:
thank you!

PR #1636, opened 2020-02-02
Updated the comment for the return type

RohanTalip:
The previous inline code highlighting of `: Int` together was confusing: the sentence as a whole seemed to imply that the return type could be found _after_ the colon *and* the `Int`, i.e. the sentence could be read to imply that `: Int` was a separate syntactic construct from the return type, even though the return type `Int` was already there in `: Int`.

### DIFF
--- a/_tour/basics.md
+++ b/_tour/basics.md
@@ -145,7 +145,7 @@ println(add(1, 2)) // 3
 ```
 {% endscalafiddle %}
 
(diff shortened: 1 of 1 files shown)


SethTisue:
CI failure is unrelated (and will be fixed by #1637)

Philippus:
thank you!

PR #1642, opened 2020-02-15
docs(tour/by-name-parameters): make more clear

yxliang01:
In the description, it can be more clear that the by-name-parameters are similar to be an expression that will be evaluated again and again each time it is used. Please correct me if wrong.

Thanks!

### DIFF
--- a/_tour/by-name-parameters.md
+++ b/_tour/by-name-parameters.md
@@ -10,7 +10,7 @@ previous-page: operators
 redirect_from: "/tutorials/tour/by-name-parameters.html"
 ---
 
-_By-name parameters_ are evaluated every time they are used. But, they won't be evaluated at all if they are unused. This is similar to replace the by-name parameters with the passed expressions. They are in contrast to _by-value parameters_. To make a parameter called by-name, simply prepend `=>` to its type.
+_By-name parameters_ are evaluated every time they are used. They won't be evaluated at all if they are unused. This is similar to replacing the by-name parameters with the passed expressions. They are in contrast to _by-value parameters_. To make a parameter called by-name, simply prepend `=>` to its type.
 ```tut
 def calculate(input: => Int) = input * 37
 ```
--- a/_tour/by-name-parameters.md
+++ b/_tour/by-name-parameters.md
@@ -10,7 +10,7 @@ previous-page: operators
 redirect_from: "/tutorials/tour/by-name-parameters.html"
 ---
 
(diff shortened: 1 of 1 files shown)


dwijnand:
> by-name-parameters are similar to be an expression that will be evaluated again and again each time it is used. Please correct me if wrong.

That's correct.

> This is similar to replace the by-name parameters by the passed expressions.

This only right if you bind that passed expression to a `def`, for example

```scala
def writeToConsole = println("hi")
val writeToConsole = println("hi") // <- not the same
```

There's a little grammatical work that needs to be done to this before it's ready to be merged.

yxliang01:
Thanks @dwijnand  I can understand the different between the given examples. Let's say now the define `writeToConsole` using `val`, the passed expression is `writeToConsole` which is essentially `Unit`(if `println` returns `Unit`). I think the sentence "This is similar to replace the by-name parameters with the passed expressions" is still correct under this case?

SethTisue:
I don't think the sentence "This is similar to replace the by-name parameters with the passed expressions" is sufficiently clear. I can see what it means, but that's because I already understand by-name parameters.

One attempt at a clearer version: "This is similar to inlining the argument expression everywhere a reference to the by-name parameter appears". But, I don't know, I still have doubts that it actually improves the overall clarity.

yxliang01:
@SethTisue "Inlining" sounds quite abstract somehow. Actually, I just want to emphasize that its functionality is similar to the `#define` in C. TBH, this functionality of Scala makes it powerful and meanwhile sometimes producing surprising results. Especially the fact that it's unclear whether we are passing an expression by-value or by-name by looking at the invocation code.

SethTisue:
Yeah, perhaps "inlining" isn't the best either :-/

Maybe “It behaves as if anywhere the by-name parameter is referenced, it is replaced with the passed expression"?

Well, I don't mean to nitpick. I think your original would also be mergeable if you like at best, but with this fix: "replacing", not "replace".

martijnhoekstra:
I suspected Dale meant removing the "But, "

I'm totally using Dale as a scapegoat for my own preferences.

SethTisue:
@martijnhoekstra I don't even want to _know_ what hideous color your bikeshed is

yxliang01:
Indeed, thanks @SethTisue and @martijnhoekstra for pointing out the grammar problems. At the moment, I can't think of a better wording for describing this while I really feel the original wording isn't so understandable :joy: This feature of Scala is simply too powerful :P But, this is not something should be blamed I think :)

SethTisue:
thank you!

PR #1646, opened 2020-02-24
Update futures.md

mukesh210:
"callbacks registered on the same future are unordered" is a crucial information. It would be better to have it in **bold** font-weight so that it is emphasised.


SethTisue:
I don't think this is the right way to emphasize something. In standard English prose, you don't just bold whole sentences like that. If you think the point deserves more emphasis, I suggest accomplishing that with rewriting and/or additional wording.

SethTisue:

SethTisue:
happy to reopen if activity resumes

PR #1653, opened 2020-03-03
Update XHtml -> xHtml in `Naming Convention` 

havinhdat:
we shouldn't capitalize first word so we're not confuse with constant values

### DIFF
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -19,7 +19,7 @@ each word is capitalized, except possibly the first word:
 
 Acronyms should be treated as normal words:
 
-    Xhtml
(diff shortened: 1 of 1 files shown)


ricky-wong:
Sorry for coming here super late. I'm confused about this change. The acronym is "`XHTML`", not "X HTML", so why would `xHtml` be appropriate? The existing `Xhtml` (upper camel case) / `xhtml` (lower camel case) seemed correct. The PR title here is kinda misleading because it was never `XHtml` to begin with (the diff shows `Xhtml`->`xHtml`). I think this was just a case of having a bad/confusing example.

PR #1655, opened 2020-03-11
Update the SIP committee members list

dwijnand:
Eugene B left, Miles joined.  And sort it (keeping Martin first).

Also, I added myself as secretary.


SethTisue:
replaced by #1664

PR #1657, opened 2020-03-12
Clarify the type of for comprehension.

slaykachu:
I know we don't want to dig into details in the tutorial, but here the sentence was generally wrong.


SethTisue:
this is going in the right direction for sure, but wdyt about dropping "usually"?

SethTisue:

SethTisue:
I've submitted my own version at #1688

PR #1658, opened 2020-03-12
Improve text grammar

rhumbertgz:
- Missing commas 
- Pronoun problems

### DIFF
--- a/_tour/traits.md
+++ b/_tour/traits.md
@@ -12,7 +12,7 @@ prerequisite-knowledge: expressions, classes, generics, objects, companion-objec
 redirect_from: "/tutorials/tour/traits.html"
 ---
 
-Traits are used to share interfaces and fields between classes. They are similar to Java 8's interfaces. Classes and objects can extend traits but traits cannot be instantiated and therefore have no parameters.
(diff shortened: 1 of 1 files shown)


_tour/traits.md
```diff
@@ -29,7 +29,7 @@ trait Iterator[A] {
 }
 ```
 
-Extending the `trait Iterator[A]` requires a type `A` and implementations of the methods `hasNext` and `next`.
+Extending the `trait Iterator[A]` requires a type `A`, and implementations of the methods `hasNext` and `next`.
```
SethTisue: adding a comma here is incorrect. a comma would only be correct here if it introduced an independent clause

if you feel the sentence is hard to read without the comma, we need to find some other solution.

rhumbertgz: No problem, maybe it just my opinion 👍🏻. 
I close this PR

SethTisue: but I liked the other two changes!

rhumbertgz: @SethTisue I'll remove the conflicting change and leave the other

rhumbertgz: @SethTisue this is the new PR[#1665](https://github.com/scala/docs.scala-lang/pull/1665)

PR #1661, opened 2020-03-14
Update generic-classes.md

umang-malhotra:
Changed 'enable us' to 'not allow us' in line 58

The line  'This would be unsound because it would enable us to enter true integers into the character stack. To conclude ' 

It doesn't seem to give it's intended meaning.  I thought it means that since we are not able to add stack[Int] to stack[Char]. That's why the next line says about scala allowing us something like variance.


umang-malhotra:
Or there should have been some line like - If we allow us something like adding Stack[Int] to Stack[Char] this would be unsound because it would enable us ...

SethTisue:
I believe the original text was correct and the change you're suggesting would make it incorrect.

The fact that it didn't seem that way to you probably indicates that the clarity of the text could be improved. But regardless, we cannot accept this change.

PR #1662, opened 2020-03-14
Fix broken links

bsolomon1124:
Link [2] is still broken, but I have no idea what that's supposed to refer to...


SethTisue:
thank you!

PR #1666, opened 2020-03-19
add extra clarity to type description

AaronDelaplane:
I don't know if this correction needs to be made.  If a type is always a subtype of itself then the way this statement exists in the documentation presently makes sense.  If a type cannot be considered to be a subtype of itself then this correction might be worth adding.  Thanks for reviewing:)

### DIFF
--- a/_tour/upper-type-bounds.md
+++ b/_tour/upper-type-bounds.md
@@ -10,7 +10,7 @@ previous-page: variances
 redirect_from: "/tutorials/tour/upper-type-bounds.html"
 ---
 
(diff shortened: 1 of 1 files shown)


SethTisue:
subtyping is reflexive; every type is a subtype of itself. when we want to exclude the type itself, we say "proper subtype"

maybe a good wording would be "refers to a subtype of type `A` (which might be a proper subtype of `A` or might be `A` itself)" ?

SethTisue:

SethTisue:
happy to reopen if activity resumes

Issue #1677, opened 2020-04-03
Wrong order of sections in chapter "case class" of scala book


marcelo-rosa:
I think Section **No mutator methods** should be moved after Section **An apply method means you don’t need new**, since the example of the first section requires the user have seen the example of the second section.

Or, you cannot execute

```
scala> christina.name
res0: String = Christina

// can't mutate the `name` field
scala> christina.name = "Fred"
<console>:10: error: reassignment to val
       christina.name = "Fred"
```
without having executed
```
scala> case class Person(name: String, relation: String)
defined class Person

// "new" not needed before Person
scala> val christina = Person("Christina", "niece")
christina: Person = Person(Christina,niece)
```

SethTisue:
@alvinj wdyt?

alvinj:
Let me take a look at this. We changed parts of the book when I added it here, and I may have moved something so that it’s out of order.

alvinj:
Wow, yeah, something is wrong with that page. If I can’t fix it today, I’ll get it fixed this weekend.

alvinj:
I forgot to reference this issue when I submitted my pull request a few days ago, but that pull request is #1679. I see that it says “All checks have failed”, but I hope that’s a minor thing.

SethTisue:
> I see that it says “All checks have failed”, but I hope that’s a minor thing.

It's unrelated — some broken links elsewhere on the site (which I'll go fix right now).

SethTisue:
>  I see that it says “All checks have failed”, but I hope that’s a minor thing.

Hmm, weird. I'll comment about that on the PR itself.

SethTisue:
fixed by #1682. thanks @marcelo-rosa for the report

PR #1682, opened 2020-04-13
In response to Issue #1677, I reversed the first two sections

alvinj:
I also simplified some other text in that same lesson.


SethTisue:
thanks Alvin!

alvinj:
You’re welcome!

PR #1684, opened 2020-04-16
Link to implicit parameter page on first mention

timokau:
When reading the tour I noticed that `implicit` is mentioned before it is defined.


SethTisue:
thank you!

PR #1685, opened 2020-04-17
The correct syntax of fixed

scyber:


scyber:
This is a correct syntax apply of function to the List.

martijnhoekstra:
I recently re-wrote the English version of that page, to make that part specifically clearer: https://docs.scala-lang.org/tour/multiple-parameter-lists.html

That part is meant to illustrate the signature that doesn't work, but doesn't do so very clearly.

SethTisue:
@scyber interested in updating the translation to match the new text?

SethTisue:
closing for inactivity, happen to reopen if activity resumes

PR #1689, opened 2020-05-01
2.13 collections doc: link to Vector changes

SethTisue:
and massage text slightly


SethTisue:
oh btw I didn't see anything in `_overviews/collections-2.13/performance-characteristics.md` that needed changing, but an especially diligent and/or knowledgeable reviewer might double-check that :-)

PR #1690, opened 2020-05-09
Add another example to self-type.md

Tabzz98:
I found the best use of self types to be accessing members of the outer scope from within an anonymous implementation of a trait, but it wasn't mentioned, so I added an example.


SethTisue:
This code compiles without using a self-type if you 1) have `trait Tweeter extends User`, and 2) use `Tweeter.this` when you need to access the outer this

Personally I find self-types to be an obscure and rarely needed language feature, and so I always steer people in the direction of `extends` instead of a self-type unless the self-type is really doing something that `extends` can't accomplish. But that doesn't seem to be the case here.

PR #1693, opened 2020-05-14
A little intuition for variance to help injestion

ozenzin:
Some "producer extends, consumer super" flavoring helping learners to anchor concept of variance to existing experiences.


SethTisue:
Honestly, I'm afriad I don't think this improves the clarity. Anyone else watching the repo have an opinion?

martijnhoekstra:
I'm afraid I have to agree with Seth. While I do think this page could be clearer, I don't believe this PR accomplishes that.

PR #1699, opened 2020-06-03
Simplify project creation instructions

oswaldo:
Instead of suggesting to call `mvn archetype:generate` and then filtering to find the recommended archetype, just give its coordinates on the mvn call already.
Also, the observation about multiple scala maven plugins is obsolete.


julienrf:
Thank you @oswaldo!

I have no experience in doing Scala projects with maven. Does one of the maintainers have experience and can review the PR?

SethTisue:
@davidB? @slandelle?

slandelle:
LGTM. Then, this archetype needs a serious update (I wasn't aware of its existence). @davidB, could you please make me a committer there as well? :)

PR #1700, opened 2020-06-04
Updates Library Author Guide regarding Nexus

gaelrenoux:
As far as I can tell, the current description does not correspond to the correct place to look anymore. There is still a Staging repository in "User Managed Repository" (the type selected by default), but it does not contain a library waiting to be released.
Since it was a head-scratcher for me when publishing my first lib, I assume it might help others.

### DIFF
--- a/_overviews/contributors/index.md
+++ b/_overviews/contributors/index.md
@@ -215,7 +215,7 @@ $ sbt publishSigned
 
 `sbt-sonatype` will package your project and ask your PGP passphrase to sign the files with your PGP key.
 It will then upload the files to Sonatype using your account credentials. When the task is finished, you can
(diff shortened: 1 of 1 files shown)


julienrf:
Hello Gael,
In my case, I still see the “Staging Repositories” link:

Screenshot from 2020-06-04 18-26-33

What do you see?

gaelrenoux:
OK, I got it: you can only see that when you are connected. When you're not (which I wasn't when I was looking for my artifact), there is no "Build Promotion" group.

So it's probably better to clarify it that way (you need to be connected, and Staging Repositories appear on the left menu in "Build Promotion"). The issue being: there is also a "Staging" in the list of repositories, and that's not what you're looking for.

gaelrenoux:
I pushed a version today and confirmed I could see the staging repository. I corrected the text to tell where to look, and to mention the fact you should be logged in.

PR #1703, opened 2020-06-09
add additional information about for-comprehention desugaring

letalvoj:
Once in a while I encounter this feature of for comprehentions. Today I was trying to figure out what exactly is going on and how the functions passed to `withFilter` looks like. Since it is not exaplained in the docs I decided to create a PR.

Plus adding few missing ``` ` ``` at the begininig of the next paragraph.


SethTisue:
thank you, nice improvement!

PR #1706, opened 2020-06-11
reshuffle books list (-Impatient, +Creative)

SethTisue:
updates https://docs.scala-lang.org/books.html

* remove Scala for the Impatient, it is outdated 😢 
* move @lihaoyi's book up to 3rd position
  * also changed the tag line from “Available now" to "Covers Scala 2.13", in the hopes that readers won't get the mistaken impression it is less up-to-date than the first two books
* in vacated 4th position, add Creative Scala /cc @noelwelsh


SethTisue:
If we can list only six books, I'm not sure Scala Puzzlers is still the best 6th book. Not saying it's a bad choice, but perhaps there's a better one? Open to suggestions. (And the Center should make the final call.)

SethTisue:
By the way, I believe that below the "big six", no one would be opposed to *also* offering a longer list of all Scala books, which I suggest would be organized by year (of latest edition). The only reason we have no such list is that no one has stepped up and produced one.

julienrf:
+1 for Creative Scala.

Another book that looks nice (I haven’t read it, though) is [Get Programming with Scala](https://www.manning.com/books/get-programming-with-scala).

I’m sure I’m missing some other books…

SethTisue:
(Daniela's book looks like it will be a strong candidate, but it's still a MEAP; I don't think we should the short list should include books that aren't finished yet.)

ekrich:
How about Alvin Alexander's book. His site is one of the most used so I think linking to his book would be great. Plus the cover page is great. https://www.amazon.com/Functional-Programming-Simplified-Alvin-Alexander-ebook/dp/B076J7CJKY

SethTisue:
for the short list, I'd probably prefer Alvin's _Scala Cookbook_ (https://www.amazon.com/Scala-Cookbook-Object-Oriented-Functional-Programming/dp/1449339611)

but then there's also prolific Alvin's The Scala Book (https://docs.scala-lang.org/overviews/scala-book/introduction.html), and it's kind of weird we don't link to that from here

ekrich:
Yeah, the cookbook is much older. The newer one is massive but linking from the free on-line one is a great idea. Honestly, I really like the cover too.

SethTisue:
I think I'm just going to merge this and leave the sixth-book question for another day.

PR #1709, opened 2020-06-11
Fix outdated sample code for querying types in Symbols doc

nogurenn:
Other parts of this have the updated `___Name()` syntax. These examples were probably just missed.

### DIFF
--- a/_overviews/reflection/annotations-names-scopes.md
+++ b/_overviews/reflection/annotations-names-scopes.md
@@ -200,7 +200,8 @@ Constant expressions are used to represent
 
 Example:
 
-    Literal(Constant(5))
+    scala> Literal(Constant(5))
+    val res6: reflect.runtime.universe.Literal = 5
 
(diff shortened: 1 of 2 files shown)


nogurenn:
I missed this part. What return type do I use? I'll add it to this PR.
```scala
@ import scala.language.experimental.macros 
import scala.language.experimental.macros

@ def impl(c: scala.reflect.macros.Context) = c.Expr[Unit](c.parse("println(2)")) 
defined function impl

@ def test = macro impl 
cmd102.sc:1: macro defs must have explicitly specified return types
def test = macro impl
    ^
Compilation Failed
```

SethTisue:
`Unit`:

```
scala 2.13.2> def test: Unit = macro impl
def test: Unit

scala 2.13.2> test
2
```

nogurenn:
I missed that big time. Sorry :(

SethTisue:
excellent, thank you!

PR #1714, opened 2020-06-19
Add missing comas

rhumbertgz:
Only a suggestion (I'm not a native speaker)


SethTisue:
Native speaker here. It's actually incorrect to add a comma before "which are executed frequently" — though I would have trouble explaining why.

The other commas I think are fine either way, so not really worth changing.

PR #1717, opened 2020-06-19
Use synonym to avoid duplicated word in a sentence

rhumbertgz:
Only a suggestion ;)


SethTisue:
I think for some kinds of programs "several" might be sufficient, but for others kinds of programs, "many" might actually be necessary. So I don't think this is an improvement.

If saying "many" bothers you, perhaps "multiple" would be a suitable replacement.

PR #1718, opened 2020-06-21
Tuple1

dgruntz:
Adds `Tuple1` to the list of explicitly mentioned tuples after the sentence "Scala uses a series of classes: ".

### DIFF
--- a/_tour/tuples.md
+++ b/_tour/tuples.md
@@ -27,7 +27,7 @@ This creates a tuple containing a `String` element and an `Int` element.
 The inferred type of `ingredient` is `(String, Int)`, which is shorthand
 for `Tuple2[String, Int]`.
 
-To represent tuples, Scala uses a series of classes: `Tuple2`, `Tuple3`, etc., through `Tuple22`.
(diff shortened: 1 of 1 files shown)


NthPortal:
I think the rationale behind leaving it out is that `Tuple1` isn't actually useful for anything, as you can just return its value. it's essentially a box.

I actually didn't know it existed, and am not sure why it does.

dgruntz:
Same for me! I already wanted to blame someone who mentioned (in a Scala Book) that tuples exist from `Tuple1` onwards to `Tuple22`, and I looked up the Scala docs and found this page. But a look into the sources showed that `Tuple1` really exists. You could also add a note following your comment, namely that `Tuple1` also exists but that it is not really useful.

> I actually didn't know it existed, and am not sure why it does (probably "symmetry" or something).
Just realized that method `tupled` is not defined on `Function1`. This method would have been a reason to define class `Tuple1`.

SethTisue:
I think including `Tuple1` without a caveat could mislead newcomers. Possible solutions include:

* mention it, but without including it in the main list; rather, add some kind of note that says that it exists but is not normally used
* just don't mention it at all (the status quo)

Since this is the language tour and the tour only aims to hit highlights and not serve as comprehensive reference documentation, I lean a bit towards the “just don't mention it" option.

PR #1719, opened 2020-06-24
Update macros context code to use whitebox context

### LINKED ISSUE #1710: Update macros doc for tree creation to use blackbox/whitebox context
https://docs.scala-lang.org/overviews/reflection/symbols-trees-types.html#tree-creation-via-parse-on-toolboxes

I'm reading about reflection, and running the macros example gives the deprecation warning:
```scala
def impl(c: scala.reflect.macros.Context) = c.Expr[Unit](c.parse("println(2)"))
                                 ^
       warning: type Context in package macros is deprecated (since 2.11.0):
                use blackbox.Context or whitebox.Context instead
```

I thought it makes sense to raise this because of the existence of two different types of `Context`s.

nogurenn:
Closes #1710

Is there a one-liner/short code I can use to illustrate the difference of whitebox and blackbox contexts? Something like

    funcThatShowsDiff(macro blackImpl)
    funcThatShowsDiff(macro whiteImpl)


SethTisue:
The underscore on `_final` is peculiar, I think. I understand that it's there to avoid the clash with the keyword `final`, but I think it would be better to avoid the clash some other way. I'll add a commit with a different renaming.

re: one-line examples, I can't think of anything offhand; I think the PR is mergeable regardless.

SethTisue:
thank you!

nogurenn:
I usually do `_foo` for specifically localized implementations where a `foo` name makes sense in the scope also, treating `_foo` as like a temp variable. I forgot how I came up with it, or where I picked it up. I remember thinking it's better than `val someVagueTempVariableName` to me. Maybe I got it from another language that implements `_foo` for internals, and exposes `foo` in its interfaces?

```scala
// an example whose form makes more sense in code like nodejs
val persons = {
  val data = db.run(personsQuery.result)
  val _persons = ... // code further processing `data`
  
  return _persons
}
```

SethTisue:
> another language that implements _foo for internals, and exposes foo in its interfaces?

(I see that in Scala codebases too, including my own.)

nogurenn:
No problem from my end! Just thought it's a nice reminder to re-assess my coding practices from time to time

PR #1723, opened 2020-06-30
Update scaladoc interface docs

coreyoconnor:
Removes the content that is only relevant to 2.11 scaladoc UI. A few edits to update the content for the current scaladoc UI.

The description for the current mechanism for expanding/collapsing sections still needs improvement. "Activate the box"?


SethTisue:
Thank you!

And yeah, there's always more room for improvement :-)

PR #1728, opened 2020-07-04
Corrected tuple definition

manishbansal8843:
Earlier statement was more ambiguous for a beginner.

### DIFF
--- a/_tour/tuples.md
+++ b/_tour/tuples.md
@@ -12,7 +12,7 @@ redirect_from: "/tutorials/tour/tuples.html"
 ---
 
 In Scala, a tuple is a value that contains a fixed number of elements, each
-with a distinct type.  Tuples are immutable.
+with a same or distinct type.  Tuples are immutable.
 
 Tuples are especially handy for returning multiple values from a method.
 


SethTisue:
I don't see how this is an improvement. Perhaps we can find some other way to improve it. What was the issue here that you saw?

manishbansal8843:
@SethTisue I feel the definition is bit ambiguous for a newcomer. The current definition somehow gives an impression that scala tupple only accepts distinct data type. It would be better if we add `same or distinct` data type. If you have any other suggestion, do let me know.

SethTisue:
The problem I have with "same or distinct" is that in the overwhelming majority of cases, the types will be distinct. The case where all the types happen to be the same is rare. Whereas phrasing it as "same or distinct" gives the impression that "same" is normal and expected, and that "distinct" is the alternative.

manishbansal8843:
Even though the majority use case will be for distinct types, I somehow feels documentation should be comprehensive. If you want, we can make it as `distinct or same` to emphasize on the `distinct` part.

SethTisue:
I don't think "distinct or same" is clear, idiomatic English. It muddies the question rather than clarifying it.

Maybe some entirely different phrasing such as "each with its own type"?

manishbansal8843:
How about this.

`In Scala, a tuple is a value that contains a fixed number of homogeneous or heterogeneous types of elements.  Tuples are immutable.`

Or we could remove the phrase all together to avoid ambiguity.

`In Scala, a tuple is a value that contains a fixed number of elements.  Tuples are immutable.`

SethTisue:
> Or we could remove the phrase all together 

The fact that the elements have individual types is one of the most crucial and important pieces of information on the whole page. Omitting it wouldn't make sense.

"Homogeneous" and "heterogeneous" are big, obscure words — I don't think they improve the clarity.

SethTisue:
I have one more idea: replace "each with a distinct type" with "each with its own type"?

manishbansal8843:
> "Homogeneous" and "heterogeneous" are big, obscure words — I don't think they improve the clarity

I was referring [this ](https://docs.scala-lang.org/overviews/scala-book/tuples.html) document for using these keywords. And same also needs to be updated i believe with the keyword "homogeneous".

SethTisue:
> I was referring this document for using these keywords

I don't like it there either :-)

SethTisue:
I went with "each with its own type" in 70a11663c67152026eb658730ea9f7d1b40901b7

Issue #1730, opened 2020-07-05
covariance definition is wrong


manishbansal8843:
Definition of covariance mentioned [here](https://docs.scala-lang.org/tour/variances.html) is completely opposite of what it is supposed to be. Same should be fixed.

PR #1731, opened 2020-07-05
Corrected covariance definition

manishbansal8843:
Fix for #1730

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -20,7 +20,7 @@ class Baz[A]  // An invariant class
 
 ### Covariance
 
-A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `A` is a subtype of `B`, then `List[A]` is a subtype of `List[B]`. This allows us to make very useful and intuitive subtyping relationships using generics.
+A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `B` is a subtype of `A`, then `List[B]` is a subtype of `List[A]`. This allows us to make very useful and intuitive subtyping relationships using generics.
 
 Consider this simple class structure:
 
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -20,7 +20,7 @@ class Baz[A]  // An invariant class
 
 ### Covariance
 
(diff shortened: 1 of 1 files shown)


_tour/variances.md
```diff
@@ -20,7 +20,7 @@ class Baz[A]  // An invariant class
 
 ### Covariance
 
-A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `A` is a subtype of `B`, then `List[A]` is a subtype of `List[B]`. This allows us to make very useful and intuitive subtyping relationships using generics.
+A type parameter `A` of a generic class can be made covariant by using the annotation `+A`. For some `class List[+A]`, making  `A` covariant implies that for two types `A` and `B` where `B` is a subtype of `A`, then `List[B]` is a subtype of `List[A]`. This allows us to make very useful and intuitive subtyping relationships using generics.
```
julienrf, line 23: This definition is equivalent to the previous one. You only have renamed `A` to `B` and `B` to `A`. We could use whatever names we want, for instance `X` and `Y`, the definition would be the same. What matters is that when the first is a the subtype of the second, then `List` of the first is also a subtype of `List` of the second.

Contravariance is in the other side: when the first of two type variables is a subtype of the second, then `List` of the second is a subtype of `List` of the first.

Does that make sense?

manishbansal8843: I agree with your definition of contravariant. However, in case of covariant definition, i did not just flip all A with B and vice versa. Originally, it was saying, if we have List[+A], and A is subtype of B then List[A] is also subtype of List[B]. This essentially says if we have List[+Cat] then List[Cat] is subtype of List [Animal]. 

Now, what i want to say is if we have List[+Animal] then List [Cat] is subtype of List [Animal]. 

Original definition is a mix of covariant and contravariant.

julienrf: I think the confusion comes from the fact that we use the same type variable name `A` in two cases although they are unrelated. When we define the type `List[+A]`, we say that `List` takes one type parameter (named `A`) that is covariant. Then, what we do is that we consider two (**other**) type variables `A` and `B` where `B <: A`, and state that `List[B] <: List[A]`. But the `A` of the second part is unrelated to the `A` of the first part.

What we could do to clear out the confusion would be to use different type variable names. For instance:

```suggestion
A type parameter `T` of a generic class can be made covariant by using the annotation `+T`. For some `class List[+T]`, making  `T` covariant implies that for two types `A` and `B` where `B` is a subtype of `A`, then `List[B]` is a subtype of `List[A]`. This allows us to make very useful and intuitive subtyping relationships using generics.
```

What do you think?

manishbansal8843: Yes. This is fine with me. Luckily, the co-incidence of using same name variables worked perfectly in case on contravariant definition. :)

PR #1732, opened 2020-07-06
Avoid deprecation warning

jorgarga:
This change fixes the following compiler warning:
  procedure syntax is deprecated for constructors: add `=`, as in method definition


SethTisue:
wdyt about also removing the curly braces?

jorgarga:
Thanks for your feedback!
I would leave the curly braces for different reasons:

- They were there already (I only added the = symbol to avoid a compiler warning).
- I am **completely** new to Scala and the braces help me identifying the method.
- I am not yet familiar with code style/language idioms. So I wouldn't know what is "better".

SethTisue:
okay cool, merging as is then! thank you

PR #1747, opened 2020-08-02
Modify example code to follow Scala coding conventions

ShapelessCat:
In the example code at the end of the doc of string-interpolation:
- Append `()` to function invocations which have side effects
- Replace `StringBuffer` with `StringBuilder`


SethTisue:
Thank you! I added a further commit that replaces `buf append ...` with `buf.append(...)`

PR #1749, opened 2020-08-04
[zh] Update collection overview

jxnu-liguobin:
Make it easy to read with chinese


eed3si9n:
@jxnu-liguobin Thanks for the contribution!

I can't review this myself so could you find someone who can? /cc @scala/docs-zh

PR #1755, opened 2020-08-13
Highlight deprecation of early definitions for initialization order

ollyw:
Document that early definitions are deprecated and going to be obsolete due to Tasty compatibly. See
https://github.com/scala/scala-dev/issues/513 for more details. I did wonder if removing the section on using early definitions entirely, but figured people will still be able to use it for a while and it is code that is seen out there in real code bases, so just a deprecation warning made sense.

### DIFF
--- a/_overviews/FAQ/initialization-order.md
+++ b/_overviews/FAQ/initialization-order.md
@@ -124,7 +124,7 @@ Note that using multiple lazy vals creates a new risk: cycles among lazy vals ca
 
 Early definitions are a bit unwieldy, there are limitations as to what can appear and what can be referenced in an early definitions block, and they don't compose as well as lazy vals: but if a lazy val is undesirable, they present another option.  They are specified in SLS 5.1.6.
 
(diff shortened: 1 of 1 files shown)


_overviews/FAQ/initialization-order.md
```diff
@@ -124,6 +124,8 @@ Note that using multiple lazy vals creates a new risk: cycles among lazy vals ca
 
 Early definitions are a bit unwieldy, there are limitations as to what can appear and what can be referenced in an early definitions block, and they don't compose as well as lazy vals: but if a lazy val is undesirable, they present another option.  They are specified in SLS 5.1.6.
 
+Note that early defintions are deprecated in Scala 2.13 and will be removed in 2.14; they will be replaced by trait parameters in 3.0. This option is not recommended if future compatibility is a concern.
```
SethTisue, line 127: typo: defintions

ollyw: Thanks for pointing that out @SethTisue.  Fixed in a new push

SethTisue:
Thank you for the improvement!

2.14 is no longer planned (https://scala-lang.org/2019/12/18/road-to-scala-3.html); I've added a commit that reflects that.

PR #1757, opened 2020-08-21
More detail about how to replace MutableList

winitzki:
The previous docs suggest to use ArrayDeque, which is not available before Scala 2.13. It is easier to use mutable.ListBuffer as a replacement solution, which will work equally in pre-2.13 and in 2.13.


SethTisue:
thank you!

PR #1758, opened 2020-08-21
Add options entries to the polish cheatsheet

Unrooted:
This PR adds options entries to the polish cheatsheet and also changes the style of it.


Unrooted:
I guess we can close this since this commit is also added to the #1760.

PR #1768, opened 2020-09-04
Update Vector-class.md

manishbansal8843:
Corrected typo. Also changed operator from `++` to `++:` for consistency. It might be confusing for beginners.

### DIFF
--- a/_overviews/scala-book/vector-class.md
+++ b/_overviews/scala-book/vector-class.md
@@ -45,20 +45,20 @@ val b = a :+ 4
 and this:
 
 ```scala
-val b = a ++ Vector(4, 5)
+val b = a ++: Vector(4, 5)
 ```
 
 The REPL shows how this works:
 
 ```scala
 scala> val a = Vector(1,2,3)
-a: Vector[Int] = List(1, 2, 3)
+a: Vector[Int] = Vector(1, 2, 3)
 
 scala> val b = a :+ 4
-b: Vector[Int] = List(1, 2, 3, 4)
+b: Vector[Int] = Vector(1, 2, 3, 4)
 
-scala> val b = a ++ Vector(4, 5)
-b: Vector[Int] = List(1, 2, 3, 4, 5)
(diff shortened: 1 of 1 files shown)


SethTisue:
I agree with the `List` to `Vector` change

I don't understand the motivation for changing `++` to `++:`?

manishbansal8843:
As per API [documentation](https://www.scala-lang.org/api/current/scala/collection/immutable/Vector.html), `++` is an alias for `concat` where as `++:` is an alias for `prependedAll`. Since, the current documentation is talking about pre-pending all the elements to a vector, I feel it is more appropriate to use `++:` rather than `++` although result would be same.

SethTisue:
I routinely see `++` in Scala code (in the books, in talks, etc) and I wasn't even aware `++:` existed. I think this is fine as-is.

I'd be happy to merge this if you reduce it to just the `List` vs `Vector` change

manishbansal8843:
@SethTisue Changed `++:` back to `++`

som-snytt:
Apparently, it is confusing, indeed. The comment that the doc was prepending is wrong. There is a paragraph about appending, followed by another about prepending.

What's the difference between `concat` and `appendedAll` again? There is probably not a simple way to make it unconfusing.

I'd suggest squashing away the confused commit.

martijnhoekstra:
I understand the motivation to write `prefix ++: a`, as `++:` is the operation that prepends to `a`. Conceptually, `prefix ++ a` is something different from `prefix ++: a`. That becomes more apparent if `prefix: List[Int]` instead of `prefix: Vector[Int]`: `prefix ++ a` then has a different outcome than `prefix ++: a`

But I really don't know whether or not that's better in the context of teaching materials and contrasting prepending and appending.

Anyway, the PR as it stands now is definitely an improvement to the previous situation, regardless of whether the bikeshed is appended to the building, or the building prepended to the bikeshed.

SethTisue:
thank you @manishbansal8843

PR #1769, opened 2020-09-04
Update collections-methods.md

manishbansal8843:
Removed methods from the list as same have not been explained in the current document.


SethTisue:
good catch — thank you!

PR #1770, opened 2020-09-04
Update tuples.md

manishbansal8843:
Added missing statements

### DIFF
--- a/_overviews/scala-book/tuples.md
+++ b/_overviews/scala-book/tuples.md
@@ -76,6 +76,11 @@ scala> val(x, y, z) = (3, "Three", new Person("David"))
 x: Int = 3
 y: String = Three
 z: Person = Person(David)
+
+scala> print(x)
+3
+scala> print(y)
+Three
 ```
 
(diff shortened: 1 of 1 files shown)


SethTisue:
I don't see what this adds, since the values were already printed?

manishbansal8843:
The values printed were part of scala REPL response. It does not show the usage as such.

SethTisue:
I don't see that as a problem. (Anyone else watching the repo have an opinion?)

SethTisue:
closing as I don't think this is mergeable as-is, but we could reopen if you want to revise it to take a different approach

som-snytt:
It's still hacktoberfest, so this PR might still be relevant. I do see that the intention is to show that tuple deconstruction in a pattern actually creates variables you can use in the ordinary way. But I agree that a better example is needed. Like maybe using an `age` variable. The example `(x, y, z)` is not very meaningful except as coordinates. Arguably leaving no space after `val` is poor style.
```
scala> val(x, y, z) = (3, "Three", new Person("David"))
val x: Int = 3
val y: String = Three
val z: Person = Person(David)
```
It's also a missed opportunity to standardize on tupple or toople.

PR #1773, opened 2020-09-08

macginitie:
The instructions as written did not work for me today, so I edited them so they would have. Presumably IntelliJ has changed its behavior since they were written. Note also, they are still off: "To the right of the Scala SDK field, click the **Create** button." There is no such field or button at that point, I assume there was in older versions of IntelliJ. I don't know how best to fix that issue.


SethTisue:
is there an IntelliJ user who could review this?

SethTisue:
pinging repo watchers — could an IntelliJ user review this, please?

nafg:
Correct, the option only appears when the menu is opened within a sources directory, which in a typical maven/sbt project layout is src/{main,test}/scala. I believe Mill and Seed projects tend to just use src but those are newer. I'm not sure why it says just src in the first place.

SethTisue:
thank you Andrew and Naftoli!

PR #1778, opened 2020-09-18
Add trait with type to foldLeft def

babo:
The original example is hard to digest as it's not clear where the A type is from.

### DIFF
--- a/_tour/multiple-parameter-lists.md
+++ b/_tour/multiple-parameter-lists.md
@@ -17,7 +17,11 @@ Methods may have multiple parameter lists.
 Here is an example, as defined on the `TraversableOnce` trait in Scala's collections API:
 
 ```
-def foldLeft[B](z: B)(op: (B, A) => B): B
+trait TraversableOnce[A] {
+  ...
+  def foldLeft[B](z: B)(op: (B, A) => B): B
+  ...
+}
 ```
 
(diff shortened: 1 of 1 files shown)


_tour/multiple-parameter-lists.md
```diff
@@ -17,7 +17,11 @@ Methods may have multiple parameter lists.
 Here is an example, as defined on the `TraversableOnce` trait in Scala's collections API:
```
julienrf, line 17:
```suggestion
Here is an example, as defined on the `Iterable` trait in Scala's collections API:
```

_tour/multiple-parameter-lists.md
```diff
@@ -17,7 +17,11 @@ Methods may have multiple parameter lists.
...
 
 ```
-def foldLeft[B](z: B)(op: (B, A) => B): B
+trait TraversableOnce[A] {
```
julienrf, line 20:
```suggestion
trait Iterable[A] {
```

julienrf (commented):
Thank you! Here are some suggestions since `TraversableOnce` has been removed from Scala 2.13.

babo:
Thanks, good point, that explains while I failed to search the original source of the trait!

julienrf (approved):
Thank you @babo! :+1:

PR #1779, opened 2020-09-18
Removed mention of String in FAQ about Context Bounds

xko:
Mention of String doesn't make sense without [View Bound example](https://stackoverflow.com/questions/4465948/what-are-scala-context-and-view-bounds) which is no longer here.


SethTisue:
thanks!

PR #1782, opened 2020-09-23
Added empty initial Overview files

alvinj:
I just created these initial files so that Jonathan and I can get started on the writing process.

### DIFF
--- a/_overviews/overview/a-taste-of-scala.md
+++ b/_overviews/overview/a-taste-of-scala.md
@@ -3,15 +3,18 @@ title: A Taste of Scala
 description: This page provides a high-level overview of the main features of Scala 3.
 ---
 
-<!-- TODO
+<!-- TODO/QUESTIONS
   - add ScalaFiddle at the appropriate places
-  - test with jekyll on my system
+  - which Control Syntax to use by default?
+  - demo “named parameters” with methods
+  - show a method with default parameters
+  - show package "exports"
+  - cover Option/Try/Either (functional error handling)
 -->
 
-<!-- TODO: this is a choppy start -->
-Our hope in this Overview documentation is to demonstrate that Scala is a beautiful, expressive programming language, with a clean, modern syntax. To help with that demonstration, this “Taste of Scala” section provides a whirlwind tour of Scala’s main features. After the initial tour in this section, the rest of the Overview will provide a few more details on the these features, and the Reference documentation will provide *many* more details.
+This “Taste of Scala” section provides a whirlwind tour of Scala’s main features. After the initial tour in this section, the rest of the Overview will provide a few more details on the these features, and the Reference documentation will provide *many* more details.
 
->In this Overview it’s assumed that you’ve used a language like Java before, and you’re ready to see a series of Scala examples to get a feel for what the language looks like. You’ll also be able to test many of the examples directly on this page, and in addition to that, you can also test anything you’d like on [ScalaFiddle.io](https://scalafiddle.io), [Scastie](https://scastie.scala-lang.org), in Worksheets with the VS Code or Intellij IDEA editors, or in the Scala REPL, which will be demonstrated shortly.
+>Throughout this Overview you’ll also be able to test many of the examples directly on this page. In addition to that, you can also test anything you’d like on [ScalaFiddle.io](https://scalafiddle.io), [Scastie](https://scastie.scala-lang.org), or in the Scala REPL, which is demonstrated shortly.
 
 
 
@@ -19,13 +22,13 @@ Our hope in this Overview documentation is to demonstrate that Scala is a beauti
 
 <!-- NOTE: i assume here that dotc/dotr will be renamed to scala/scala -->
 
-Ever since the book, *C Programming Language*, it’s been a tradition to begin programming books with a “Hello, world” example, and not to disappoint, this is one way to write that example in Scala:
+A “Hello, world” example in Scala goes as follows. First, Put this code in a file named *Hello.scala*:
 
 ```scala
 @main def hello = println("Hello, world")
 ```
 
-To see how this works, put that line of code in a file named *Hello.scala*, and then compile it with `scalac`:
+Next, compile the code with `scalac`:
 
 ```sh
 $ scalac Hello.scala
@@ -44,8 +47,7 @@ hello.class
 hello.tasty
 ```
 
-<!-- TODO: wording -->
-Just like Java, the *.class* files are bytecode files, and they’re ready to run in the JVM. Now you can run the main `hello` method with the `scala` command:
+Like Java, the *.class* files are bytecode files, and they’re ready to run in the JVM. Now you can run the main `hello` method with the `scala` command:
 
 ```sh
 $ scala hello
@@ -53,15 +55,12 @@ Hello, world
 ```
 
 Assuming that worked, congratulations, you just compiled and ran your first Scala application.
-<!-- TODO: more explanation on how that works? -->
 
 
 
 ## The Scala REPL
 
-The Scala REPL (“Read-Evaluate-Print-Loop”) is a command-line interpreter that you use as a “playground” area to test your Scala code. We introduce it early here so you can use it with the code examples that follow.
-
-Assuming that you’ve downloaded and installed Scala — such as from the [Scala download page](https://www.scala-lang.org/download) — you start a REPL session by running the `scala` command at your operating system command line. When you do this you’ll see a “welcome” prompt that looks like this:
+The Scala REPL (“Read-Evaluate-Print-Loop”) is a command-line interpreter that you use as a “playground” area to test your Scala code. You start a REPL session by running the `scala` command at your operating system command line. When you do so, you’ll see a “welcome” prompt like this:
 
 <!-- TODO: update this when it’s ready -->
 ```scala
@@ -72,19 +71,36 @@ Type in expressions for evaluation. Or try :help.
 scala> _
 ```
 
-Because the REPL is a command-line interpreter, it sits there waiting for you to type something. Inside the REPL you can type Scala expressions to see how they work:
+The REPL is a command-line interpreter, so it sits there waiting for you to type something. Now you can type Scala expressions to see how they work:
 
-```scala
-scala> val x = 1
-x: Int = 1
+````
+scala> 1 + 1
+val res0: Int = 2
 
-scala> val y = x + 1
-y: Int = 2
-```
+scala> 2 + 2
+val res1: Int = 4
+````
 
-As those examples show, after you type your expressions in the REPL, it shows the result of each expression on the line following the prompt.
+As shown in the output, if you don’t assign a variable to the result of an expression, the REPL creates variables named `res0`, `res1`, etc., for you. You can use these variable names in future expressions:
 
-As mentioned earlier, if you prefer a browser-based testing environment you can also use [ScalaFiddle.io](https://scalafiddle.io) or [Scastie.scala-lang.org](https://scastie.scala-lang.org).
+````
+scala> val x = res0 * 10
+val x: Int = 20
+````
+
+Notice that the REPL output also shows the result of your expressions.
+
+You can run all sorts of experiments in the REPL. This example shows how to create and then call a `sum` function:
+
+````
+scala> def sum(a: Int, b: Int): Int = a + b
+def sum(a: Int, b: Int): Int
+
+scala> sum(2, 2)
+val res2: Int = 4
+````
+
+As mentioned earlier, if you prefer a browser-based playground environment, you can also use [ScalaFiddle.io](https://scalafiddle.io) or [Scastie.scala-lang.org](https://scastie.scala-lang.org).
 
 
 
@@ -92,17 +108,20 @@ As mentioned earlier, if you prefer a browser-based testing environment you can 
 
 When you create a new variable in Scala, you declare whether the variable is immutable or mutable:
 
-- `val` is an *immutable* variable — like `final` in Java. We recommend always creating a variable with `val`, unless there’s a specific reason you need a mutable variable.
-- `var` creates a *mutable* variable, and should only be used when you have a variable whose contents will be modified over time.
+- `val` creates an *immutable* variable — like `final` in Java. You should always create a variable with `val`, unless there’s a reason you need a mutable variable.
+- `var` creates a *mutable* variable, and should only be used when a variable’s contents will change over time.
 
 These examples show how to create `val` and `var` variables:
 
 ```scala
-val x = 1   // immutable
-var y = 0   // mutable
+val a = 0         // immutable
+val b = "Hello"
+
+var c = 1         // mutable
+var d = "world"
 ```
 
-In an application, once you create a `val` field, it can’t be reassigned. The second line of code here creates an error message:
+In an application, a `val` field can’t be reassigned. You’ll generate a compiler error if you try to reassign one:
 
 ```scala
 val msg = "Hello, world"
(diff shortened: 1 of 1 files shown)


_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+-->
+
+
+
+## Data Modeling
+
+Scala supports both functional programming (FP) and object-oriented programming (OOP), as well as a fusion of the two paradigms. This section provides a quick overview of data modeling in OOP and FP.
+
+
+### OOP data modeling (traits and classes)
```
julienrf, line 568: I might be nitpicky but for me in OOP we don’t really model _data_. It’s more about modeling operations, imho.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+#### Traits
+<!--
+- traits as interface
+- implemented methods
+- class extends traits
+- traits as mixins
+-->
+
+Traits are like interfaces in other languages, but they can also contain implemented methods. They provide a great way for you to organize behaviors into small, modular units. When you want to create concrete implementations of attributes and behaviors, classes and objects can extend traits, mixing in as many traits as needed to achieve the desired behavior.
```
julienrf, line 580: I think interfaces can contain implemented methods in most modern languages?

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+trait Runner:
+  def startRunning(): Unit = println("I’m running")
+  def stopRunning(): Unit = println("Stopped running")
+```
+
+Given those traits, here’s a `Dog` class that extends all of those traits while providing a behavior for the abstract `speak` method:
+
+```scala
+class Dog(name: String) extends Speaker with TailWagger with Runner:
```
julienrf, line 600:
```suggestion
class Dog(name: String) extends Speaker, TailWagger, Runner:
```

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+  def speak(): String = "Woof!"
+```
+
+Notice how the class extends the traits with the `extends` and `with` keywords.
+
+Similarly, here’s a `Cat` class that implements those same traits while also overriding two of the concrete methods it inherits:
+
+```scala
+class Cat(name: String) extends Speaker with TailWagger with Runner:
```
julienrf, line 609:
```suggestion
class Cat(name: String) extends Speaker, TailWagger, Runner:
```

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+c.stopRunning()         // "No need to stop"
+```
+
+<!-- TODO: links  -->
+If that code makes sense — great, you’re comfortable with traits as interfaces. If not, don’t worry, they’re explained in more detail in the Data Modeling sections of this Overview and the Reference documentation.
+
+<!-- TODO: show traits that take parameters here? -->
+
+
+#### Classes
```
julienrf, line 633: Note that we have used class definitions above without introducing the concept of class.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+  def printFullName() = println(s"$firstName $lastName")
+
+val p = Person("Julia", "Kern")
+println(p.firstName)   // "Julia"
+p.lastName = "Manes"
+p.printFullName()      // "Julia Manes"
+```
+
+
+### FP data modeling
```
julienrf, line 661: I ithink it would be nice to write “functional programming” rather than “FP”. What do you think of the following?

```suggestion
### Data Modeling (Functional Programming Style)
```

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+Enums are covered in detail in the Data Modeling section of this Overview, and in the Reference documentation.
+
+
+#### Case classes
+
+A *case class* is an extension of the base Scala class. Case classes provide features that make them useful for functional programming. They have all of the functionality of a regular class, and more. When the compiler sees the `case` keyword in front of a `class`, it generates code for you, with these benefits:
+
+- Case class constructor parameters are public `val` fields by default, so accessor methods are generated for each parameter.
+- An `apply` method is created in the companion object of the class, so you don’t need to use the `new` keyword to create a new instance of the class.
```
julienrf, line 737: We don’t need `new` anymore for regular classes as well: https://dotty.epfl.ch/docs/reference/other-new-features/creator-applications.html

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+- An `unapply` method is generated, which lets you use case classes in more ways in `match` expressions.
+- A `copy` method is generated in the class. This provides a way to clone an object while making updates to its values during the cloning process.
+- `equals` and `hashCode` methods are generated.
```
julienrf, line 740: I think this is the most important difference between classes and case classes, and for me this is often the criterion that determines whether I need a class or a case class. I believe we should go further and explain what does the equality mean on case classes vs regular classes.
Or maybe we do it later in the overview?

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+  vocation: String
+)
+
+// create an instance of the case class
+val p = Person("Reginald Kenneth Dwight", "Singer")
+
+// a good default toString method
+p                // Person = Person(Reginald Kenneth Dwight,Singer)
+
+// can access its fields, but they are immutable
```
julienrf, line 758:
```suggestion
// can access its fields, which are immutable
```

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+// when you need to make a change, use the `copy` method
+// to “update as you copy”
+val p2 = p.copy(name = "Elton John")
+p2               // Person = Person(Elton John,Singer)
+```
+
+See the Data Modeling sections of this Overview and the Reference documentation for many more details on case classes.
+
+
+
```
julienrf, line 771: I would expect a mention of sealed traits in the “FP data modeling” section

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+- Top-level functions
+- Functions in objects
+- Lambdas
+- Higher-order functions (HOFs)
+
+<!-- TODO: when to say “function” vs when to say “method” -->
+<!-- TODO: show the `val` function syntax? -->
+
+
+###  Top-level functions
```
julienrf, line 833: I’m not sure we should have this section in the Taste page. Instead, I suggest a “Top-level definitions” page in the overview.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+Functions don’t have to be in the body of a class, enum, or object. They can also exist on their own as a *top-level function*. For instance, when this code is saved in a file, the functions `add` and `double` would be called top-level functions because they exist outside the scope of a class, enum, or object:
+
+```scala
+package foo
+
+def add(a: Int, b: Int): Int = a + b
+def double(a: Int): Int = 2 * a
```
julienrf, line 841: These are methods.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+@main def topLevelFunctions() =
+  val x = add(1, 2)
+  val y = double(x)
+  println(s"y = $y")
+```
+
+
+### Functions in objects
```
julienrf, line 850: This example illustrates methods, we could move it up to the “methods” section?

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+
+## Extension methods
+
+<!-- TODO: i may not be showing the latest syntax -->
+*Extension methods* let you add new methods to closed classes. For instance, if you want to add a `hello` method to the `String` class, just create an extension method:
+
+<!-- TODO: scalafiddle -->
+```scala
+def (s: String) hello: String = s"Hello, ${s.capitalize}"
```
julienrf, line 946: The syntax has changed again, IIRC it should look like the following:

```suggestion
extension (s: String)
  def hello: String = s"Hello, ${s.capitalize}"
```

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+```scala
+def (s: String) makeInt(radix: Int): Int = Integer.parseInt(s, radix)
+"1".makeInt(2)      // Int = 1
+"10".makeInt(2)     // Int = 2
+"100".makeInt(2)    // Int = 4
+```
+
+
+
+## Collections classes
```
julienrf, line 975: I’m not sure we should go into these details here. I would simply say that Scala provides a rich set of collections, which have a rich set of methods, and show a few examples, but without showing the tables listing the collection classes and the collection methods.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+Tuples are nice for those times when you want to put a collection of heterogenous types in a little collection-like structure.
+
+<!--
+TODO: state how tuples can be like an unstructured class, or for when a class is overkill?
+-->
+
+
+
+## Contextual Abstractions
```
julienrf, line 1177: I think this section is too detailed too, but I’m not sure how we should present the benefits of implicits without going into all these explanations.

What do you think of the following?

---

Under certain circumstances, the Scala compiler can “write” some parts of your programs. For instance, consider a program that sorts a list of addresses by two criterias: city name and then street name:

~~~ scala
val addresses: List[Address] = ...
addresses.sortBy(address => (address.city, address.street))
~~~

The sorting algorithm needs to compare addresses by first comparing their city names, and then also their street names when the city names are the same. However, we don’t need to manually define this ordering relation because the compiler is able to summon it automatically based on an existing ordering relation for comparing `String` values.

---

What do you think?

julienrf (commented):
Thank you @alvinj :)

I really like how you explain basic things. However, I often think we give too much information for a “Taste of Scala” (but others may have a different opinion). Please see my comments below.

Also, please don’t open PRs against `scala/docs.scala-lang:master` but create a branch on `scalacenter/docs.scala-lang.org` instead. (for instance `scala-3`)

alvinj:
> Also, please don’t open PRs against `scala/docs.scala-lang:master`

I’m *very* sorry if I put this on the master. I thought I created a branch called `overview` and did my work there. I’m currently wrestling with a printer/scanner, but when I finish with that I’ll look into this.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+val p = Person("Margaret Hamilton")
+
+// a class field or method
+println(s"Name: ${p.name}")   // "Name: Margaret Hamilton"
+
+// an equation
+println(s"2 + 2 = ${2 + 2}")   // prints "2 + 2 = 4"
+```
+
+#### Other interpolators
```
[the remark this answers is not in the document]
alvinj: Yes, that’s a good idea. My only reason for mentioning them is that I remember the first time I saw the `s` before a string I thought it was a misprint in a book. So I like to mention it, but yes, the details can come later.

_overviews/overview/a-taste-of-scala.md
alvinj: One of my questions for you (all) is, what is the difference between a method and a function, or when should I use those two words? I know that technically a function is defined as a `val`, but is there ever a time when a `def` method should be referred to as a function?

_overviews/overview/a-taste-of-scala.md
alvinj: I’ll update this.

_overviews/overview/a-taste-of-scala.md
```diff
@@ -0,0 +1,1329 @@
...
+
+In your code you can also use the characters `L`, `D`, and `F` (and their lowercase equivalents) to specify `Long`, `Double`, and `Float` fields:
+
+```scala
+val x = 1_000L   // val x: Long = 1000
+val y = 2.2D     // val y: Double = 2.2
+val z = 3.3F     // val z: Float = 3.3
+```
+
+When you need really large numbers, use the `BigInt` and `BigDecimal` types:
```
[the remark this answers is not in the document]
alvinj: I’ll update this.

_overviews/overview/a-taste-of-scala.md
alvinj: Yes, that’s true. I’ll update this.

_overviews/overview/a-taste-of-scala.md
alvinj: Yes, I have a little bit of a “chicken and the egg” thing going on. I’ll try to resolve this.

_overviews/overview/a-taste-of-scala.md
alvinj: I like it!

_overviews/overview/a-taste-of-scala.md
alvinj: Thanks, I’ll take care of this. I knew that `new` wasn’t needed, but I was debating about what to do here. I’ll just delete that line.

_overviews/overview/a-taste-of-scala.md
alvinj: That’s a good point, thanks! I think that sounds like a detail I should refer readers to later in the overview. I’ll add something here.

I remember reading in *Programming in Scala*, that “The biggest advantage of case classes is that they support pattern matching.” I debated about including that here, and then I thought it was a detail I should discuss later.

_overviews/overview/a-taste-of-scala.md
alvinj: Thanks, I’ll add this.

_overviews/overview/a-taste-of-scala.md
alvinj: Yes, if I did anything here I probably should have made a small “Top-level definitions” subsection. I had a mistake in my notes and I *thought* this was something that @b-studios suggested for this section, but I see that it was something I added in myself.

I did think that this was a nice new feature of Scala 3 to highlight. Do you think a “Top-level definitions” subsection is worth having in the Taste, or just have this as a later section in the Overview? If I don’t have that as a subsection here, I can refer readers to that section earlier on when I show the `@main def...` syntax.

_overviews/overview/a-taste-of-scala.md
alvinj: Yes, I’ll move it up there.

_overviews/overview/a-taste-of-scala.md
alvinj: Thanks. @b-studios caught this one, and I think I’m going to do something like this here:

```scala
extension (s: String):
  def hello: String = s"Hello, ${s.capitalize}"
  def aloha: String = s"Aloha, ${s.capitalize}"

"world".hello    // "Hello, World"
"friend".aloha   // "Aloha, Friend"
```

_overviews/overview/a-taste-of-scala.md
alvinj: I agree, those are too many details here. I’ll move those to the later Collections section. That will definitely help to shorten this — thanks!

alvinj:
> Why do you call this "sequence" methods? Because they are defined on `Seq`? I think at this high-level I would just talk about "collections".

@b-studios, there were probably too many thoughts going on in my head at this time. I was thinking about trying to avoid a discussion of the trait/class hierarchy here. I was also going to make a distinction between sequences and maps, so I was thinking about “methods available on sequences.” But I like @julienrf’s suggestion to simplify this section. When I do that, I’ll get rid of that terminology.

_overviews/overview/a-taste-of-scala.md
alvinj: Yes, this is a tough one. I’ll think about this some more, but I do like what you have here. It describes what can happen without getting into the details, and it’s motivational in that it makes me want to click a “Learn more about this feature” link in this section.

_overviews/overview/a-taste-of-scala.md
julienrf: The difference is purely technical: a function is a value (of type `A => B`), whereas a method is a member on a type. Functions come from FP, whereas methods come from OOP. But, in the end, they do the same: computing something for some input parameters. This is why it is easy to convert a method to a function.

PR #1800, opened 2020-10-22
Copying description is unclear

palashahuja:
It looks like `message4.copy` creates a deep copy instead of a shallow copy. I wasn't really sure so I tested this using some sample code, as well and it pretty much looks like a deep copy. (Please be kind, I am a newbie!)
```
case class Message(sender: String, var recipient: String, body: String)
val message4 = Message("julien@bretagne.fr", "travis@washington.us", "Me zo o komz gant ma amezeg")
val message5 = message4.copy(sender = message4.recipient, recipient = "claire@bourgogne.fr")
message5.sender  // travis@washington.us
message5.recipient // claire@bourgogne.fr
message5.body  // "Me zo o komz gant ma amezeg"
message4.recipient = "abc@new.ny"
message5.sender // travis@washington.us
```

### DIFF
--- a/_tour/case-classes.md
+++ b/_tour/case-classes.md
@@ -44,7 +44,7 @@ val messagesAreTheSame = message2 == message3  // true
 Even though `message2` and `message3` refer to different objects, the value of each object is equal.
 
 ## Copying
-You can create a (shallow) copy of an instance of a case class simply by using the `copy` method. You can optionally change the constructor arguments.
+You can create a (deep) copy of an instance of a case class simply by using the `copy` method. You can optionally change the constructor arguments.
 ```tut
 case class Message(sender: String, recipient: String, body: String)
 val message4 = Message("julien@bretagne.fr", "travis@washington.us", "Me zo o komz gant ma amezeg")


SethTisue:
What do deep vs. shallow mean to you? It's not clear to me what in the transcript you've provided would lead you to describe this as a "deep" copy.

The reason I think describing it as a shallow copy is correct is that `message4.body eq message5.body` is `true`. The fields that weren't replaced contain the exact same references.

som-snytt:
Jinx. You might try adding the clarifying experiment.
```
scala> case class X(n: Int) ; case class C(x: X)
class X
class C

scala> val c = C(X(42)) ; c.x eq c.copy().x
val c: C = C(X(42))
val res1: Boolean = true

scala> val c = C(X(42)) ; c.x eq c.copy(c.x.copy()).x
val c: C = C(X(42))
val res3: Boolean = false
```

palashahuja:
> What do deep vs. shallow mean to you? It's not clear to me what in the transcript you've provided would lead you to describe this as a "deep" copy. Can you explain your reasoning?
> 
> The reason I think describing it as a shallow copy is correct is that `message4.body eq message5.body` is `true`. The fields that weren't replaced contain the exact same references. No fresh copy of the string in `message4.body` is made.

I just feel that the description is unclear. It led me to believe that message5 is a shallow copy of message4, which is not the case. I did not consider that the body attribute is being referenced by both the objects.

som-snytt:
I can see that the example leaves a few things unexplained.

It might help to show `msg.copy()` first, then `msg.copy(recipient = "claire@bourgogne.fr")`.

It's "only the tour", but the link to the book is just a bit clearer. It side-steps deep vs shallow by perhaps incorrectly calling it a clone. The example is more fun but also arguably more culturally-specific. As in Chicago culture. I would update the example to use the Boston Red Sox. Or a `Player` class with data `"Mookie Betts"`, then `betts.copy(team = Dodgers)`.

PR #1803, opened 2020-10-27
Fixed traits.md file error

artemkorsakov:
The `/tour/traits.html` previous page is `/tour/named-arguments.html` - not `/tour/classes.html` page!


SethTisue:
thank you!

PR #1804, opened 2020-10-27
Fixed default-parameter-values.md file error 

artemkorsakov:
The `/tour/default-parameter-values.html` previous page is `/tour/classes.html` - not `/tour/annotations.html` page!


SethTisue:
thank you!

PR #1813, opened 2020-11-10
remove postfix ops from style guide

SethTisue:
Scala 2.13 errors by default, so I don't think the style guide
needs to cover this any more


som-snytt:
It should go in the special "high style" guide.

PR #1817, opened 2020-11-16
Update generic-classes.md

Nexus01:
replace `{ elements = x :: elements }` with `: Unit = \n elements = x :: elements` to remove `warning: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `push`'s return type`


SethTisue:
thank you!

PR #1822, opened 2020-11-19
Style cleanup for value class restriction overview

ivan-klass:


SethTisue:
mind rebasing? that'll pull in #1832 and #1834, so then we'll see whether the CI failure here has anything to do with your changes or not

SethTisue:
thank you!

PR #1830, opened 2020-11-25
Edit variability explanations for clarify

raindev:
Avoid "substitute for" construct, which can easily be confusing and understood in opposite ways.

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -32,7 +32,7 @@ case class Cat(name: String) extends Animal
 case class Dog(name: String) extends Animal
 ```
 
-Both `Cat` and `Dog` are subtypes of `Animal`. The Scala standard library has a generic immutable `sealed abstract class List[+A]` class, where the type parameter `A` is covariant. This means that a `List[Cat]` is a `List[Animal]` and a `List[Dog]` is also a `List[Animal]`. Intuitively, it makes sense that a list of cats and a list of dogs are each lists of animals, and you should be able to substitute either of them for a `List[Animal]`.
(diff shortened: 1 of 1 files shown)


martijnhoekstra:
I always have to look up how substitute for works. The dictionary gives the example "Don't believe anyone who tries to convince you that low-fat cooking sprays can substitute for real butter or olive oil and not taste any different.", which I wholeheartedly agree with. Currently we're claiming here that you won't taste the difference substituting a `printer[Animal]` for a `printer[Cat]`. That's correct, and this PR switches it around.

I'd applaud a different PR that doesn't rely on "substitute for" to make it clearer rather than unclearer to what's used in place of what.

raindev:
Thanks for the answer Martijn. It's a lesson for me to double check before fixing a mistake that really should have been noticed by others. You're of course correct (from [Wiktionary](https://en.wiktionary.org/wiki/substitute#Usage_notes)):

> "Substitute butter for olive oil" = Use butter instead of olive oil
> "Substitute butter with olive oil" = Use olive oil instead of butter

I have attempted to replace "substitute for" by a simpler and hopefully clearer constructs.

SethTisue:
mind rebasing? that'll pull in #1832 and #1834, so then we'll see whether the CI failure here has anything to do with your changes or not

raindev:
Seth, I have rebased the PR. The pipeline have failed twice with the same symptoms (I don't think related to the changes):

> Exception in thread "main" scala.reflect.internal.MissingRequirementError: object scala in compiler mirror not found.

SethTisue:
I'll look into that separately, then — thank you!

PR #1835, opened 2020-11-28
Fix description of Failure

coverbeck:
The Failure contains the entire exception, not just the exception message.

See [source code](https://github.com/scala/scala/blob/6495d905b31fa6daf3c28a2c130835a238a2310f/src/library/scala/util/Try.scala#L215)


SethTisue:
thank you!

PR #1844, opened 2020-12-10
reboot the Scala FAQ

SethTisue:
gah, nobody has touched this thing in years. and, perhaps, because it's so neglected, nobody in our chat rooms and forums ever refers anyone here. (I exaggerate a bit, but it's mostly true IMO)

but people probably still find it through Google. and if we make it good, then perhaps people in chat rooms and forums _will_ refer people here more

so let's do something about it.  changes so far:

* don't duplicate content from Stack Overflow; this was a worthy but IMO failed experiment
* discourage long writeups; encourage links to answers elsewhere
* add a few general questions at the top that I know to be especially frequently asked
* drop the question about `breakOut`, it is not an FAQ these days

this PR is just a start, much room for improvement remains, but IMO this is already mergeable

The "More questions" thing is admittedly awkward -- it's the only answer in the old FAQ that wasn't just a duplicate of something from SO. Doing something about that could be future work.

PR #1845, opened 2020-12-10
add redirects for removed FAQ answer pages

SethTisue:
redirecting to the [main FAQ page](https://docs.scala-lang.org/tutorials/FAQ/index.html) is perhaps not the _ideal_ solution, perhaps we could have redirected to the SO answers the removed pages were based on, or redirected to the Wayback Machine, or I don't know what

...but personally I think this is good enough. merging on fast track since #1844 just got merged.


SethTisue:
I have verified, post-merge, that e.g. if I search "scala FAQ how does yield work" on Google, the top search result redirects as expected

PR #1846, opened 2020-12-10
fix up FAQ

SethTisue:
followup to #1844

There was some question about what to do with the one FAQ answer that wasn't just a copy of something from Stack Overflow. I've handled it by leaving the answer page in place but fixing the look of the questions page so this question is formatted like the others.

I hope the historically conscious will be amused by my retaining the old "one-question FAQ" as the first question in the technical-question section :-)

I think it's fine to grandfather in this one page, but I don't think we should add any more pages like this in the future.

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -65,6 +65,16 @@ get poor results, try surrounding the symbol with double quotes.
 
 ## Specific technical questions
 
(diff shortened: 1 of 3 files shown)


SethTisue:
Merging quickly because #1844 was already merged and because I want to publicize the page to contributors, but happy to respond to feedback and make further adjustments.

PR #1848, opened 2020-12-11
Adding more FAQs

BalmungSan:
I propose the addition of the following FAQs, each is independent of each other but I just thought it was easier to make a single PR with all and remove if necessary than making multiple PRs.

Note that I added them in my own personal order of importance, meaning that I am strongly opinionated on adding the first ones and thus not so sure if the last ones are worth it.


SethTisue:
This was already mergeable as-is IMO. But I added a commit that rewords some questions and adds brief answers before some of the links. wdyt?

SethTisue:
I was going to add a commit with some suggested changes, but it seems you didn't enable "Allow edits from maintainers"?

I'll just merge it, and then send a followup PR.

BalmungSan:
@SethTisue ups, sorry! It is usually enabled by default in all projects. Anyways thanks!

SethTisue:
sequel: https://github.com/scala/docs.scala-lang/pull/1849

PR #1849, opened 2020-12-11
FAQ: include short answers before linking to longer texts

SethTisue:
sequel to #1848. review by @BalmungSan?

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -79,7 +79,7 @@ official documentation that addresses it, not just an FAQ answer -->
 
 see the [Scala 2.13 Collections Guide](https://docs.scala-lang.org/overviews/collections-2.13/introduction.html)
 
-### What are Scala context bounds (`[T : Foo]`)?
+### What are context bounds (`[T : Foo]`)?
 
 [answer on Stack Overflow](https://stackoverflow.com/a/4467012)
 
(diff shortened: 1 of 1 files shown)


_overviews/FAQ/index.md
```diff
@@ -110,28 +110,56 @@ One would hope so, but doing it that way was tried and it proved
...
+Types are primarily a compile-time concept. At compile time,
+every expression is assigned a type by the compiler.
+
+Classes are primarily a runtime concept and are platform-dependent.
+At runtime on the JVM, every value is either a primitive value
+or an instance of exactly one class.
+
+Some type information exists only at compile time, for multiple
+reasons, most notoriously [type
+erasure](https://en.wikipedia.org/wiki/Type_erasure).
```
BalmungSan, line 138: Big +1 on this, probably on the very first JVM things I learned and it was due **Scala**.

_overviews/FAQ/index.md
```diff
@@ -110,28 +110,56 @@ One would hope so, but doing it that way was tried and it proved
...
-### What does <:< means?
+### What does `<:<` mean?
 
-[blog post: Generalized type constraints](https://blog.bruchez.name/2015/11/generalized-type-constraints-in-scala.html)
+It's a "type constraint", and it comes from the standard library,
+not from the language itself.  See [this blog
+post](https://blog.bruchez.name/2015/11/generalized-type-constraints-in-scala.html).
 
-### Best way to model optional arguments
+### I dislike requiring callers to wrap optional arguments in `Some(...)`; is there a better way?
```
BalmungSan, line 158: 😄

_overviews/FAQ/index.md
```diff
@@ -110,28 +110,56 @@ One would hope so, but doing it that way was tried and it proved
...
 
-[answer on Stack Overflow](https://stackoverflow.com/a/65256691/4111404)
+Not really. See [this answer on Stack Overflow](https://stackoverflow.com/a/65256691/4111404).
 
-### Why prfer implicit val over implicit object
+### Why is `implicit val` usually recommended over `implicit object`?
 
-[answer on Stack Overflow](https://stackoverflow.com/a/65258340/4111404)
+The latter has a singleton type, which is too specific.  See [answer on Stack
```
BalmungSan, line 164: Maybe a link to something that explains what is a singleton type?
Or maybe that link should be in the SO answer itself?

BalmungSan (approved):
I really like this, I even thought about doing that in my previous PR. I didn't because the previous ones were only links, which opens the question if we should do the same for the previous ones?

_overviews/FAQ/index.md
SethTisue: Sadly the language tour doesn't have a "Singleton" types section. I agree there's room for improvement here, but I'm not going to tackle it right now.

SethTisue:
> I really like this, I even thought about doing that in my previous PR. I didn't because the previous ones were only links, which opens the question if we should do the same for the previous ones?

I think we should. Perhaps I'll get to it in some future PR, and/or perhaps someone else will.

PR #1852, opened 2020-12-14
Adding method access control to the FAQ

BalmungSan:


SethTisue:
Is this actually an FAQ? I feel like it isn't, and also it seems extremely broad: “What can you tell me, generally, about this entire area of Scala's functionality?"

SethTisue:
It's a shame there isn't an "Access" section of https://docs.scala-lang.org/tour/tour-of-scala.html

BalmungSan:
@SethTisue actually that is a good criticism, it is not really a **FAQ** but rather _"basic knowledge"_ (whatever that means) that is often useful _([for example](https://www.reddit.com/r/scala/comments/k3uoqr/got_a_quick_question_ask_here_november_30_2020/gfqtwc3?utm_source=share&utm_medium=web2x&context=3))_. Being honest I agree with you that this should be part of the **Tour* rather than a **FAQ**.

I think the best is to close this, but the blog by Alvin can serve as inspiration for a section of the **Tour** 😉

PR #1860, opened 2020-12-20
Update for-comprehensions.md

yufengzh:
Directly passing `println` into `foreach` would suffice.


b-studios:
It is true, that this would suffice. However, I think eta-expanding it here makes it easier for beginners to understand what is going on. On the other hand, it requires readers to be able to read lambda syntax, which might be more complicated then passing `println`.

SethTisue:
I have no strong opinion either way...

SethTisue:
It's true that some readers might wonder _why_ writing it the short way it works, but I don't think they will wonder what it does, so I don't think it is likely to be a real distraction. So I'm going to go ahead and merge this. Thank you!

PR #1861, opened 2020-12-21
Scala 3 Documentation

b-studios:
Accompanying the release of Scala 3-M3, this PR adds a first initial draft of the Scala 3 documentation. Besides adding content, we also needed to adjust the styles to incorporate Scala 2 and Scala 3 documentation, while making them clearly distinguishable.

Discussion and reviews can be found here:

https://github.com/scalacenter/docs.scala-lang/pull/3


b-studios:
I tried to fix all broken links. However, there are a few remaining, that need resolution from third parties. In particular, the ones to the reference docs are still broken (which is a [known issue](https://github.com/lampepfl/dotty/pull/10849)). 

Also, there are links to Twitter which apparently give some `4xx`.

julienrf:
I’m not sure about the failures related to Twitter. Has something similar happened in the past?

b-studios:
The CI checker seems to have parameters to ignore specific external links. We should enable this for twitter, if the problem remains.

julienrf:
I’m going to ignore the CI failures related to twitter links and scala3doc links for now so that we can get the up to date documentation on docs.scala-lang.org now.

PR #1864, opened 2020-12-22
Use Em Dash Syntax in Scala 3 doc

adpi2:
Em Dash symbol is coded `---` in markdown and `&mdash` in html.

### Result in markdown
image

### Result in html
image

### DIFF
--- a/_overviews/scala3-book/ca-given-imports.md
+++ b/_overviews/scala3-book/ca-given-imports.md
@@ -34,7 +34,7 @@ object B:
 
 ## Discussion
 
-The wildcard selector `_` brings all definitions other than givens or extensions into scope, whereas a `given` selector brings all *givens* — including those resulting from extensions — into scope.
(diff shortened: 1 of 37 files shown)


b-studios (approved):

julienrf (approved):
Thank you Adrien!

We could improve it even further by removing the spaces before and after the em dashes, would you like to do it?

https://www.scribbr.com/language-rules/dashes/#the-em-dash-marking-a-break-in-a-sentence

b-studios:
Regarding removing spaces: I personally don't like the style of removed whitespace around dashes. But maybe it's just a matter of taste.

adpi2:
I think you are right @julienrf . The correct typography for `em-dash` is without spaces.

julienrf (approved):
Great, thank you!

The changes look good but I don’t think this PR should contain this commit: https://github.com/scala/docs.scala-lang/pull/1864/commits/613a394e396921b459da0edbd92462cc5e9f484d

adpi2:
Oups...

It should be better now :)

PR #1866, opened 2020-12-23
Updated per a comment, then cleaned up other areas

alvinj:
Updated the `List` hierarchy per a comment, then reviewed and cleaned up other areas.

### DIFF
--- a/_overviews/scala3-book/why-scala-3.md
+++ b/_overviews/scala3-book/why-scala-3.md
@@ -18,8 +18,8 @@ It’s hard to list every benefit of Scala, but a “Top Ten” list might look 
 
 1. Scala embraces a fusion of functional programming (FP) and object-oriented programming (OOP)
 2. Scala is statically typed, but often feels like a dynamically typed language
-3. Scala’s syntax is concise, but still readable; it’s often referred to as *expressive*
-4. *Implicits* in Scala 2 were a defining feature, and they have been improved and simplified in Scala 3
+3. Scala’s syntax is concise, but still readable; it’s often referred to as _expressive_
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/why-scala-3.md
```diff
@@ -33,38 +33,31 @@ It might be hard to list every benefit of Scala, but a Top Ten list looks like t
...
-    LinearSeq[A],
-    LinearSeqOps[A, List, List[A]],
-    StrictOptimizedLinearSeqOps[A, List, List[A]],
-    StrictOptimizedSeqOps[A, List, List[A]],
-    IterableFactoryDefaults[A, List],
-    DefaultSerializable
+  extends AbstractSeq[A] with LinearSeq[A] with LinearSeqOps[A, [A] =>> List[A],
+  List[A]] with StrictOptimizedLinearSeqOps[A, [A] =>> List[A], List[A]] with StrictOptimizedSeqOps[A, [A] =>> List[A],
+  List[A]] with IterableFactoryDefaults[A, [A] =>> List[A]]
+  with DefaultSerializable
```
julienrf, line 55: I might be missing some context but what is the motivation for showing this complex class definition on the page “why Scala 3”?

I personally find it a bit scary, especially because it refers to some traits that should be implementation details only (we didn’t have “transparent traits” at that time). Also, I don’t think `[A] =>> List[A]` is clearer than `List`.

alvinj: I’ll see what I can do about that. The intent was to show something of a real-world example, but that may be a little too intense.

b-studios (approved):

b-studios:
@julienrf is it ok, if I go ahead and merge this PR?

PR #1867, opened 2020-12-23
A brief ‘TASTy for End Users’ page, per Anatolii’s request

alvinj:
I didn’t add any Front Matter to this page yet because that will require renumbering pages in the book, depending on where this page is added.


_overviews/scala3-book/tasty-for-end-users.md
```diff
@@ -0,0 +1,150 @@
...
+As you can imagine, having a complete representation of your code in a platform-neutral form has many benefits:
+
+{% comment %}
+TODO: These lines mostly come from https://www.scala-lang.org/blog/2018/04/30/in-a-nutshell.html, and I don’t know exactly what the first bullet point means:
+{% endcomment %}
+
+- The compiler uses it to support separate compilation.
+- The Scala _Language Server Protocol_-based language server uses it to support hyperlinking, command completion, documentation, and also for global operations such as find-references and renaming.
+- Tasty makes an excellent foundation for a new generation of reflection-based macros.
+- A build tool can use it to cross-build on different platforms and migrate code from one binary version to another.
```
sjrd, line 119: This is a common misconception: you *can't* reuse the same tasty and recompile it for a different platform. You'll still have on tasty for the JVM, one tasty for JavaScript, and one tasty for Native, if you cross-compile for all three.

See https://www.scala-lang.org/2020/11/03/scalajs-for-scala-3.html#aside-what-about-tasty for details.

(The migrate code between binary versions is a valid use case, although that won't be relevant for a while.)

_overviews/scala3-book/tasty-for-end-users.md
```diff
@@ -0,0 +1,150 @@
...
+- Optimizers and analyzers can use it for deep code analysis and advanced code generation.
+
+There’s one more big benefit:
+
+- Because the Scala 2.13.4 compiler can read Tasty files, and the Scala 3 compiler can read Scala 2.13 class files, you can intermix Scala 3 and Scala 2.13 code in the same project.
```
sjrd, line 124: This is not really an advantage of Tasty. Is Scala 3 had used pickles like Scala 2, even with a different format, we would still have been able to do this. The fact that we use tasty for this is simply because that's where the information is, not because tasty is any better at storing that information than what we had before.

_overviews/scala3-book/tasty-for-end-users.md
alvinj: Thanks, I didn’t know that. I’ll update this.

_overviews/scala3-book/tasty-for-end-users.md
alvinj: Okay, thanks, I’ll delete this.

sjrd (approved):

_overviews/scala3-book/tasty-for-end-users.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Tasty benefits
+
+As you can imagine, having a complete representation of your code in a platform-neutral form has many benefits:
+
+{% comment %}
+TODO: These lines mostly come from https://www.scala-lang.org/blog/2018/04/30/in-a-nutshell.html, and I don’t know exactly what the first bullet point means:
+{% endcomment %}
+
+- The compiler uses it to support separate compilation.
```
b-studios, line 116: Regarding your comment: AFAIK the idea is the following: storing typed abstract syntax trees as intermediate results allows you to reuse them, instead of having to read and typecheck scala files from scratch. This is particularly important for features like inline and macros, that require access to the original typed tree, which is  not present in class files.

b-studios (commented):
I am also not sure where to put this document. Maybe we can add it to the guides section of the website?

https://docs.scala-lang.org/scala3/guides.html

Since it is a single page, it could be move to `scala3/guides/tasty-overview.md` or similar -- and then linked to in `scala3/guides.md`.

alvinj:
> I am also not sure where to put this document.

I don’t know if Anatolii had a place where he wanted it? If the book had an Appendix I could see it there. I also started to write a doc on how Metals and Bloop work, but I didn’t know where that would go, so I stopped working on it.

SethTisue:
> Maybe we can add it to the guides section of the website

SGTM

b-studios:
I moved the tasty docs to a guide in PR #1889 which supersedes this one.

PR #1868, opened 2020-12-23
Fix performance discussion on opaque types

b-studios:
Abstract types are erased to their bound and thus this discussion of _performance_ overhead for them is wrong. Thanks @sjrd for pointing this out!

### DIFF
--- a/_overviews/scala3-book/types-opaque-types.md
+++ b/_overviews/scala3-book/types-opaque-types.md
@@ -38,7 +38,7 @@ println((l2 + l3).toDouble) // prints 4.999...
 While the class `Logarithm` offers a nice abstraction for `Double` values that are stored in this particular logarithmic form, it imposes severe performance overhead: For every single mathematical operation, we need to extract the underlying value and then wrap it again in a new instance of `Logarithm`.
 
 
-## Preventing the Overhead
+## Module Abstractions
 Let us consider another approach to implement the same library.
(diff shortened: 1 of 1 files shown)


b-studios:
@julienrf could you look at this PR? I am sadly no maintainer and can't mark you as reviewer ;)

_overviews/scala3-book/types-opaque-types.md
```diff
@@ -98,8 +97,12 @@ Programming against the abstract module `Logarithms` can be very tedious and oft
 def someComputation(L: Logarithms)(init: L.Logarithm): L.Logarithm = ...
 ```
 
+#### Boxing Overhead
+Type abstractions, such as `type Logarithm` erase to their bound (which is `Any` in our case).
```
julienrf, line 101: Could we add a link around “erase” to some page somewhere that describes what type erasure is?

julienrf (approved):
Thank you Jonathan! This looks good to me, I just have a small request.

_overviews/scala3-book/types-opaque-types.md
b-studios, line 101:
```suggestion
Type abstractions, such as `type Logarithm` [erase](https://www.scala-lang.org/files/archive/spec/2.13/03-types.html#type-erasure) to their bound (which is `Any` in our case).
```

PR #1870, opened 2020-12-24
Small edits while reviewing a comment by Vinz

alvinj:
I made edits to hopefully improve this file as I was reviewing [this unapply comment by Vinz](https://github.com/scalacenter/docs.scala-lang/pull/3#discussion_r543402183).

### DIFF
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -8,7 +8,7 @@ next-page: domain-modeling-oop
 ---
 
 
-Scala 3 provides us with many different tools to model the world around us:
+Scala 3 provides many different constructs so we can model the world around us:
 
 - Classes
 - Objects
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -436,8 +449,8 @@ class Dog(name: String, var age: Int) extends Pet(name):
 
 val d = Dog("Fido", 1)
 ```
-Traits are more flexible to compose (you can mix in multiple traits, but only extend one class) and should most of the time be preferred to classes.
-The rule of thumb is to use classes whenever you want to create instances of a particular type and traits when you want to decompose and reuse behaviour.
+Traits are more flexible to compose --- you can mix in multiple traits, but only extend one class --- and should be preferred to classes and abstract classes most of the time.
```
b-studios, line 452: Somewhere else, we omit spaces around the dashes.
```suggestion
Traits are more flexible to compose---you can mix in multiple traits, but only extend one class---and should be preferred to classes and abstract classes most of the time.
```

b-studios (approved):

PR #1874, opened 2020-12-24
Resolving TODO tags, Part 2

alvinj:
This is a second batch of trying to resolve my TODO tags in the documents.


_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -8,15 +8,11 @@ next-page: fp-functions-are-values
...
-{% comment %}
-TODO: Use someone else’s definition?
-{% endcomment %}
-
 Another feature that Scala offers to help you write functional code is the ability to write pure functions.
 A _pure function_ can be defined like this:
 
 - A function `f` is pure if, given the same input `x`, it always returns the same output `f(x)`
-- The function’s output depends *only* on its input variables and its internal algorithm
+- The function’s output depends _only_ on its input variables and its internal algorithm
```
b-studios, line 15:
```suggestion
- The function’s output depends _only_ on its input variables and its implementation
```

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -8,15 +8,11 @@ next-page: fp-functions-are-values
...
 - It doesn’t modify its input parameters
 - It doesn’t mutate any hidden state
 - It doesn’t have any “back doors”: It doesn’t read data from the outside world (including the console, web services, databases, files, etc.), or write data to the outside world
```
b-studios, line 18:
```suggestion
- it only computes the output and does not modify the world around it.

This implies:
- It doesn’t modify its input parameters
- It doesn’t mutate any hidden state
- It doesn’t have any “back doors”: It doesn’t read data from the outside world (including the console, web services, databases, files, etc.), or write data to the outside world
```

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -42,23 +38,23 @@ These `String` methods are also pure functions:
...
 > If you’re interested in how methods can be used like functions, see the [Eta Expansion][eta] discussion.
 
 
 
 ## Examples of impure functions
 
-Conversely, the following functions are *impure* because they violate the definition.
+Conversely, the following functions are _impure_ because they violate the definition.
 
 The `foreach` method on collections classes is impure because it’s only used for its side effects, such as printing to STDOUT.
```
b-studios, line 50: In fact, this is not true. It will probably be called with impure functions but `List(1,2,3).foreach { () }` is perfectly pure.

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -42,23 +38,23 @@ These `String` methods are also pure functions:
...
 
 > A great hint that `foreach` is impure is that it’s method signature declares that it returns the type `Unit`.
 > Because it doesn’t return anything, logically the only reason you ever call it is to achieve some side effect.
-> Similarly, *any* method that returns `Unit` is going to be an impure function.
```
b-studios, line 58: Like foreach above, not completely true. `Unit` only _indicates_ that it might be executed for its side-effects. You can have (useless) pure methods that return `Unit`.

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -111,7 +107,7 @@ If you understand that code, you’ll see that it meets the pure function defini
 
 The first key point of this section is the definition of a pure function:
 
-> A *pure function* is a function that depends only on its declared inputs and its internal algorithm to produce its output.
+> A _pure function_ is a function that depends only on its declared inputs and its internal algorithm to produce its output.
 > It does not read any other values from “the outside world”---the world outside of the function’s scope---and it doesn’t modify any values in the outside world.
```
b-studios, line 111: Same as above
```suggestion
> A _pure function_ is a function that depends only on its declared inputs and its implementation to produce its output.
> It only computes its output and does not depend on or modify the outside world.
```

b-studios (commented):
@alvinj maybe you can look again at the definition of a pure function. Otherwise this PR mostly affects formatting and comments and looks good to me.

_overviews/scala3-book/fp-pure-functions.md
alvinj: Is there any way to reword this, or is it best to just delete the `foreach` sentence and the three bullet points after it?

alvinj: If you want to keep this discussion, here’s a stab at rewording it. FWIW, the main reason I like this discussion is that it helps people keep an eye out for methods that return `Unit`. I know it was something I never considered many years ago. Here’s the rewording:

While the `foreach` method on collections classes is technically pure, it’s typically used for side effects, such as printing to STDOUT.

> A great hint that `foreach` might be used for side effects is that it returns the type `Unit`.
> Because it doesn’t return anything, logically the only reason you ever call it is to achieve some side effect.
> Similarly, _any_ method that returns `Unit` is a candidate to be an impure function.

b-studios: Sorry for not following up earlier. I think your argument is perfectly sensible. Maybe it is easiest to actually say exactly that: "Keep an eye out for methods that return `Unit`. Since they do not return anything useful, this return type indicates that they are only called for their side effects"?

_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -398,7 +397,12 @@ For more details, see the remainder of these modeling lessons.
 ## Abstract classes
 
 {% comment %}
-TODO: I have some notes on when to use abstract classes, and can update this section.
+LATER: If anyone wants to update this section, our comments about abstract classes and traits are on Slack. The biggest points seem to be:
```
b-studios, line 400:
```suggestion
LATER: If anyone wants to update this section, here are a few points that could be mentioned:
```

b-studios (approved):

PR #1875, opened 2020-12-25
Resolving TODO tags, Part 3

alvinj:
This is the third batch of changes related to closing out my TODO tags in the files.


_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -25,13 +25,13 @@ It presents the similarities and differences between Java and Scala at a high le
 
 At a high level, Scala shares these similarities with Java:
 
-- Scala code is compiled to *.class* files, packaged in JAR files, and runs on the JVM
-- It’s an object-oriented programming (OOP) language
+- Scala code is compiled to _.class_ files, packaged in JAR files, and runs on the JVM
+- It’s an [object-oriented programming][modeling-oop] (OOP) language
 - It’s statically typed
-- Both languages have support for immutable collections, lambdas, and higher-order functions
+- Both languages have support for immutable collections, lambdas, and [higher-order functions][hofs]
```
b-studios, line 31: Sorry, this might be a stupid question, but which immutable collections in Java are you referring to? As far as I remember, most collections in the stdlib in Java are mutable. There exists Guava, but that's not the stdlib. Maybe things improved and I just did not notice it :)

_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -68,22 +68,22 @@ TODO: add a point about how the type system lets you express details as desired
...
 - Scala has state of the art, third-party, open source functional programming libraries
 - Scala case classes are like records in Java 14; they help you model data when writing FP code, with built-in support for concepts like pattern matching and cloning
-- Thanks to features like by-name parameters, infix notation, optional parentheses, extension methods, and higher-order functions, you can create your own “control structures” and DSLs
+- Thanks to features like by-name parameters, infix notation, optional parentheses, extension methods, and [higher-order functions][hofs], you can create your own “control structures” and DSLs
 - Scala files do not have to be named according to the classes or traits they contain
-- Many other goodies: companion classes and objects, macros, union and intersection types, toplevel definitions, numeric literals, multiple parameter lists, default values for parameters, named arguments, and more
+- Many other goodies: companion classes and objects, macros, [union][union-types] and [intersection][intersection-types], [toplevel definitions][toplevel], numeric literals, multiple parameter lists, default values for parameters, named arguments, and more
```
b-studios, line 86: toplevel definitions are already in the list above.

_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -964,8 +964,8 @@ Examples of how to create instances of immutable collections.
 
 ### Mutable collections classes
 
-Scala has mutable collections classes like `ArrayBuffer`, `Map`, and `Set`, in its *scala.collection.mutable* package.
-After importing them into the current scope, they’re created just like the immutable `List`, `Vector`, `Map`, and `Set` examples just shown.
+Scala has mutable collections classes like `ArrayBuffer`, `Map`, and `Set` in its _scala.collection.mutable_ package.
```
b-studios, line 967: I remember a review mentioned `Array`. I think it is very natural for Java programmers to use `Array` in Scala and should be mentioned here.

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -38,25 +38,25 @@ At a high level, Scala shares these *similarities* with Python:
 
 - Both are high-level programming languages, where you don’t have to concern yourself with low-level concepts like pointers and manual memory management
 - Both have a relatively simple, concise syntax  
-- Both are functional programming (FP) languages
+- Both are [functional programming][fp-intro] (FP) languages
```
b-studios, line 41: Maybe say: "support a functional style of programming" like in most other places. Otherwise this triggers controversial discussions.

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -38,25 +38,25 @@ At a high level, Scala shares these *similarities* with Python:
...
+Also at a high level, the _differences_ between Python and Scala are:
   
 - Python is dynamically typed, and Scala is statically typed  
   - Though it’s statically typed, Scala features like type inference make it feel like a dynamic language
-- Python is interpreted, and Scala code is compiled to *.class* files, and runs on the Java Virtual Machine (JVM)
+- Python is interpreted, and Scala code is compiled to _.class_ files, and runs on the Java Virtual Machine (JVM)
 - In addition to running on the JVM, the [Scala.js](https://www.scala-js.org) project lets you use Scala as a JavaScript replacement
 - The [Scala Native](https://scala-native.readthedocs.io/en/v0.3.9-docs) project lets you write “systems” level code, and compiles to native executables
-- Everything in Scala is an *expression*: constructs like `if` statements, `for` loops, `match` expressions, and even `try`/`catch` expressions all have return values
+- Everything in Scala is an _expression_: constructs like `if` statements, `for` loops, `match` expressions, and even `try`/`catch` expressions all have return values
```
b-studios, line 57: Is this really different in python?

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -68,17 +68,16 @@ This section looks at the similarities you’ll see between Python and Scala whe
...
-- With Scala 3’s toplevel definitions you can put method, field, and other definitions anywhere
-  - One difference is that Python can operate without even declaring a single method, while Scala 3 can’t do _everything_ at the toplevel; for instance, a `@main def` method is required to start a Scala application
+- With Scala 3’s [toplevel definitions][toplevel] you can put method, field, and other definitions anywhere
+  - One difference is that Python can operate without even declaring a single method, while Scala 3 can’t do _everything_ at the toplevel; for instance, a [main method][main-method] (`@main def`) is required to start a Scala application
 
 ### Programming level differences
 
 Also at a programming level, these are some of the differences you’ll see every day when writing code:  
 
 - Scala’s syntax is extremely consistent:
```
b-studios, line 78: Here you say "syntax", but then the most points that follow talk about ecosystem (library, style of coding, etc.)

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -89,22 +88,23 @@ Also at a programming level, these are some of the differences you’ll see ever
...
 ### Features compared with examples
 
 Given that introduction, the following sections provide side-by-side comparisons of Python and Scala programming language features.
 
-> The general Python standard is to indent code with four spaces, but in the following examples only two spaces are used.
-> This is only done so the examples can be shown side by side.
+{% comment %}
+TODO: Update the Python examples to use four spaces. I started to do this, but then thought it would be better to do that in a separate PR.
+{% endcomment %}
+
```
b-studios, line 107: Maybe we can record TODOs in issues, rather than inline comments. That would make it easier to track them and would not show up in diffs.

_overviews/scala3-book/scala-for-java-devs.md
alvinj: When I looked into this I started with the `List`, `Map`, and `Set` links on [this page](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collections.html), and found that these methods create collections that are “unmodifiable”: `List.of`, `List.copyOf`, `Map.of`, `Map.ofEntries`, `Map.copyOf`, `Set.of`, and `Set.copyOf`. They then discuss what they mean by _unmodifiable_ in the Unmodifiable Collections section on [this page](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#unmodifiable).

b-studios: Ok, great - I happily retract :) given that these exists already for a long time I cannot recall what my problem with immutable collections was back in the time. Thanks for clarifying.

alvinj: Okay, I modified that line to only mention lambdas and HOFs, and added a line under “Differences” about the immutable collections.

_overviews/scala3-book/scala-for-java-devs.md
alvinj: Yes, there was another discussion somewhere about `Array`. I don’t think I got an answer to my question about where `Array` should be used, but I agree, it’s definitely natural for Java programmers to want to use it. I’ll update this text.

_overviews/scala3-book/scala-for-python-devs.md
alvinj: Sounds good. I’ll work on removing them all this week.

_overviews/scala3-book/scala-for-java-devs.md
SethTisue: Mentioning `Array` seems valid, but I would suggest that it not simply be included without additional comment. Perhaps something like: "Normal Scala code doesn't use `Array`. It is useful for interoperating with other languages (Java, JavaScript), and it may also be useful when writing low-level code that needs to squeeze maximum performance out of the underlying platform." That kind of thing.

I see a lot of beginners in our chat rooms and forums who are using `Array` not because it's actually an appropriate choice for what they're doing, but simply because they don't know any better.

alvinj: I added a discussion about `Array` and also added @SethTisue’s comments as a note after that, so I’m resolving this conversation.

_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -962,8 +963,34 @@ Examples of how to create instances of immutable collections.
...
+```scala
+val nums = Array(1, 2, 3, 4, 5)
+val doubledNums = nums.map(_ * 2)
+val filteredNums = nums.filter(_ > 2)
+```
+
+Because the Scala `Array` is represented in the same way as the Java `array`, you can easily use Java methods that return arrays in your Scala code.
+
+> Despite that discussion of `Array`, bear in mind that normal Scala code doesn’t use `Array`.
+> It’s primarily useful for interoperating with other languages (Java, JavaScript), and it may also be useful when writing low-level code that needs to squeeze maximum performance out of the underlying platform. But in general, when you need to use a sequence, the Scala idiom is to prefer immutable sequences like `Vector` and `List`, and then use `ArrayBuffer` if and when when you really need a mutable sequence.
```
b-studios, line 993:
```suggestion
```

I would simply drop this comment since I find it judging on what "normal Scala code" is.

PR #1880, opened 2021-01-03
Queue support constant time prepend

exoego:
Ref.

* https://github.com/scala/docs.scala-lang/pull/650 fixed docs for Scala 2.12
* https://github.com/scala/scala/blob/1851c5d637e55c8af786f428bddc9758a44c583c/src/library/scala/collection/immutable/Queue.scala#L113


SethTisue:
thank you!

PR #1884, opened 2021-01-04
Fix names of View's sub classes

exoego:
There are no `View.Append` or `View.Prepend` classes.
Instead, `View.Appended` and `View.Prepend` exist.

Refer

* https://github.com/scala/scala/blob/1851c5d637e55c8af786f428bddc9758a44c583c/src/library/scala/collection/View.scala#L366
* https://github.com/scala/scala/blob/1851c5d637e55c8af786f428bddc9758a44c583c/src/library/scala/collection/View.scala#L377


julienrf (approved):
Good catch, thank you @exoego!

PR #1885, opened 2021-01-04
Fix continuous integration by ignoring some more urls

b-studios:
Continuous integration is broken because of two reasons:

1) Twitter API responds with `400` which indicates missing permission
2) Some API links generated by the Scala 3 doctool are currently not available

This PR addresses both by ignoring `400` status errors and ignoring the missing doctool url (temporarily).


b-studios:
Strange... ignoring did the trick for me locally. I'll try `ignore-file` on the conflicting file instead.

b-studios:
Ok, I changed the ignore pattern and we are now good to go.

PR #1897, opened 2021-01-14
Add Scala 3 macro tutorial quote type matching section

nicolasstucki:


_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
 
 ### Matching types
 
-*Coming soon*
+So far we assumed that we knew statically the type within the quote patterns.
```
b-studios, line 252:
```suggestion
So far, we assumed that the types within quote patterns would be statically known.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+Quote patterns also allow for generic types and existential types which we will see in this section.
```
b-studios, line 253:
```suggestion
Quote patterns also allow for generic types and existential types, which we will see in this section.
```

b-studios: Maybe say: "In general, generic types and existential types are not statically known." to connect the two sentences.

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+
+#### Generic types in patterns
+
+Consider the `exprOfOption` that we have already seen
```
b-studios, line 257:
```suggestion
Consider the function `exprOfOption` that we have already seen:
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+```scala
+def exprOfOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Expr[T]] =
+  x match
+    case '{ Some($x: T) } => Some(x) // x: Expr[T]
```
b-studios, line 261:
```suggestion
    case '{ Some($x: T) } => Some(x) // x: Expr[T]
                // ^^^ type ascription with generic type T
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+    ...
+```
+
+Note that this time we have added the `T` explicitly in the pattern even though it can be inferred.
```
b-studios, line 265:
```suggestion
Note that this time we have added the `T` explicitly in the pattern, even though it could be inferred.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+The pattern is using a generic type `T` which requires a given `Type[T]` in scope.
```
b-studios, line 266:
```suggestion
By referring to the generic type `T` in the pattern, we are required to have a given `Type[T]` in scope.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+This implies that `$x: T` will only match if `x` is of type `Expr[T]`.
+In this particular case this condition will always be true.
+
+Now consider the following variant where we do not know statically the type within the option in `x`.
```
b-studios, line 270:
```suggestion
Now consider the following variant where `x` is an optional value with a (statically) unknown element type.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+    case _ => None
+```
+This time the pattern ` Some($x: T)` will only match if the type of the option is `Some[T]`.
+
+```scala
+exprOfOptionOf[Int]('{ Some(3) })   // Some('{3})
+exprOfOptionOf[Int]('{ Some("a") }) // None
+```
+
+#### Type variables in patterns
```
b-studios, line 285:
```suggestion
#### Type variables in quoted patterns
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+
+Quoted code may contain types that are not known outside of the quote.
+We can match on them using pattern type variables.
+Just as in a normal pattern, the type variables are written using lower case names.
+
+```scala
+def exprOptionToList(x: Expr[Option[Any]])(using Quotes): Option[Expr[List[Any]]] =
+  x match
+    case '{ Some($x: t) } =>
```
b-studios, line 294:
```suggestion
    case '{ Some($x: t) } =>
                // ^^^ this binds the type `t` in the body of the case
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+    case '{ None } =>
+      Some('{ Nil })
+    case _ => None
+```
+
+The pattern `$x: t` will match an expression of any type and `t` will be bound to the type of the pattern.
+This type is only valid in the right-hand side of the `case`, in the example we can use it to construct the list `List[t]($x)` (`List($x)` would also work).
+As this is a type that is not statically known we need a given `Type[t]` in scope, luckily the quoted pattern will automatically provide this.
+
+The simple `case '{ $expr: tpe } =>` pattern is extramlely useful if we want to konw the precise type of the expression.
```
b-studios, line 305:
```suggestion
The simple `case '{ $expr: tpe } =>` pattern is very useful if we want to know the precise type of the expression.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+```scala
+val expr: Expr[Option[Int]] = ...
+expr match
+  case '{ $expr: tpe } =>
+    Type.show[tpe] // could be: Option[Int], Some[Int], None, Option[1], Option[2], ...
+    '{ val x: tpe = $expr; x } // binds the value without widening the type
+    ...
+```
+
+In some cases we need to define a pattern variable that is referenced several times or has some bounds.
```
b-studios, line 315:
```suggestion
In some cases we need to define a pattern variable that is referenced several times or has some type bounds.
```

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+To achieve this it is possible to create pattern variables at the start of the pattern using `type t` with a type pattern variable.
+
+```scala
+def fuseMap[T: Type](x: Expr[List[T]])(using Quotes): Expr[List[T]] = x match {
+  case '{ type u; type v; ($ls: List[`u`]).map($f: `u` => `v`).map($g: `v` => T) } =>
```
b-studios, line 320:
```suggestion
  case '{ 
    type u
    type v
    ($ls: List[`u`])
      .map($f: `u` => `v`)
      .map($g: `v` => T) 
    } =>
```

b-studios: I am aware that this is consistent with pattern matching, but still find it confusing that.we have a different escaping mechanism for types and terms...

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -249,7 +249,106 @@ case ...
...
+    '{ $ls.map(x => $g($f(x))) }
+  case _ => x
+}
+```
+
+Here we define two type variables `u` and `v` and then refer to them using `` `u` `` and `` `v` ``.
+We do not refer to them using `u` or `v` because those would be interpreted as new type variables and hence duplicates.
```
b-studios, line 327: Maybe explain the analogy to normal type variables in patterns so that users can use this as a reference?

nicolasstucki: Do you know where I could link to the documentation of this feature?

b-studios: Currently I think only the Scala 2 specification - since we don't have a full spec, yet. So maybe this: https://www.scala-lang.org/files/archive/spec/2.13/08-pattern-matching.html#stable-identifier-patterns

b-studios (approved):

PR #1898, opened 2021-01-15
Fix typos and minor issues in inline.md

aumann:
I hope this does not fall into the "large change" category. Please let me know if you prefer strict separation of typo vs semantic fixes.

I will try and add a small review comment to the "semantic" fixes.

### DIFF
--- a/_overviews/scala3-macros/tutorial/inline.md
+++ b/_overviews/scala3-macros/tutorial/inline.md
@@ -46,7 +46,7 @@ Then an optimization called "constant folding" is applied by the compiler, which
 > However, this form of constant propagation is _best-effort_ and not guaranteed.
 > Scala 3.0 also supports `final val`-inlining as _best-effort_ inlining for migration purposes.
 
-Currently, only constant expression may appear on the right-hand side of an inline value definition.
+Currently, only constant expressions may appear on the right-hand side of an inline value definition.
 Therefore, the following code is invalid, though the compiler knows that the right-hand side is a compile-time constant value:
 
 ```Scala
@@ -75,7 +75,7 @@ Therefore, the compiler inlines the following call
 
 ```scala
 logged(logLevel, getMessage()) {
-  computeSomthing()
+  computeSomething()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-macros/tutorial/inline.md
```diff
@@ -86,8 +86,8 @@ val level   = logLevel
 def message = getMessage()
 
 println(s"[$level]Computing $message")
-val res = computeSomthing()
-println(s"[$level]Result of $tag: $res")
+val res = computeSomething()
+println(s"[$level]Result of $message: $res")
```
aumann, line 90: 'tag' does not appear anywhere else and if I understand correctly the by-name parameter should be evaluated twice (i.e. the call to the 'message' def)

_overviews/scala3-macros/tutorial/inline.md
```diff
@@ -213,7 +213,7 @@ assert1(x, "error1")
 val cond = x
 def msg = "error1"
 if !cond then
-    throw new Exception("error1")
+    throw new Exception(msg)
```
aumann, line 216: similar: msg is by-name, and thus called here

_overviews/scala3-macros/tutorial/inline.md
```diff
@@ -379,17 +379,17 @@ It is also possible to create _abstract inline definitions_.
...
 
 This forces the implementation of `log` to be an inline method and also allows `inline` parameters.
 Counterintuitively, the `log` on the interface `InlineLogger` cannot be directly called. The method implementation is not statically known and we thus do not know what to inline.
-Calling an abstract inline methods thus results in an error.
+Calling an abstract inline method thus results in an error.
 The usefuleness of abstract inline methods becomes apparent when used in another inline method:
 
 ```scala
-inline def logged(logger: Logger, x: Any) =
+inline def logged(logger: InlineLogger, x: Any) =
```
aumann, line 392: I'm not sure here but 'Logger' does not appear in this example, and this way we actually call the method on 'InlineLogger', that is we call the _abstract method_

b-studios (approved):

PR #1911, opened 2021-01-27
Move ci script to Github Actions

vincenzobaz:
This PR introduces a Github Action workflow for the repositry


vincenzobaz:
The UI looks like [this](https://github.com/vincenzobaz/docs.scala-lang/runs/1776302291?check_suite_focus=true)

vincenzobaz:
I think it is appropriate to use the "Squash and merge" button if the review is positive

scripts/ci.sh
```diff
@@ -1,21 +0,0 @@
...
-. /usr/local/rvm/scripts/rvm
-
-set -eux
-
-bundle install
-./scripts/run-mdoc.sh
-rm -r /tmp/mdoc-out/
-bundle exec jekyll build
-
-# Checking for docs.scala-lang/blob/master leads to a chicken and egg problem because of the edit links of new pages.
```
sjrd, line 12: We should preserve this comment in the new `build.yml`.

sjrd (changes requested):
We should also remove `.drone.yml`, I believe.

sjrd (approved):

SethTisue (approved):

PR #1912, opened 2021-01-27
Fixed typo

poslegm:
I am not sure if this is a typo or an obviously incorrect example. In second case it would be great to add comment like
```scala
val res2: Option[String] = db.get(Age) // <-- would not compile
```


dongxuwang:
LGTM

Philippus (approved):

SethTisue:
thanks!

PR #1916, opened 2021-02-01
fix some 404 links by removing them

SethTisue:


SethTisue:
@nicolasstucki @b-studios I'm going to merge this to get CI green again, but perhaps you'd like to follow up with a better fix

PR #1918, opened 2021-02-02
Would prefer if the previously declared `t` was used

Magua:


SethTisue:
@alvinj lgty?

(and is @-mentioning you the best way to get review on Scala Book changes?)

alvinj:
I actually prefer the original approach. The intent isn’t too assign the tuple to `t`, and then assign `t` to another variable. That could also be misleading. If it would help, we could do something like this:

```
def getATuple = (11, "Eleven", new Person("Eleven"))
```

then:

```
val t = getATuple
```

then this:

```
val (num, string, person) = getATuple
```

I’m pressed for time atm, but I could make that change this weekend.

(And yes, I receive the notifications, so this is fine.)

Magua:
I believe the point of this line is to show deconstruction of a tuple with pattern matching. So I liked `t` to be used to add no extra fuss. But no biggie.

alvinj:
> I believe the point of this line is to show deconstruction of a tuple with pattern matching. So I liked `t` to be used to add no extra fuss. But no biggie.

Thanks for this. Part of my problem here was trying to look at this when I was in a rush, I shouldn’t have done that. I do like the general idea of what you proposed.

PR #1923, opened 2021-02-14
scala 3 installation via cs

xhudik:
original command have installed scala 2 (that was pretty misleading). Now `cs` install repl and compiler for Scala 3. Now it is consistent with http://dotty.epfl.ch/#getting-started, where the same 2 commands are used

### DIFF
--- a/scala3/getting-started.md
+++ b/scala3/getting-started.md
@@ -27,7 +27,8 @@ Install it on your system with the following instructions.
 <div class="main-download">
   <div id="download-step-one">
     <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a> then run:</p>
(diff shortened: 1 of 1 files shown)


sjrd:
Scala 3 doesn't have a stable release yet. Why would we recommend that as the *default* way to install Scala?

Philippus:
> Scala 3 doesn't have a stable release yet. Why would we recommend that as the _default_ way to install Scala?

I think this PR is changing the docs at https://docs.scala-lang.org/scala3/getting-started.html.

sjrd:
Ah, good point.

xhudik:
any idea why check is failing?  (`External link https://www.scala-lang.org/api/2.10.1/ failed: 404 No error`)

Philippus:
Not sure why that url is gone, but it's not related to your PR I think.

Philippus (approved):

SethTisue:
thank you!

PR #1924, opened 2021-02-14
Updated Hello World example

furkanaktas:
Hello World example is updated without using App trait


furkanaktas:
PR is closed because of open from master branch.

PR #1925, opened 2021-02-14
Update Hello World example

furkanaktas:
Hello World example is updated without using App trait

### DIFF
--- a/_overviews/scala-book/prelude-taste-of-scala.md
+++ b/_overviews/scala-book/prelude-taste-of-scala.md
@@ -38,8 +38,10 @@ Before we jump into the examples, here are a few important things to know about 
 Ever since the book, *C Programming Language*, it’s been a tradition to begin programming books with a “Hello, world” example, and not to disappoint, this is one way to write that example in Scala:
 
 ```scala
-object Hello extends App {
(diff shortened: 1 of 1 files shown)


b-studios:
What is your motivation behind this change? IIUC in Scala 2 the preferred way of writing a simple app is using the `App` trait. At least for beginners and quick sketches this is a good way of doing it.

furkanaktas:
> What is your motivation behind this change? IIUC in Scala 2 the preferred way of writing a simple app is using the `App` trait. At least for beginners and quick sketches this is a good way of doing it.

I totally agree with you however document says in following steps
`If you’re coming to Scala from Java, scalac is just like javac, and that command creates two files:
- Hello$.class
- Hello.class
`
 but `Hello$delayedInit$body.class` also created with `App` trait approach and there is no info about this file. Document totally matches when using classical approach.

b-studios:
I see you point. For consistency we need to do one of two things:

1) mention that there are also a few other helper files generated (without explaining them)
2) use the traditional approach

I really don't know what's the best way to approach this from an educational perspective. Luckily with Scala 3 this decision goes away...

PR #1926, opened 2021-02-19
Replace old FunSuite with newer AnyFunSuite

AlinGabriel:
The change is reflecting the newer tutorial example: When running: `sbt new scala/scalatest-example.g8`
the build.sbt file inside  contains the newer "scalatest" % "3.2.2" but the example above was written using the old "scalatest" % "3.0.8" in which the FunSuite was replaced by the newer AnyFunSuite which has a different path in the scalatest library.


martijnhoekstra:
Build failure unrelated.

Looks like a net positive to me, and mergeable as is and even better with an import.

SethTisue:
I don't have a strong an opinion on whether to `import` or not, so, merging! thanks Alin & Martijn

PR #1928, opened 2021-02-22
No with keyword

aesteve:
The example solely uses the `extends` keyword, not `with`


Philippus:
Duplicate of https://github.com/scala/docs.scala-lang/pull/1927, great minds think alike. :)

aesteve:
Sorry I should have double-checked before opening the PR.

PR #1930, opened 2021-02-23
remove d in Double

xhudik:
e.g. `4e-10` - is a scientific format of number where `e` is a power, but `4e-10d` looks like a typo to me. If not what `d` stands for?


b-studios:
I agree it is not necessary at that point. @alvinj maybe added it to distinguish `4e-10d` (double) from `4e-10f`) float.

alvinj:
Yes, removing this is fine. (FWIW, I tracked that `d` down to a blog post I wrote back in 2012, and it referenced another doc that was written in 1998.  :smile:)

PR #1932, opened 2021-02-24
Added to the tuple intro, related to #1918

alvinj:
Added details to the tuple introduction, related to PR #1918, including showing the resulting values in `t` and the pattern matching/extractor approach.

### DIFF
--- a/_overviews/scala-book/prelude-taste-of-scala.md
+++ b/_overviews/scala-book/prelude-taste-of-scala.md
@@ -521,7 +521,11 @@ price: scala.math.BigDecimal = 123.45
 volume: Long = 101202303
 ```
 
-Tuples are nice for those times when you want to put a little “bag” of things together for a little while.
+Tuples are nice for those times when you want to quickly (and temporarily) group some things together.
+If you notice that you are using the same tuples multiple times, it could be useful to declare a dedicated case class, such as:
+```scala
+case class StockInfo(symbol: String, price: BigDecimal, volume: Long)
+```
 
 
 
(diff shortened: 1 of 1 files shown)


Magua (approved):

_overviews/scala-book/prelude-taste-of-scala.md
```diff
@@ -462,33 +462,66 @@ There are many (many!) more methods available to Scala collections classes, and
...
+(11, 11.0, "Eleven")
 ```
 
-You can create a tuple that contains three different types like this:
+This is known as a `Tuple3`, because it contains three elements.
+
+Tuples are convenient in many places, such as where you might use an ad-hoc class in other languages. For instance, you can return a tuple from a method instead of returning a class:
+
+```scala
+def getAaplInfo() = {
```
julienrf, line 476: I suggest showing the result type for two reasons. First, this is good practice, second, it shows the type syntax for tuples.

```suggestion
def getAaplInfo(): (String, BigDecimal, Long) = {
```

alvinj: I debated about doing that, so yes, that’s great. Thanks!

_overviews/scala-book/prelude-taste-of-scala.md
```diff
@@ -462,33 +462,66 @@ There are many (many!) more methods available to Scala collections classes, and
...
+Once again, the REPL shows the result:
+
+```scala
+scala> val (symbol, price, volume) = getAaplInfo()
+symbol: String = AAPL
+price: scala.math.BigDecimal = 123.45
+volume: Long = 101202303
+```
+
+Tuples are nice for those times when you want to put a little “bag” of things together for a little while.
```
b-studios, line 524:
~~~suggestion
Tuples are nice for those times when you want to quickly (and temporarily) group some things together.
If you notice that you are using the same tuples multiple times, it could be useful to declare a dedicated case class, such as:
```scala
case class StockInfo(symbol: String, price: BigDecimal, volume: Long)
```
~~~

b-studios (approved):

PR #1936, opened 2021-02-26
traits->classes

xhudik:
the whole section `Planning for Extension` is about classes (not traits). 

It also might be that I just didnt get the sentence:
*However, since traits are designed as the primary means of decomposition, a class that is defined in one file cannot be extended in another file. In order to allow this, the base class needs to be marked as `open`*


b-studios:
Oh, I think the point is that you should use traits if you want to decompose a system, not classes.

xhudik:
@b-studios  - yeah - now it is clear to me. yep - i'm closing the issue - thanks for the clarification!

Issue #1937, opened 2021-02-26
better explanaton of `open` keyword


xhudik:
Since `open` keyword is a new concept of Scala 3. It would desire a bit more info in: https://docs.scala-lang.org/scala3/book/domain-modeling-oop.html (section: Planning for Extension). This is a first place where a reader finds this term.

A possible source of inspiration: 
https://medium.com/@knoldus/what-is-open-class-in-scala-3-14876e047fd4

PR #1938, opened 2021-02-26
a bit more context for `open` feature

xhudik:
https://github.com/scala/docs.scala-lang/issues/1937

### DIFF
--- a/_overviews/scala3-book/domain-modeling-oop.md
+++ b/_overviews/scala3-book/domain-modeling-oop.md
@@ -125,11 +125,11 @@ class SoftwareDeveloper(name: String, favoriteLang: String)
 ```
 However, since _traits_ are designed as the primary means of decomposition,
 a class that is defined in one file _cannot_ be extended in another file.
-In order to allow this, the base class needs to be [marked as `open`][open]:
+In order to allow this, the base class needs to be marked as `open`:
 ```scala
 open class Person(name: String)
 ```
-Having to explicitly mark classes as open avoids many common pitfalls in OO design.
+[`open`](open) is a new feature of Sala 3. Having to explicitly mark classes as open avoids many common pitfalls in OO design.
 In particular, it requires library designers to explicitly plan for extension and for instance document the classes that are marked as open with additional extension contracts.
 
 {% comment %}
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/domain-modeling-oop.md
```diff
@@ -125,11 +125,11 @@ class SoftwareDeveloper(name: String, favoriteLang: String)
 ```
 However, since _traits_ are designed as the primary means of decomposition,
 a class that is defined in one file _cannot_ be extended in another file.
-In order to allow this, the base class needs to be [marked as `open`][open]:
+In order to allow this, the base class needs to be marked as `open`:
 ```scala
 open class Person(name: String)
 ```
-Having to explicitly mark classes as open avoids many common pitfalls in OO design.
+[`open`](open) is a new feature of Sala 3. Having to explicitly mark classes as open avoids many common pitfalls in OO design.
```
b-studios, line 132:
```suggestion
Marking classes with [`open`](open) is a new feature of Sala 3. Having to explicitly mark classes as open avoids many common pitfalls in OO design.
```

martijnhoekstra:
The link to open doesn't seem to work, html-proofer chokes on it.

Where should it link?

xhudik:
@martijnhoekstra , this is out of scope of this pull request. 

But you are right, link seems not to be working:
original page: https://docs.scala-lang.org/scala3/book/domain-modeling-oop.html links to non-existing https://docs.scala-lang.org/scala3/book/open


At the bottom of the page you can see: 
```
[open]: {{ site.scala3ref }}/other-new-features/open-classes.html [trait-params]: {{ site.scala3ref }}/other-new-features/trait-parameters.html
```
So it should points to https://dotty.epfl.ch/docs/reference/other-new-features/open-classes.html

@b-studios , @alvinj  - can you pls check whether the link above is ok? (i dont know how linking here work). If not, how we can point (link) to https://dotty.epfl.ch/docs/reference/other-new-features/open-classes.html ?

b-studios:
The problem is indeed the wrong use of Markdown linking syntax. I didn't catch that one, but you replaced `[marked as open][open]`  with `[open](open)`. Please note the round parens versus square brackets. Links in square brackets can be defined later on with `[label]: link...`. Round braces enclose actual relative or absolute paths.

xhudik:
@b-studios aha - so this is a difference from a "normal" markdown where you have [](). Ok, i'll try to repopen the pull request and fix it.

PR #1941, opened 2021-02-28
enums added

xhudik:
In Scala 3, enums can have methods as well (and they are public by default).

### DIFF
--- a/_overviews/scala3-book/methods-most.md
+++ b/_overviews/scala3-book/methods-most.md
@@ -267,7 +267,7 @@ This method works just like the previous method that used an `if`/`else` express
 
 ## Controlling visibility in classes
 
-In classes, objects, and traits, Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
+In classes, objects, traits and enums Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
 
 ```scala
 class Dog:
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/methods-most.md
```diff
@@ -267,7 +267,7 @@ This method works just like the previous method that used an `if`/`else` express
 
 ## Controlling visibility in classes
 
-In classes, objects, and traits, Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
+In classes, objects, traits and enums Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
```
b-studios, line 270:
```suggestion
In classes, objects, traits, and enums Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
```

xhudik: those `,` I'm not sure - based on: https://www.grammarly.com/blog/comma-before-and/ you can put `, and` ; but I'm suspicious about `enums, ` at the end of the list. Some native speaker might help here - mr @alvinj sounds like native :)

b-studios: IIRC we use the Oxford comma in lists most over the book. About the last comma, I am actually also not sure :) I just added it back, because it was there in the first place.

alvinj: Yes, I’d keep the last comma. The classic example example of this is shown on [this zazzle.com page](https://www.zazzle.com/lets_eat_grandma_commas_save_lives_poster-228790639368409132). :)

lets-eat-grandma

PR #1942, opened 2021-03-04
scala 3 syntax for import

xhudik:


b-studios:
As mentioned in #1956 it would be great to catch them all in one PR.

b-studios:
I guess this one fixes imports additional to #1956

PR #1943, opened 2021-03-04
scala 3 doesnt use _ in imports

xhudik:
removed a part describing _ in imports, since Scala 3 uses * instead (https://dotty.epfl.ch/docs/reference/changed-features/imports.html)

### DIFF
--- a/_overviews/scala3-book/packaging-imports.md
+++ b/_overviews/scala3-book/packaging-imports.md
@@ -83,7 +83,7 @@ Those are discussed at the end of this chapter.
 
 A note before moving on:
 
-- Import clauses are not required for accessing members of the same package.
+> Import clauses are not required for accessing members of the same package.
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/packaging-imports.md
```diff
@@ -81,12 +81,9 @@ They’re explained more in the subsections that follow.
 Import statements are also used to import `given` instances into scope.
 Those are discussed at the end of this chapter.
 
-Two notes before moving on:
+A note before moving on:
 
 - Import clauses are not required for accessing members of the same package.
```
b-studios, line 86: Now that it is only a single note, maybe drop `A note before moving on:` and  replace by:

```suggestion
> Import clauses are not required for accessing members of the same package.
```
?

b-studios:
Not necessary anymore after #1956

PR #1945, opened 2021-03-06
Modify explanation of example

MikeG112:
I believe this example is meant to show that all objects in Scala are of type `Any`. 

I am new to Scala, and may have misinterpreted this, but in any case I think the existing suggestion that the example shows that numerous types "are all objects just like every other object" is difficult to interpret having read only the preceding introductory tour docs.


martijnhoekstra:
The failing build is not related to this PR, but is #1953 

This looks good to me. I think the original intent of the text was to echo the "everything is an object" maxim. I don't think that matters. Your wording sounds clearer than it was before in my eye. I don't know how things are able to sound in my eye, but it is what it is.

MikeG112:
Thanks for the review! That makes sense re the maxim, and I like that. Sounds good in my eye too.

PR #1947, opened 2021-03-06
reginald instead of not-defined p

xhudik:
`p` is not defined within the article. I suggest to replace `p` with properly defined `reginald`


SethTisue:
good catch, thanks!

Issue #1948, opened 2021-03-06
too general link


xhudik:
In [functional error handling](https://github.com/scala/docs.scala-lang/blob/master/_overviews/scala3-book/fp-functional-error-handling.md#L99) (lines 99-100) -it is discussed how to treat/consume  `Option[Int]`. Author provided 2 answers and added that there more possibilities with link to: `See the [Reference documentation][reference]`.  The link to whole reference guide to Scala 3 is too general.

I'd suggest to delete the link, or to provide more accurate one.

PR #1949, opened 2021-03-06
general link removed

xhudik:
General link to whole Scala 3 reference is misleading: https://github.com/scala/docs.scala-lang/issues/1948

### DIFF
--- a/_overviews/scala3-book/fp-functional-error-handling.md
+++ b/_overviews/scala3-book/fp-functional-error-handling.md
@@ -320,5 +320,3 @@ This section was long, so let’s give it a quick review:
 - Options can also be used for optional constructor or method parameters
 
 
-
-[reference]: {{ site.scala3ref }}/overview.html
--- a/_overviews/scala3-book/fp-functional-error-handling.md
+++ b/_overviews/scala3-book/fp-functional-error-handling.md
@@ -96,11 +96,6 @@ There are two common answers, depending on your needs:
 - Use a `match` expression
 - Use a `for` expression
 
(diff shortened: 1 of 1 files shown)


SethTisue:
I agree that the general link is not very helpful.

@alvinj is this the fix you'd prefer, though?

alvinj:
> is this the fix you'd prefer

That’s a very interesting question. :) 

The initial plan/hope was to link to a more-complete “Reference” book, something like an online version of _Programming in Scala_. But it may be a long time before that happens.

Hmmm, I don’t see any other detailed info under docs.scala-lang.org ...

One option is to un-comment the comment on that page. I’d have to look at that to see how ugly that would be. We could also make it an appendix page, or I could donate one of these pages to make them an appendix as well:

* [How to use the Option/Some/None pattern](https://alvinalexander.com/scala/best-practice-option-some-none-pattern-scala-idioms/)
* [How to use higher-order functions (HOFs) with Option](https://alvinalexander.com/scala/how-use-higher-order-functions-option-some-none-match-expressions/)

(I’m going down this road because I think it will be a while before we can link to a Reference book on docs.scala-lang.org.)

All that being said, that link can be deleted because I don’t think there’s anything under the Dotty docs site for this. Also, it looks like this code at the bottom of that page should also be removed because it’s only referenced that one time:

```
[reference]: {{ site.scala3ref }}/overview.html
```

b-studios:
I agree, deleting the link for now seems like the best option.

xhudik:
ok, shoud be done now!

PR #1805, opened 2020-10-27
Add consistency to parameter lists formatting

arbitrary-dev:
To make it more in line with argument list formatting of #classes

PR #1952, opened 2021-03-07
Add consistency to parameter lists formatting

arbitrary-dev:
To make it consistent with argument list formatting of [Classes](https://docs.scala-lang.org/style/declarations.html#classes).

Previously: #1805

Alternative offered, please review once again @SethTisue

### DIFF
--- a/_style/declarations.md
+++ b/_style/declarations.md
@@ -224,13 +224,20 @@ There are three main reasons you should do this:
         List("").foldLeft[Int](0, _ + _.length)
 
 For complex DSLs, or with type-names that are long, it can be difficult
-to fit the entire signature on one line. In those cases, align the
-open-paren of the parameter lists, one list per line (i.e. if you can't
-put them all on one line, put one each per line):
+to fit the entire signature on one line. In those cases, split the parameter
+lists, one parameter per line with
+[trailing commas](https://docs.scala-lang.org/sips/trailing-commas.html#motivation)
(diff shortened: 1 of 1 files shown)


SethTisue:
Is there evidence that this style is considered standard? Is it widely used?

arbitrary-dev:
As I already mentioned in [this comment](https://github.com/scala/docs.scala-lang/pull/1952#issue-586265601), this change is only to bring consistency in the current scala docs, because in "Classes" this particular formatting is already used for long argument lists, but in this place formatting is different.

>Is there evidence that this style is considered standard?
Not that I know.

>Is it widely used?
I see more of these:
```
def method(with: Very, long: Long)(
  parameter: List[]): Unit = …
```

…rather than parentheses alignment, which I only observed in C world.

Also this one is from `scalatest`:
https://github.com/scalatest/scalatest/blob/3.2.x-new/jvm/core/src/main/scala/org/scalatest/Engine.scala#L125

Overall, I'm not insisting. Could be just a proposal, an alternative, I preserved parentheses alignment case in the docs.

Me personally find this one more pleasing than others and it also facilitates having cleaner git diffs in refactorings, when parameters are each on their own lines without any parentheses. Plus reading vertically is way easier than finding parameters in horizontal plane, especially for people using terminals with short width.

arbitrary-dev:
Updated, please check.

SethTisue:
thank you!

PR #1955, opened 2021-03-10
Replace occurrences of FunSuite with AnyFunSuite

AlinGabriel:
My apologies. Forgot to replace all occurrences of FunSuite with AnyFunSuite.
Also added imports and updated hyperlink to reflect the newer class.


martijnhoekstra:
Looking good to me.

PR #1958, opened 2021-03-12
g8 changed for Scala 3

xhudik:
`sbt new scala/hello-world.g8` creates Scala 2 Hello world. We need Scala 3 (as the whole `scala3/getting-started.html` is about Scala 3)

### DIFF
--- a/scala3/getting-started.md
+++ b/scala3/getting-started.md
@@ -69,15 +69,15 @@ sbt compiles, runs, and tests your Scala code.
 To create a new Scala project with sbt:
 
 1. `cd` to an empty folder.
-1. Run this command `sbt new scala/hello-world.g8`.
+1. Run this command `sbt new scala/scala3.g8`.
 This pulls the ['hello-world' template][template-url] from GitHub.
 It also creates a _target_ folder, which you can ignore.
-1. When prompted, name the application `hello-world`.
-   This will create a project called "hello-world".
+1. When prompted, name the application `scala-3-project-template`.
+   This will create a project called "scala-3-project-template".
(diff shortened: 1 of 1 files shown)


scala3/getting-started.md
```diff
@@ -69,15 +69,15 @@ sbt compiles, runs, and tests your Scala code.
...
 
 1. `cd` to an empty folder.
-1. Run this command `sbt new scala/hello-world.g8`.
+1. Run this command `sbt new scala/scala3.g8`.
 This pulls the ['hello-world' template][template-url] from GitHub.
 It also creates a _target_ folder, which you can ignore.
-1. When prompted, name the application `hello-world`.
-   This will create a project called "hello-world".
+1. When prompted, name the application `scala-3-project-template`.
+   This will create a project called "scala-3-project-template".
```
Philippus, line 76: I don't think it makes sense to name the generated project template. scala3.g8 is the template.

xhudik: indeed, but I couldn't find any better alternative for **Hello world** in Scala 3. Even it is a template it is still Scala 3 while the original was Scala 2. I do believe it is better for end user (who wants to try Scala 3 project)

Are you aware of any Scala 3 Hello world project?

xhudik: I was thinking and removed the word `template` - does it a help a bit?

b-studios: Thanks for catching this one @xhudik. It looks good to me. Unless @Philippus objects, I would go ahead and merge.

Philippus (approved):

PR #1962, opened 2021-03-13
scalatest version update

xhudik:
scalatest goes well with scala(dotty) versions - so variable `{{site.scala-3-version}}` should be working here as well

### DIFF
--- a/_overviews/scala3-book/scala-tools.md
+++ b/_overviews/scala3-book/scala-tools.md
@@ -301,7 +301,7 @@ version := "0.1"
 scalaVersion := "{{site.scala-3-version}}"
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.3.0-SNAP3" % Test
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/scala-tools.md
```diff
@@ -301,7 +301,7 @@ version := "0.1"
 scalaVersion := "{{site.scala-3-version}}"
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.3.0-SNAP3" % Test
+  "org.scalatest" %% "scalatest" % "{{site.scala-3-version}}" % Test
```
Philippus, line 304: I don't think that's correct, scalatest doesn't follow the same version numbering as scala. I think it is good to change it to a non-SNAP version. I think version `3.2.6`, currently the latest version, works with Scala 3.

xhudik:
indeed, i got confused with some `coursier` results. Thanks @Philippus . I'm closing the request!

PR #1967, opened 2021-03-17
Add FAQ for Scala 3 macros: summon all from tuple

nicolasstucki:
Also fix broken link to ersed-terms documentation.

### DIFF
--- a/_overviews/scala3-macros/faq.md
+++ b/_overviews/scala3-macros/faq.md
@@ -23,3 +23,34 @@ x match
   case '{ $x: t } =>
     // `x: Expr[X & t]` where `t` is the precise type of `x`
 ```
+
+## How do I summon all types of a tuple type?
+If I have a type `(T1, T2, ...)` how do I generate the term for `(summon[T1], summon[T2], ...)` or get the individual expressions with the summoned values?
+
+Depending on you use case the way you will summon them will vary.
+The code you will need will depend on the kind of output you want (`Expr[Tuple]`, `List[Expr[Any]]`, or something else) and how you need errors to be reported.
+Here are two examples that should give you the basic skeleton for your variant of this code.
+
+```scala
+  def summonAllInList[T](using Type[T])(using Quotes): List[Expr[Any]] = {
+    Type.of[T] match
+      case '[ head *: tail ] =>
+        Expr.summon[head] match
+          case Some(headExpr) => headExpr :: summonAllInList[tail]
+          case _ => quotes.reflect.report.throwError(s"Could not summon ${Type.show[head]}")
+      case '[ EmptyTuple ] => Nil
+      case _ => quotes.reflect.report.throwError(s"Could not `summonAllInList` of tuple with unknown size: ${Type.show[T]}")
+  }
+```
+
+```scala
+  def summonAll[T](using Type[T])(using Quotes): Option[Expr[Tuple]]] = {
+    Type.of[T] match
+      case '[ head *: tail ] =>
+        for headExpr <- Expr.summon[head]
+            tailExpr <- summonAll[tail]
+        yield '{ headExpr *: tailExpr }
+      case '[ EmptyTuple ] => Some('{ EmptyTuple })
+      case _ => None
+  }
+```
--- a/_overviews/scala3-macros/faq.md
+++ b/_overviews/scala3-macros/faq.md
@@ -54,3 +54,21 @@ Here are two examples that should give you the basic skeleton for your variant o
       case _ => None
   }
 ```
+
+## How do I summon an expression for statically unknown type
+
+If you have a `TypeRepr` use
+```scala
+val tpe: TypeRepr = ...
+Implicits.search(tpe) match
+  case result: ImplicitSearchSuccess => result.tree
+  case _ =>
+```
+
+If there you have a `Type[_]` use
+```scala
+val tpe: Type[_] = ...
+tpe match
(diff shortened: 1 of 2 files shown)


_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,34 @@ x match
...
+Depending on you use case the way you will summon them will vary.
+The code you will need will depend on the kind of output you want (`Expr[Tuple]`, `List[Expr[Any]]`, or something else) and how you need errors to be reported.
+Here are two examples that should give you the basic skeleton for your variant of this code.
+
+```scala
+  def summonAllInList[T](using Type[T])(using Quotes): List[Expr[Any]] = {
+    Type.of[T] match
+      case '[ head *: tail ] =>
+        Expr.summon[head] match
+          case Some(headExpr) => headExpr :: Expr.summonAll[tail]
```
bishabosha, line 39: is `Expr.summonAll` meant to be `summonAllInList`?

nicolasstucki: My bad. It it supposed to be a recursive call to `summonAllInList`

nicolasstucki: Fixed here and in the second example

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
   case '{ $x: t } =>
     // `x: Expr[X & t]` where `t` is the precise type of `x`
 ```
+
+## How do I summon all types of a tuple type?
+If I have a type `(T1, T2, ...)` how do I generate the term for `(summon[T1], summon[T2], ...)` or get the individual expressions with the summoned values?
+
+Depending on you use case the way you will summon them will vary.
```
b-studios, line 30:
```suggestion
Depending on your use case the way you will summon them will vary.
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+The code you will need will depend on the kind of output you want (`Expr[Tuple]`, `List[Expr[Any]]`, or something else) and how you need errors to be reported.
```
b-studios, line 31:
```suggestion
In particular, the code you will need depends on the kind of output you want (`Expr[Tuple]`, `List[Expr[Any]]`, or something else) and how you need errors to be reported.
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+Here are two examples that should give you the basic skeleton for your variant of this code.
```
b-studios, line 32:
```suggestion
Here are two examples that should give you the basic skeleton for two different variant of this code.
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+      case '[ head *: tail ] =>
+        for headExpr <- Expr.summon[head]
+            tailExpr <- summonAll[tail]
+        yield '{ headExpr *: tailExpr }
+      case '[ EmptyTuple ] => Some('{ EmptyTuple })
+      case _ => None
+  }
+```
+
+## How do I summon an expression for statically unknown type
```
b-studios, line 58:
```suggestion
## How do I summon an expression for statically unknown types?
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+
+If you have a `TypeRepr` use
```
b-studios, line 60:
```suggestion
You can summon an expression from either a `TypeRepr` or a `Type` as shown below.

If you have a `TypeRepr` use:
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+```scala
+val tpe: TypeRepr = ...
+Implicits.search(tpe) match
+  case result: ImplicitSearchSuccess => result.tree
+  case _ =>
+```
+
+If there you have a `Type[_]` use
```
b-studios, line 68:
```suggestion
Instead, if you have a `Type[_]` use:
```

_overviews/scala3-macros/faq.md
```diff
@@ -23,3 +23,52 @@ x match
...
+```scala
+val tpe: Type[_] = ...
+tpe match
+  // Given the name `a` to the type in tpe and provide a given `Type[a]`
```
b-studios, line 72: I don't fully understand this comment here. We do assign a name to our type that we want to search an implicit for, right? What does the latter part of the comment mean?

nicolasstucki: Not sure how to explain it in other words. Technically with the user should assume is that 
```scala
tpe match
  case '[a] => 
    Expr.summon[a]
```

is equivalent to 

```scala
tpe match
  case tpe => 
    type a
    given Type[A] = tpe.asInstanceOf[Type[A]]
    Expr.summon[a]
```

nicolasstucki: This implies that we can use a given `Type[a]` on the RHS of the case. In particular for `Expr.summon[a](using given$Type_a)`.

b-studios: Ok, I see. Maybe 
```
// (1) Use `a` as the name of the unknown type and (2) bring a given `Type[a]` into scope
```
?

Not sure whether this is any better.

PR #1968, opened 2021-03-17
update JDK compat page for JDK 16 release

SethTisue:


SethTisue:
@lrytz you might still take a quick look, if you want

PR #1970, opened 2021-03-18
alter wording in SIP process about flags

SethTisue:
I don't think either the Scala 2 or 3 teams were aware this wording existed; I suggest we bring the wording in line with de facto reality.

This is just a draft for now because I may want to adjust the wording again once https://github.com/lampepfl/dotty/issues/11774 has been addressed.

### DIFF
--- a/_sips/sip-submission.md
+++ b/_sips/sip-submission.md
@@ -150,10 +150,11 @@ the formal evaluation phase.
 
 ### Merging the proposal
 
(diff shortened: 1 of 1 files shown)


SethTisue:
https://github.com/lampepfl/dotty/issues/11774 is resolved now, and I don't see a need for further changes

sjrd (approved):

PR #1976, opened 2021-04-03
fix type alias

KisaragiEffective:
make `F` higher-order type variable in L.485


joroKr21 (approved):
Thanks 👍

SethTisue:
separately, I pushed a commit that should fix CI

PR #1977, opened 2021-04-04
Update types-variance.md

nMoncho:
Should immutable arrays be consumers? I think is this referring to mutable arrays being invariant


joroKr21 (approved):
Nice catch 👍

PR #1978, opened 2021-04-06
a weird sentence changed

xhudik:


SethTisue:
the original wording wasn't actually wrong, though I think "call site" would be little more idiomatic than "use site"

but actually, I like your version better, it's plainer and more straightforward

PR #1983, opened 2021-04-10
Conversion Between Option and Collection

mariogalic:
- Clarify if `Option` is considered a collection by  summarising discussion at [Make Option extend IterableOnce #8038](https://github.com/scala/scala/pull/8038)
- Document implicit conversion `option2iterable`

### DIFF
--- a/_overviews/collections-2.13/conversion-between-option-and-the-collections.md
+++ b/_overviews/collections-2.13/conversion-between-option-and-the-collections.md
@@ -1,0 +1,53 @@
+---
+layout: multipage-overview
+title: Conversion Between Option and the Collections
+partof: collections-213
+overview-name: Collections
+
+num: 18
+previous-page: conversions-between-java-and-scala-collections
+
+permalink: /overviews/collections-2.13/:title.html
+---
+`Option` can be seen as a collection that has zero or exactly one element, and it provides a degree of interoperability with the collection types found in the package `scala.collection`. In particular, it implements the interface `IterableOnce`, which models the simplest form of collections: something that can be iterated over, at least once. However, `Option` does not implement the more comprehensive interface of `Iterable`. Indeed, we cannot provide a sensible implementation for the operation [`fromSpecific`](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/collection/Iterable.scala#L173), which is supposed to create an `Option` from a collection of possibly more than one element. Starting from [Scala 2.13](https://github.com/scala/scala/pull/8038), `Option` was made an `IterableOnce` but not an `Iterable`.
+
+Hence `Option` can be used everywhere an `IterableOnce` is expected, for example, when calling `flatMap` on a collection (or inside a for-comprehension)
+
+```scala mdoc
+for {
+  a <- Set(1)
+  b <- Option(41)
+} yield (a + b)
+// : Set[Int] = Set(42)
+``` 
+
+since the operation `flatMap` on the type `Set[Int]` takes a function returning an `IterableOnce`:
+
+```
+def flatMap[B](f: Int => IterableOnce[B]): Set[B]
+```
+
+Although `Option` does not extend `Iterable`, there exists an [implicit conversion](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/Option.scala#L19) between `Option` and `Iterable`
+
+
+```
+implicit def option2Iterable[A](xo: Option[A]): Iterable[A]
+```
+
+so although `Option[A]` is not a full collection it can be _viewed_ as one. For example,
+
+```scala mdoc
+Some(42).drop(1)
+// : Iterable[Int] = List()
+```
+
+expands to
+
+```scala mdoc
+Option.option2Iterable(Some(42)).drop(1)
+// : Iterable[Int] = List()
+```
+
+because `drop` is not defined on `Option`. A downside of the above implicit conversion is that instead of getting back an `Option[A]` we are left with an `Iterable[A]`. For this reason, `Option`’s documentation carries the following note:
+
+> Many of the methods in `Option` are duplicative with those in the `Iterable` hierarchy, but they are duplicated for a reason: the implicit conversion tends to leave one with an `Iterable` in situations where one could have retained an `Option`.
--- a/_overviews/collections-2.13/conversions-between-java-and-scala-collections.md
+++ b/_overviews/collections-2.13/conversions-between-java-and-scala-collections.md
@@ -6,6 +6,7 @@ overview-name: Collections
 
 num: 17
 previous-page: creating-collections-from-scratch
+next-page: conversion-between-option-and-the-collections
 
 languages: [ru]
 permalink: /overviews/collections-2.13/:title.html


SethTisue:
why closed?

(sometimes reviews are slow in this repo, sorry about that)

mario-galic:
@SethTisue Ah apologies, I thought the PR was just noise so closed it.

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,66 @@
...
+     |   b <- Option(41)
+     | } yield (a + b)
+val res0: scala.collection.immutable.Set[Int] = Set(42)
+``` 
+
+since `Set#flatMap` takes function to `IterableOnce`
+
+```
+scala> Set(1).flatMap
+def flatMap[B](f: Int => scala.collection.IterableOnce[B]): scala.collection.immutable.Set[B]
```
julienrf, line 40: This is not a valid Scala expression that users can copy-paste in a REPL, I think it is misleading to show it as such.

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,66 @@
...
+expands to
+
+```
+scala> Option.option2Iterable(Some(42)).drop(1)
+val res1: Iterable[Int] = List()
+```
+
+because `drop` is not defined on `Option`. Downside of the above implicit conversion is that instead of getting back specific `Option[A]` we are left with more general `Iterable[A]`. For this reason `Option` scaladoc carries the following [note](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/Option.scala#L124):
+
+> Many of the methods in here (Option) are duplicative with those in the Traversable (Iterable) hierarchy, but they are duplicated for a reason: the implicit conversion tends to leave one with an Iterable in situations where one could have retained an Option.
```
julienrf, line 66:
```suggestion
> Many of the methods in `Option` are duplicative with those in the `Iterable` hierarchy, but they are duplicated for a reason: the implicit conversion tends to leave one with an `Iterable` in situations where one could have retained an `Option`.
```

julienrf (changes requested):
Thank you @mario-galic! I agree that it would be great to document better the relation between `Option` and the Scala collections, thank you so much for helping with that!

I’m not sure about some directions taken in the article, I’ve left some comments to explain my point of view. Please let me know what you think about it!

mario-galic:
@julienrf Thank you for feedback! I have tried to address the comments.

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+```
+def fromSpecific(coll: IterableOnce[A]): C
+```
+
+For example if we attempted to defined it as `coll.headOption`, then we would throw away some elements when `coll` has more than one element. For this reason, starting [Scala 2.13](https://github.com/scala/scala/pull/8038), `Option` was made an `IterableOnce` but not an `Iterable`. To help clarify the difference between the two, mentally rename `Iterable` to `Collection` whilst `IterableOnce` becomes `Iterable`.
+
+Hence `Option` can be used everywhere an `IterableOnce` is expected, for example, when calling `flatMap` on a collection (or inside a for-comprehension)
```
julienrf, line 21: I would propose to go less into the technical details. I think what is important is to let the reader what is supported, but I don’t think we should explain much what it would entail to make `Option` implement `Iterable`. (Maybe others disagree?)
Also, I don’t think it helps to propose a mental model where `Iterable` is renamed to `Collection` and `IterableOnce` is renamed to `Iterable`.

```suggestion
`Option` can be seen as a collection that has zero or exactly one element, and it provides a degree of interoperability with the collection types found in the package `scala.collection`. In particular, it implements the interface `IterableOnce`, which models the simplest form of collections: something that can be iterated over, at least once. However, `Option` does not implement the more comprehensive interface of `Iterable`. Indeed, we cannot provide a sensible implementation for the operation [`fromSpecific`](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/collection/Iterable.scala#L173), which is supposed to create an `Option` from a collection of possibly more than one element. Starting from [Scala 2.13](https://github.com/scala/scala/pull/8038), `Option` was made an `IterableOnce` but not an `Iterable`.

Hence `Option` can be used everywhere an `IterableOnce` is expected, for example, when calling `flatMap` on a collection (or inside a for-comprehension)
```

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+scala> for {
+     |   a <- Set(1)
+     |   b <- Option(41)
+     | } yield (a + b)
+val res0: scala.collection.immutable.Set[Int] = Set(42)
+``` 
+
+since `Set#flatMap` takes function to `IterableOnce`
```
julienrf, line 31:
```suggestion
since the operation `flatMap` on the type `Set[Int]` takes a function returning an `IterableOnce`:
```

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+def flatMap[B](f: Int => scala.collection.IterableOnce[B]): scala.collection.immutable.Set[B]
```
julienrf, line 34:
```suggestion
def flatMap[B](f: Int => IterableOnce[B]): Set[B]
```

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Having said `Option` is not a full collection in the sense of extending `Iterable`, nevertheless there exists an [implicit conversion](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/Option.scala#L19) between an `Option` and `Iterable` (or `Collection` in our mental model)
```
julienrf, line 37:
```suggestion
Although `Option` does not extend `Iterable`, there exists an [implicit conversion](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/Option.scala#L19) between `Option` and `Iterable`
```

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+expands to
+
+```
+scala> Option.option2Iterable(Some(42)).drop(1)
+val res1: Iterable[Int] = List()
+```
+
+because `drop` is not defined on `Option`. Downside of the above implicit conversion is that instead of getting back specific `Option[A]` we are left with more general `Iterable[A]`. For this reason `Option` scaladoc carries the following [note](https://github.com/scala/scala/blob/6c68c2825e893bb71d6dc78465ac8c6f415cbd93/src/library/scala/Option.scala#L124):
```
julienrf, line 58:
```suggestion
because `drop` is not defined on `Option`. A downside of the above implicit conversion is that instead of getting back an `Option[A]` we are left with an `Iterable[A]`. For this reason, `Option`’s documentation carries the following note:
```

julienrf (commented):
Thanks for your quick response!

What do you think of the following additional suggestions?

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,53 @@
+---
+layout: multipage-overview
+title: Conversion Between Option and Collection
```
julienrf, line 3:
```suggestion
title: Conversion Between Option and the Collections
```

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,53 @@
...
+partof: collections-213
+overview-name: Collections
+
+num: 18
+previous-page: conversion-between-java-and-scala-collections
```
julienrf, line 8: Should we also set the `next-page` of the page `conversion-between-java-and-scala-collections`?

_overviews/collections-2.13/conversion-between-option-and-collection.md
```diff
@@ -0,0 +1,53 @@
...
+
+Hence `Option` can be used everywhere an `IterableOnce` is expected, for example, when calling `flatMap` on a collection (or inside a for-comprehension)
+
+```
+scala> for {
+     |   a <- Set(1)
+     |   b <- Option(41)
+     | } yield (a + b)
+val res0: scala.collection.immutable.Set[Int] = Set(42)
+```
```
julienrf, line 22: Can we use mdoc to typecheck and compute the result of this snippet? I couldn’t find how to do that in README.md maybe another contributor can help us?

mario-galic:
@julienrf They look good. Personally, the "mental model" which I picked up from reading conversation at [Make Option extend IterableOnce #8038](https://github.com/scala/scala/pull/8038) made the distinction click for me, although conceptualising collection to be something that can be iterated at least once also seems perfectly valid view.

_overviews/collections-2.13/conversion-between-option-and-collection.md
mario-galic: I was able to [process single markdown file](https://scalameta.org/mdoc/docs/installation.html#process-single-markdown-file) locally like so

```
coursier launch org.scalameta:mdoc_2.12:2.2.20 -- --in _overviews/collections-2.13/conversion-between-option-and-the-collections.md
```

This generates output markdown file with 

```scala
Some(42).drop(1)
// res1: Iterable[Int] = List()
```

Should I commit these outputs?

julienrf: No, the output of mdoc should be automatically displayed on the rendered website. Did you try to run the website locally? https://github.com/scala/docs.scala-lang#quickstart

mario-galic: I tried running but mdoc evaluation does not show up. It seems currently even in production mdoc fences are not rendered evaluated, for example, see how [for-comprehensions.md](https://raw.githubusercontent.com/scala/docs.scala-lang/main/_tour/for-comprehensions.md) is rendered at https://docs.scala-lang.org/tour/for-comprehensions.html

julienrf: Oh, good point, maybe they are just typechecked, but not rendered… Maybe there is a modifier to use to display the result?

mario-galic: It looks like they are only typechecked according to https://github.com/scala/docs.scala-lang/blob/main/contribute.md#code-blocks

julienrf: Thanks for investigating! Then I think what we can do is to show the output as an (unchecked) comment:

~~~ scala
for {
  x <- Set(41)
  y <- Option(1)
} yield x + y
// : Set[Int] = Set(42)
~~~

mario-galic: Done. As a side note, seems because similar workarounds are used throughout docs to show outputs it would be hard to ever switch on automatic mdoc output rendering since it would effectively display duplicate outputs.

julienrf (approved):
:+1:  Thank you for your contribution!

mariogalic:
Ops I think i screwed up git history here by trying to open another PR :(

julienrf:
Hello @mariogalic, would it be possible for you to clean the git history? Ideally, each contribution should be self-contained in a single commit.

mariogalic:
Hi @julienrf, it should be done - I squashed all commits into single one and force pushed.

PR #1985, opened 2021-04-12
remove remaining references to tut

SethTisue:
tut is obsoleted by mdoc, and also tut will disappear when Bintray disappears

note that I didn't replace the mini how-to-tut tutorial with a mini how-to-mdoc tutorial, because I am lazy.  if a volunteer wants to add that back in for mdoc, that would be cool.

fyi @tpolecat @ckipp01

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,4 @@ _site
 vendor/bundle
 .idea/
 /coursier
-/tut-tmp/
 .sass-cache/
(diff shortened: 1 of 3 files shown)


ckipp01 (approved):
👍🏼

SethTisue:
fyi @julienrf, who probably wrote the tut tutorial I excised

julienrf:
Thanks for the @mention, I’ve added a note to myself to add back this part of the tutorial, but with mdoc instead.

PR #1992, opened 2021-04-23
Update links to point to Scala 2.13 specific documents

mariogalic:
* This 2.13 document was pointing to older Scala introduction overview
* Point "The Collections API" link which talks about common operations to more specific overviews/collections-2.13/trait-iterable.html


SethTisue:
Thank you!

PR #1999, opened 2021-04-30
Add more sections to Scala 3 macro reflection tutorial 

nicolasstucki:


_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -12,13 +12,23 @@ It provides a view on the *Typed Abstract Syntax Trees* and their properties suc
 ## How to use the API
 
 Accessing this API needs an import that depends the current `Quotes`.
+The following definition
```
b-studios, line 15: The text is somewhat fragmented here. Should it be something like "The following code example illustrates / gives the definition for / ..."?

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -12,13 +12,23 @@ It provides a view on the *Typed Abstract Syntax Trees* and their properties suc
 ## How to use the API
 
 Accessing this API needs an import that depends the current `Quotes`.
```
b-studios, line 14:
```suggestion
Accessing this API needs an import that depends on the current `Quotes`, which are implicitly in scope.
```

b-studios: Maybe this needs a bit more explanation.  This usage pattern is somewhat obvious to me, but it might not to newcomers. What about:

"The reflection API is defined in the type `Quotes`.  The actual instance depends on the current scope, in which quotes or quoted pattern matching is used. Hence, every macro method receives `Quotes` as an additional argument. Since `Quotes` is contextual, to access it members we either need to name the parameter, or summon it. The following definition of the standard library provides the canonical way of accessing it."

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
 Each type is also a module with some _static-ish_ methods, for example in the [`TypeReprModule`](https://dotty.epfl.ch/api/scala/quoted/Quotes$reflectModule$TypeReprModule.html) we can find the method `TypeRepr.of[T]` which will create an instance of `Type` containing `T`.
 
 
-## Relation with expressions
-<!-- Term vs Expr -->
-<!-- Safty -->
-*Coming soon*
+## Relation with Expr/Type
+
+Expressions `Expr[T]` can be seen as wrapper around a `Term` that knows statically the type of the term.
```
b-studios, line 46:
```suggestion
Expressions `Expr[T]` can be seen as wrappers around a `Term`, where `T` is the statically known type of the term.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+Bellow we use the `asTerm` extension method to transform the expression into a term.
```
b-studios, line 47:
```suggestion
Below we use the extension method  `asTerm` to transform the expression into a term.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+This extension method is only available after importing `quotes.reflect.asTerm`.
+Then we use `asExprOf[Int]` to transform the term back into `Expr[Int]`.
+This operation will fail if the term does not have provided type or if the term is not a valued expression.
```
b-studios, line 50: Maybe remind users what a `valued` expression is?

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+
+```scala
+def f(x: Expr[Int])(using Quotes): Expr[Int] =
+  import quotes.reflect.*
+  val tree: Term = x.asTerm
+  val expr: Expr[Int] = tree.asExprOf[Int]
+  expr
+```
+
+Similarly, we can also a `Type[T]` as a wrapper over a `TypeRepr` that knows statically its type.
```
b-studios, line 60:
```suggestion
Similarly, we can also see `Type[T]` as a wrapper over `TypeRepr`, with `T` being the statically known type.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+To get a `TypeRepr` we use `TypeRepr.of[X]` which expects a given `Type[X]` in scope (similar to `Type.of[X]`).
+We can also transform it back into a `Type[?]` using the `asType` method.
+As the type of `Type[?]` is not statically know we need to name it with an existential type to use it, can be done using the `'[t]` pattern.
```
b-studios, line 63:
```suggestion
As the type of `Type[?]` is not statically know we need to name it with an existential type to use it. This can be achieved using a `'[t]` pattern.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+  import quotes.reflect.*
+  val tpe: TypeRepr = TypeRepr.of[T]
+  tpe.asType match
+    case '[t] => '{ val x: t = ${...} }
+  ...
+```
+
+## Macro API design
+
+It will be often useful to create helper methods or extractors that to perform some common logic of your macros.
```
b-studios, line 76:
```suggestion
It will be often useful to create helper methods or extractors that perform some common logic of your macros.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+
+The simples methods will be those that only mention `Expr`, `Type` and `Quotes` in their signature.
```
b-studios, line 78:
```suggestion
The simplest methods will be those that only mention `Expr`, `Type`, and `Quotes` in their signature.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+  def unapply(using Quotes)(x: quotes.reflect.Term) =
+    ...
+    Some(y)
+```
+
+> **Avoid saving the `Quotes` context in a field.**
+> `Quotes` in fields inevitably make its use harder by hitting errors involving `Quotes` with different paths.
+>
+> Usually these patterns have been seen in code that uses the Scala 2 ways to define extension methods or contextual unapplies.
+> Now that we have `given` parameters that can be added before other parameters, all these old workarounds are no needed.
```
b-studios, line 115:
```suggestion
> Now that we have `given` parameters that can be added before other parameters, all these old workarounds are not needed anymore.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+> The new abstraction make it simpler on the definition site, use site and also have a better performance.
```
b-studios, line 116:
```suggestion
> The new abstraction make it simpler on the definition site, use site, and also have a better performance.
```

b-studios: What are these peformance gains here?

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+
+## Debugging
+
+### Runtime checks
+
+Unlike `Expr`, it is not possible to guaranty statically that the generated code will be well-formed.
```
b-studios, line 122:
```suggestion
Unlike `Expr`, with `Term` it is not possible to guaranty statically that the generated code will be well-formed.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+
+It is recommended to enable the `-Xcheck-macros` flag while developing macros or on the tests for the macro.
+This flag will enable extra runtime checks that will try to find ill-formed trees or types as soon as they are created.
+
+There is also the `-Ycheck:all` flag that checks all compiler invariants for tree well-formedness.
+These check will usually fail with an assertion error.
+
+### Printing the trees
+
+The `toString` of types in `quotes.reflect` is no great for debugging as they show the internal representation rather than the `quotes.reflect` representation.
```
b-studios, line 133:
```suggestion
The `toString` methods of types in `quotes.reflect` are not great for debugging as they show the internal representation rather than the `quotes.reflect` representation.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+In many cases these are similar but they may lead the debugging process astray.
+
+Hence, `quotes.reflect.Printers` provide a set of useful printers for debugging.
+Notably the `TreeStructure`, `TypeReprStructure` and `ConstantStructure` can be quite useful.
+These will print the three structure following loosely the extractors that would be needed to match it.
```
b-studios, line 138:
```suggestion
These will print the tree structure following loosely the extractors that would be needed to match it.
```

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +41,126 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
+
+```scala
+tree match
+  case Ident(_) =>
+  case Select(_, _) =>
+  ...
+  case _ =>
+    throw new MatchError(tree.show(using Printer.TreeStructure))
+```
+This way if a case is missed the error will report a familiar structure that can be copy pasted to start fixing the issue.
```
b-studios, line 155:
```suggestion
This way, if a case is missed the error will report a familiar structure that can be copy-pasted to start fixing the issue.
```

_overviews/scala3-macros/tutorial/reflection.md
nicolasstucki: It was a typo. It should be `valid expression`.

_overviews/scala3-macros/tutorial/reflection.md
nicolasstucki: no object instantiations for extension methods.

_overviews/scala3-macros/tutorial/reflection.md
nicolasstucki: I added an example

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -93,11 +93,11 @@ def evalAndUse[X](x: Expr[X])(using Type[X])(using Quotes) = '{
 This code will be equivalent to the more verbose
 ```scala
 def evalAndUse[X](x: Expr[X])(using t: Type[X])(using Quotes) = '{
-  val x2: t.T = $x
+  val x2: t.Underlying = $x
   ... // use x2
 }
 ```
-Note that `Type` has a type member called `T` that refers to the type held within the `Type`, in this case `t.T` is `X`.
+Note that `Type` has a type member called `T` that refers to the type held within the `Type`, in this case `t.Underlying` is `X`.
```
bishabosha, line 100:
```suggestion
Note that `Type` has a type member called `Underlying` that refers to the type held within the `Type`, in this case `t.Underlying` is `X`.
```

_overviews/scala3-macros/tutorial/reflection.md
b-studios: Great, thanks!

_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -31,11 +44,127 @@ As it is a subtype of `Tree` we can also look into the `TreeMethods` to find mor
...
-<!-- Term vs Expr -->
-<!-- Safty -->
-*Coming soon*
+## Relation with Expr/Type
+
+Expressions `Expr[T]` can be seen as wrappers around a `Term`, where `T` is the statically known type of the term.
+Below we use the extension method `asTerm` to transform the expression into a term.
+This extension method is only available after importing `quotes.reflect.asTerm`.
+Then we use `asExprOf[Int]` to transform the term back into `Expr[Int]`.
+This operation will fail if the term does not have provided type or if the term is not a valid expression.
```
b-studios, line 53:
```suggestion
This operation will fail if the term does not have the provided type (here `Int`) or if the term is not a valid expression.
```

b-studios:
LGTM besides maybe this small final remark above.

PR #2000, opened 2021-04-30
Avoid text fragmentation with inline code

nicolasstucki:
This change removes the extra padding around inline code snippets to make the text simpler to read. 

## New 

Screenshot 2021-04-30 at 11 29 10

Screenshot 2021-04-30 at 11 30 11

## Old
Screenshot 2021-04-30 at 11 22 40
Screenshot 2021-04-30 at 11 16 08

### DIFF
--- a/_sass/layout/type-md.scss
+++ b/_sass/layout/type-md.scss
@@ -156,8 +156,8 @@
             background: $gray-lighter;
             color: #667b83;
             // border: 1px solid #ced7d7;
-            padding: 0 6px;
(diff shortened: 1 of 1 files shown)


b-studios:
I am fine with this change, but since it not only affects the Scala 3 docs, but the whole site, maybe @SethTisue also wants to say something about it?

SethTisue (approved):
I've never understood why the old way was chosen. I guess a designer must have chosen it, but I've never liked it.

PR #2001, opened 2021-05-01
Update Value Classes Intro

aelfric:
When I was reading the documentation, I was afraid I couldn't use value classes because they were labeled as a "new" feature and I was working on a 2.11 project.


SethTisue:
Thank you!

PR #2008, opened 2021-05-10
Correct links to coursier installation

vincenzobaz:
This PR addresses the issues found after https://github.com/scala/docs.scala-lang/pull/1996 which are linked to the platform specific instructions.

There is no need to maintain detailed information about installation of CS as they are on the project's website.
Unix and osx have the same instructions.

The Japanese version needs to be fixed as well.

### DIFF
--- a/_getting-started/index.md
+++ b/_getting-started/index.md
@@ -115,24 +115,21 @@ There are a multitude of mailing lists and real-time chat rooms in case you want
  -- This is handled in `resources/js/functions.js`.
  --> 
 <div style="display:none" id="stepOne-linux">
-       <code class="hljs">$ curl -Lo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup </code> <br>
-</div>
- 
-<div style="display:none" id="stepOne-unix">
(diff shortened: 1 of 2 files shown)


vincenzobaz:
I am not sure if this works for OSx users.

I would like to drop the platform specific text and just redirect to coursier's installation guide which then differentiates platforms

julienrf:
Hey @vincenzobaz is there a description of the issues found after #1996?



> I would like to drop the platform specific text and just redirect to coursier's installation guide which then differentiates platforms

Maybe eventually we should aim for the opposite: hosting the `cs` documentation directly in docs.scala-lang.org, so that we have everything in one place.

vincenzobaz:
I think it is better if we let the coursier website maintain instructions to install coursier.

b-studios:
In https://github.com/scala/scala-lang/pull/1246/files#diff-427692f739e1a5aad82fee647e2fa31f8aed46ab3fc37db0e6f8e4305ee34ff9R38-R75 I redirect to the coursier website. Eventually, I would like to drop the various getting started guides in favor of the one at scala-lang.org.

PR #2009, opened 2021-05-10
Update fp-pure-functions.md

b-studios:
The discussion of impurity was technically not correct. 
`foreach` alone is not impure. 
Also `const Unit` is a perfectly pure function.

So we need to be a bit more precise here in order to not confuse beginners.

### DIFF
--- a/_overviews/scala3-book/fp-pure-functions.md
+++ b/_overviews/scala3-book/fp-pure-functions.md
@@ -51,7 +51,7 @@ Most methods on the Scala collections classes also work as pure functions, inclu
 Conversely, the following functions are _impure_ because they violate the definition.
 
 - `println` -- methods that interact with the console, files, databases, web services, sensors, etc., are all impure.
-- `getHour` -- date and time related methods are all impure because their output depends on something other than their input parameters 
+- `currentTimeMillis ` -- date and time related methods are all impure because their output depends on something other than their input parameters 
 - `sys.error` -- exception throwing methods are impure because they do not simply return a result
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -50,24 +50,20 @@ Most methods on the Scala collections classes also work as pure functions, inclu
 
 Conversely, the following functions are _impure_ because they violate the definition.
 
-The `foreach` method on collections classes is impure because it’s only used for its side effects, such as printing to STDOUT.
+- `println` -- methods that interact with the console, files, databases, web services, sensors, etc., are all impure.
+- `getHour` -- date and time related methods are all impure because their output depends on something other than their input parameters
```
julienrf, line 54: Calling `getHour` on a `LocalTime` always returns the same value, so this might be confusing. Consider maybe using `currentTimeMillis`?

julienrf (approved):

_overviews/scala3-book/fp-pure-functions.md
b-studios: Thanks you are, I was just using those since @alvinj mentioned them in the original doc.

b-studios, line 54:
```suggestion
- `currentTimeMillis ` -- date and time related methods are all impure because their output depends on something other than their input parameters 
```

alvinj: Thanks, `currentTimeMillis` is a much more obvious example!

Issue #1711, opened 2020-06-17
Scala compiler Options needs updating


gavinbaumanis:
https://docs.scala-lang.org/overviews/compiler-options/index.html
mentions;
-optimize or -optimise Enables optimizations.

This option is now deprecated.
(probably need to check all others, too.)

SethTisue:
@gavinbaumanis interested in returning to this?

gavinbaumanis:
I am returning to this now. Thanks @SethTisue.

gavinbaumanis:
@SethTisue  : Do you want a giant issue / document at the end of this  - or do you want individual issues raised for the things that I find?
I don't mind either way.

SethTisue:
uh, I guess a single issue for now, and then I suppose we might choose later to split things out

gavinbaumanis:
@SethTisue : OK so I am editing the code to correct the documentation (I am happy to go through it and see if I can do it) - Or am I just flagging the issues I have found between the two different flavours (*.YML and the docs)? 
Thanks!

SethTisue:
@gavinbaumanis the ambition level is up to you — any amount of improvement is better than nothing. Is the automatic generation aspect even currently working? If not, that seems like the single most important thing to address.

gavinbaumanis:
@SethTisue  : I am happy to try it all! : Though, I have had a look at the code and it is beyond me at the moment to fix without context / help. Do you know anyone who knows about the document generation engine - that I could ask directly - at the times I get stuck? Thanks.

SethTisue:
> Do you know anyone who knows about the document generation engine

I believe @exoego is the original author. @som-snytt and I can try to help as well. Usually the best way to get help is to open a draft PR in scala/scala and ask for it there. Chat rooms are catch-as-catch-can, but asking things on https://gitter.im/scala/contributors definitely works at least some of the time.

SethTisue:
Is there a new volunteer who would like to tackle this?

som-snytt:
It's funny the issues that seemed so important right before the global pandemic.

Shorla:
Hi @SethTisue, I don't mind getting started on this.

Shorla:
How can I get started on it?

SethTisue:
@Shorla are you able to run `generateDocsData`?

Shorla:
@SethTissue I tried running it with docs/mdoc, it said commamd not found.
How do I run it?

som-snytt:
[...]

PR #2014, opened 2021-05-11
Update release and target options

BalmungSan:
I also noticed that `17` also works, but I guess it is still experimental according to [**JDK COMPATIBILITY**](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html). So, I am not sure if I should also add that; maybe with an _(experimental)_ flag.
Also, maybe we should drop everything below `8`?

----

BTW I also noticed that the `scalac` built-in help command shows:

> -release <release>           Compile for a specific version of the Java platform. Supported targets: 6, 7, 8, 9

Which AFAIK is wrong, where may I fix that?

### DIFF
--- a/_data/compiler-options.yml
+++ b/_data/compiler-options.yml
@@ -260,7 +260,7 @@
       type: "String"
       arg: "release"
       default:
-    description: "Compile for a specific version of the Java platform. Supported targets: 6, 7, 8, 9"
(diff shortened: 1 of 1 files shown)


SethTisue:
This file is supposed to be auto-generated — what we need is a volunteer to get that going again: https://github.com/scala/docs.scala-lang/issues/1711

som-snytt:
Relatedly https://github.com/scala/scala/pull/9982

PR #2020, opened 2021-05-14
Log.log -> Logger.log

siddhant3s:
The text doesn't matches the example code. I believe this should be Logger.log. I also clarified what will happen on the runtime.


nicolasstucki (approved):

siddhant3s:
I hope there is nothing else to do from my end.

PR #2021, opened 2021-05-14
Added missing `inline`

siddhant3s:
I believe that the author intended to have inline here since they talk about its expansion.


b-studios:
LGTM, @nicolasstucki this was intended, right?

siddhant3s:
Please let me know if there are any steps that need to be taken from my end. Thanks

julienrf:
Thank you @siddhant3s !

PR #2027, opened 2021-05-18
SIP - Sealed Types

dwijnand:
[Rendered](https://github.com/dwijnand/docs.scala-lang/blob/sealed-types/_sips/sips/2020-05-18-sealed-types.md)
[Discussion](https://contributors.scala-lang.org/t/sip-sealed-types/5082)


_sips/sips/2020-05-18-sealed-types.md
```diff
@@ -0,0 +1,278 @@
...
+opaque type Neg <: Num = Num
+```
+
+The match desugars into an ordinal method, that reuses the logic to associate an ordinal for each
+case:
+
+```scala
+extension (n: Num):
+  def ordinal: Int = (n: Int) match {
+    case n if n > 0 => 0
```
Jasper-M, line 96: I think the ordinal cases are in the wrong order.

dwijnand: Thank you, fixed.

joroKr21:
Is there a discussion already?

lrytz:
@joroKr21 https://contributors.scala-lang.org/t/sip-sealed-types/5082/12

dwijnand:
Why can't this be merged?

SethTisue:
Sure, we can merge it. But I think the filename should be `2021-05-18-sealed-types.md` (not `2020...`)?

Issue #2028, opened 2021-05-18
Documentation to take advantage of larger screens


fernandoracca:
When reading on reasonably large screens, particularly 4K monitors, the main area for content is really small, and there's huge amount of wasted empty space.

This selector changes a single property which expands to fill the screen, which makes a large difference. 

```javascript
document.querySelectorAll(“.wrap").forEach ( el => el.style.maxWidth = "2048px" )
```

Consider including such support directly on the website.

fernandoracca:
I actually have a PR ready. It's tested on both my laptop's monitor and an external 27" 4K monitor, and it fills up the space nicely.

b-studios:
I agree that the docs benefit from this change:

![image](https://user-images.githubusercontent.com/408265/118983847-2653aa80-b97d-11eb-80bd-9f4ea2673c07.png)


However, the landing page really is not designed with this in mind:

![image](https://user-images.githubusercontent.com/408265/118983899-39ff1100-b97d-11eb-9df5-ba8279fe0c52.png)

fernandoracca:
Thanks for the feedback @b-studios .
I'll look into it tonight, to see if i can also improve that page. I'm not much of a CSS person, so i can't promise a substantial change. 
If you have any suggestion of how to structure it, please do let me know.

gkepka:
Completed by #2030

PR #2030, opened 2021-05-18
style: Improve UX for users with large screens #2028

fernandoracca:
Expand the main content area to fill up all available space.

### DIFF
--- a/_sass/vendors/neat/settings/_grid.scss
+++ b/_sass/vendors/neat/settings/_grid.scss
@@ -22,7 +22,7 @@ $grid-columns: 12 !default;
 ///
 /// @type Number (Unit)
 ///
-$max-width: 2048px!default;
+$max-width: 1280px!default;
 
 /// When set to true, it sets the box-sizing property of all elements to `border-box`. Set with a `!global` flag.
 ///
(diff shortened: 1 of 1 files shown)


b-studios:
Also, it would be important to make the same change over at https://github.com/scala/scala-lang

b-studios:
Just for comparison: 

docs now

image

docs with 1280px

image

landing page with 1280px

image

scala home with 1280px

b-studios:
LGTM, @SethTisue or @julienrf could we have a second pair of eyes?

b-studios (approved):

julienrf (approved):

julienrf:
I personally don’t like when rows are too wide but 1280px seems OK to me. Maybe an alternative solution to take advantage of large screens would be to show sidebars with additional information (although we already have some sidebars on some pages) rather than increasing the width of the main content? Anyway, that could be worked on a separate PR, I think this PR finds a good compromise.

b-studios:
@fernandoracca could you maybe open a PR with the same change at https://github.com/scala/scala-lang ? This way we make sure the two pages are consistent.

PR #2044, opened 2021-05-21
Mention symbols `$` and `_` in operators.md

jeremyrsmith:
Explicitly mention that `$` and `_` are considered "letters" for operator precedence.

Only updated those translations which included the precedence list (rather blindly – hopefully in a way that makes sense in all these languages)


jeremyrsmith:
I did this because I was surprised by the behavior of `$` as far as operator precedence based on the current docs; I don't think of it as a letter, but looking at the compiler source reveals that indeed it (along with `_`) is explicitly included in the "letter" category. Just figured it wouldn't hurt to explicitly mention it (even if it's not something that people are encouraged to use)

Jasper-M (approved):
LGTM. Perhaps this should also be added to the spec.

som-snytt:
The spec says what a letter is. https://github.com/scala/scala/blame/v2.13.6/spec/01-lexical-syntax.md#L23

Case (upper/lower) in [pattern match](https://github.com/scala/scala/commit/a853608f025d1beb61f7ae06603d34b342386d05) was subsequently tweaked @martijnhoekstra oh actually the reasoning was that underscore is an exception https://github.com/scala/scala/pull/8110#discussion_r289917248

Note that in other tickets, "you're not supposed to use `$`".

martijnhoekstra:
"you're not supposed to use $" is IMO an excellent rule for creating identifiers, but if you have one for, I don't know, interop reasons, where you need to interoperate with someone who doesn't care about reserved characters, you're going to have to deal with it. Not documenting the precedence rule for that as, I don't know, punishment for having to deal with it doesn't make sense to me. I don't have an opinion on whether it should be mentioned again here that $ is reserved for compiler generated identifiers.

Extending this PR with spec work in https://scala-lang.org/files/archive/spec/2.13/06-expressions.html#infix-operations with these precedence rules would sound right to me, either in this PR or as a follow-up: this PR looks useful and good to me on its own too. That "all letters" here means everything in the lexical syntax rule `letter` which includes `_` and `$` isn't at all obvious to me.

SethTisue:
thank you! volunteer to submit a spec PR?

som-snytt:
I'll follow up on my other PR, where I was already mired in speck. I see the ask here is to call out _ and $ in precedence table, and of course I'm not going to say they're letters, what's the problem? but I will say what's a "special" character? At my funeral, I hope they call me a special character.

PR #2049, opened 2021-05-24
Fixed what I believe to be typos

nscarcella:
I think this `if` expression is missing either a `then` or parenthesis.

I'm not so sure about the other change (replacing `Any` type for `Matchable`), since it seems to be valid Scala code, at least on worksheets. But maybe it would be better presented this way, to avoid conflicting with this quote from other section of the book:

> The top-type `Any` has a subtype Matchable, which is used to mark all types that we can perform pattern matching on. We will not go into details here, but in summary, it means that we cannot pattern match on values of type `Any`, but only on values that are a subtype of `Matchable`.

### DIFF
--- a/_overviews/scala3-book/methods-most.md
+++ b/_overviews/scala3-book/methods-most.md
@@ -233,7 +233,7 @@ Here’s a method named `isTruthy` that implements the Perl definitions of `true
 
 ```scala
 def isTruthy(a: Any) =
-  if a == 0 || a == ""
+  if a == 0 || a == "" then
     false
   else
     true
--- a/_overviews/scala3-book/methods-most.md
+++ b/_overviews/scala3-book/methods-most.md
@@ -261,8 +261,9 @@ def isTruthy(a: Matchable) = a match
   case _ => true
 ```
 
-This method works just like the previous method that used an `if`/`else` expression.
+This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type because it's a safer alternative that will still admit any concrete value or reference.
 
+For more details on the `Matchable` trait, see the [Reference documentation][reference_matchable].
 
 
 ## Controlling visibility in classes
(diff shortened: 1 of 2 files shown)


_overviews/scala3-book/methods-most.md
```diff
@@ -256,7 +256,7 @@ A `match` expression can also be used as the entire method body, and often is.
 Here’s another version of `isTruthy`, written with a `match` expression :
 
 ```scala
-def isTruthy(a: Any) = a match
+def isTruthy(a: Matchable) = a match
```
julienrf, line 259: May I ask to add a comment/link about `Matchable`?

julienrf (changes requested):
Thank you @nscarcella, I agree with your changes but I think we should give a short explanation about `Matchable` (or just a link to another page of the docs that explains `Matchable`)

_overviews/scala3-book/methods-most.md
nscarcella: Hi @julienrf! I added a short `Matchable` mention and a link to the more complete reference page. Let me know if the phrasing is right.

_overviews/scala3-book/methods-most.md
```diff
@@ -256,13 +256,14 @@ A `match` expression can also be used as the entire method body, and often is.
...
   case 0 | "" => false
   case _ => true
 ```
 
-This method works just like the previous method that used an `if`/`else` expression.
+This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type because it's a safer alternative that will still admit any concrete value or reference.
```
julienrf, line 264: What about the following?

```suggestion
This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type to accept any value that supports pattern matching.
```

julienrf (approved):
Thank you @nscarcella !

PR #2051, opened 2021-05-25
Update MOOCs links and descriptions

julienrf:
Vincenzo noticed that this page is still available through the Scala 2 banner, which is partially wrong because most of the courses have already been updated to Scala 3. I think we need to take a bit more time to overhaul the co-existence of the Scala 2 and Scala 3 documentation (most likely, we want to move most of the current content of https://docs.scala-lang.org/learn.html into a page https://docs.scala-lang.org/scala2/learn.html, so that we keep the URL https://docs.scala-lang.org/learn.html for documentation up to date with the latest stable release of Scala. But this is not the path that has been taken so far… e.g. https://docs.scala-lang.org/overviews/scala-book/introduction.html shows the Scala 2 book, not Scala 3’s). In the meantime, I would like to merge this as it is, at least for SEO reasons.

### DIFF
--- a/learn.md
+++ b/learn.md
@@ -36,7 +36,7 @@ this course in the following video:
   <iframe width="560" height="315" src="https://www.youtube.com/embed/MSDJ7ehjrqo" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 </div>
 
-After taking this course, you might be interested in growing further your
+After taking this course, you might be interested in improving your
 skills in specific areas by taking the courses [Parallel Programming],
 [Big Data Analysis with Scala and Spark], or [Programming Reactive Systems].
 
--- a/learn.md
+++ b/learn.md
@@ -61,6 +61,16 @@ and Spark.
 teaches how to write responsive, scalable, and resilient systems with the
 library Akka.
 
+### Scala 2 Courses
+
+The above courses use Scala 3 (except the Spark courses). If needed, you can find
+the (legacy) Scala 2 version of our courses here:
+
+- [Functional Programming Principles in Scala (Scala 2 version)](https://www.coursera.org/learn/scala2-functional-programming)
+- [Functional Program Design (Scala 2 version)](https://www.coursera.org/learn/scala2-functional-program-design)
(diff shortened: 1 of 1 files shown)


julienrf:
Rendering looks like this:
Screenshot_2021-05-25 Online Resources

learn.md
```diff
@@ -11,25 +11,75 @@ There are a handful of websites where you can interactively run Scala code in yo
...
+modeling business domains, implementing business logic, designing large
+systems made of components, handling errors, manipulating data, running 
+concurrent tasks in parallel, testing your code). You can learn more about
+this course in the following video:
+
+<div style="text-align: center">
+  <iframe width="560" height="315" src="https://www.youtube.com/embed/MSDJ7ehjrqo" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+</div>
+
+After taking this course, you might be interested in growing further your
```
b-studios, line 39:
```suggestion
After taking this course, you might be interested in improving your
```

learn.md
```diff
@@ -11,25 +11,75 @@ There are a handful of websites where you can interactively run Scala code in yo
...
+skills in specific areas by taking the courses [Parallel Programming],
+[Big Data Analysis with Scala and Spark], or [Programming Reactive Systems].
+
+### Scala Specialization
+
+The [Scala Specialization] provides a hands-on introduction to functional programming using Scala. You can access the courses
+material and exercises by either signing up for the specialization or auditing the courses individually. The
+   specialization has the following courses.
+* [Functional Programming Principles in Scala],
```
b-studios, line 48: I never used whitespace in markdown url shortcuts. Did you check that this links correctly?

b-studios (approved):
LGTM (see a few comments).

I agree that we need a structural change regarding the Scala 2 and Scala 3 contents. If you want to render it with a Scala 3 header, you can try adding the following to the preface:

```
scala3: true
```

learn.md
julienrf: Yes :)

julienrf:
Thank you for the review, @b-studios!

I’ve tried adding `scala3: true` in the metadata and indeed it shows a Scala 3 banner. But I don’t think we should do this change now, because the `learn` page contains mixed content about Scala 2 and Scala 3. So, for now I’m leaving it without the `scala3: true` setting.

PR #2052, opened 2021-05-27
Suggest Tuple.fromProductTyped for Unapply

stephennancekivell:


stephennancekivell:
opps. Its not exactly the same.

PR #2072, opened 2021-06-06
Fix typo in macro tutorial

armanbilge:
Nearly tripped me up 😉


julienrf:
Good catch :blush: thank you for reporting, this has been fixed in another PR.

PR #2073, opened 2021-06-07
Some suggestions in the Scala 3 book (2)

julienrf:
I’ve reviewed the following chapters:
- a first look at types
- control structures
- domain modeling

Here are some simple suggestions, or typos fixes.

Maybe @alvinj you are interested in reviewing?

### DIFF
--- a/_overviews/scala3-book/control-structures.md
+++ b/_overviews/scala3-book/control-structures.md
@@ -144,26 +144,14 @@ scala> for i <- ints do println(i)
 3
 ````
 
-When you need a multiline block of code following the `for` generator, use either of these approaches:
+When you need a multiline block of code following the `for` generator, use the following syntax:
 
 ```scala
-// option 1
 for
   i <- ints
 do
   val x = i * 2
   println(s"i = $i, x = $x")
-
-// option 2
-for (i <- ints)
-  val x = i * 2
-  println(s"i = $i, x = $x")
-
-// option 3
-for (i <- ints) {
-  val x = i * 2
-  println(s"i = $i, x = $x")
-}
 ```
 
 ### Multiple generators
@@ -276,7 +264,7 @@ val list =
   yield
     i * 2
 
-// result: list == Vector(20, 22, 24)
+// list: IndexedSeq[Int] = Vector(20, 22, 24)
 ```
 
 After that `for` expression runs, the variable `list` is a `Vector` that contains the values shown.
@@ -288,7 +276,7 @@ This is how the expression works:
    It multiples it by `2`, then yields the value `22`.
    You can think of these yielded values as accumulating in a temporary holding place.
 3. Finally the loop gets the number `12` from the range, multiplies it by `2`, yielding the number `24`.
-  The loop completes at this point and yields the final result, the `Vector(20,22,24)`.
+  The loop completes at this point and yields the final result, the `Vector(20, 22, 24)`.
 
 {% comment %}
 NOTE: This is a place where it would be great to have a TIP or NOTE block:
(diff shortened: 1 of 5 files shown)


_overviews/scala3-book/control-structures.md
```diff
@@ -144,26 +144,14 @@ scala> for i <- ints do println(i)
...
-// option 2
-for (i <- ints)
-  val x = i * 2
-  println(s"i = $i, x = $x")
-
-// option 3
-for (i <- ints) {
-  val x = i * 2
-  println(s"i = $i, x = $x")
-}
```
julienrf, line 166: I suggest not showing too many options, unless there is a clear rule that says which one to pick when. Here, I would just show the preferred option.

_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -33,27 +33,27 @@ class Movie(var name: String, var director: String, var year: Int)
...
-    this()
+  public Person(String name, String vocation) {
     // assigning the fields
-    name = _name
-    vocation = _vocation
+    this.name = name;
+    this.vocation = vocation;
+  }
+}
 ```
```
julienrf, line 51: The original Scala code was not valid. My commit uses a Java version, which makes sense only to Java users, so I’m not sure it’s a good idea, maybe we should just remove that code snippet?

alvinj:
> Maybe @alvinj you are interested in reviewing?

Hi @julienrf. Yes, I’ll be glad to review these, though it might be Saturday before I can get to them.

alvinj:
> though it might be Saturday before I can get to them.

Sorry, please make that Sunday, June 13th. (I’ll review them then.)

_overviews/scala3-book/control-structures.md
alvinj: Sounds good to me!

_overviews/scala3-book/domain-modeling-oop.md
```diff
@@ -89,12 +89,11 @@ When designing software in Scala, it’s often helpful to only consider using cl
 NOTE: I think “leaves” may technically be the correct word to use, but I prefer “leafs.”
 {% endcomment %}
 
-```text
-traits               T1   T2           ...  T3
-composed traits      S extends T1, T2  ...  S extends T2, T3
-classes              C extends S, T3
-instances            new C
-```
+| Traits          | `T1`, `T2`, `T3`
```
alvinj, line 92: At the moment I can’t run Jekyll on my computer to render this, but if that table syntax works, this is fine by me.

_overviews/scala3-book/domain-modeling-tools.md
alvinj: I’m sorry about that initial Scala code, that definitely wasn’t the right approach.

I can go either way on this. On the one hand, just showing the Scala code shows how concise it is. People who use any other language can make the “conciseness” comparison in their head.

My original logic was (a) I assume that most developers have worked with Java, and (b) I wanted to appeal to Java developers. But again, they can make that comparison in their head. (And if we show Java, why not show Python as well?)

So yes, after thinking about it, I would delete that comparison to Java.

alvinj (changes requested):
All of these changes look great — thanks!

_overviews/scala3-book/domain-modeling-oop.md
julienrf: Sorry, I should have included a screenshot. It looks like this:

Screenshot from 2021-06-14 10-27-37

_overviews/scala3-book/domain-modeling-tools.md
julienrf: OK, thanks for sharing your thoughts. I have removed the Java code snippet.

_overviews/scala3-book/domain-modeling-oop.md
alvinj: No problem on this. I just had a problem with Jekyll and its tools, and I didn’t have time to look into how to fix it.

PR #2078, opened 2021-06-12
Fix why-scala-3.md

ShapelessCat:
The second `LinearSeq` of the clause "traits like `LinearSeq` and `LinearSeq`" should be another trait.


julienrf (approved):

julienrf:
Thank you!

PR #2079, opened 2021-06-14
Suggestions in the Scala 3 book (part 4)

julienrf:
I’ve reviewed the following chapters:
- Scala Collections
- Functional Programming
- Types and the Type System
- Contextual Abstractions

I am kindly requesting a review by @alvinj.

### DIFF
--- a/_overviews/scala3-book/types-type-classes.md
+++ b/_overviews/scala3-book/types-type-classes.md
@@ -23,7 +23,7 @@ In Scala 3, _type classes_ are just _traits_ with one or more type parameters, l
 trait Show[A]:
   def show(a: A): String
 ```
-Instances of `Show` for a particular type `A` witness that `A` we can show an instance of type `A`.
+Instances of `Show` for a particular type `A` witness that we can show (i.e., produce a text representation of) an instance of type `A`.
 For example, let’s look at the following `Show` instance for `Int` values:
 
 ```scala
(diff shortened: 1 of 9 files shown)


_overviews/scala3-book/types-type-classes.md
```diff
@@ -23,7 +23,7 @@ In Scala 3, _type classes_ are just _traits_ with one or more type parameters, l
 trait Show[A]:
   def show(a: A): String
 ```
-Instances of `Show` for a particular type `A` witness that `A` we can show an instance of type `A`.
+Instances of `Show` for a particular type `A` witness that we can show (ie, produce a text representation of) an instance of type `A`.
```
alvinj, line 26: Thanks, this is a good addition. I would just change “ie,” to “i.e.,”.

If you’re interested in another change in this file, I just saw a heading with this text:

#### Automatically passing Type Class Instances

I think that should be:

#### Automatically passing type class instances

I can change this separately if you prefer, but I just noticed it and thought I could/should mention it now.

alvinj (changes requested):
Again, these edits look great!

_overviews/scala3-book/types-type-classes.md
julienrf: Thanks, I’ve applied both suggestions.

PR #2080, opened 2021-06-14
fix: fix syntax issue with arguments

SourceCodeBot:
while reading, this line seems to be wrong to me.
What would you say?


Philippus (approved):

julienrf (approved):
Good catch, thanks!

PR #2083, opened 2021-06-15
Fix typo on line 131 of domain-modeling-oop.md

elibaldwin:
Very simple typo. Was: "Sala 3", should be: "Scala 3".
I've never actually submitted a PR before so hopefully I did it the correct way.


SethTisue:
thank you!

PR #2084, opened 2021-06-18
Create CODEOWNERS

xuwei-k:
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+/_ja @scala/docs-ja


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+/_ja @scala/docs-ja
```
eed3si9n, line 1: Maybe we should add similar ownership for zh and ko?

eed3si9n (approved):

PR #2086, opened 2021-06-19
example has wrong scala version

pjfanning:
I could be wrong but I think this example should have scala version 2.12.3 as opposed to scala 3.


julienrf (approved):
Good catch, thanks!

julienrf:
Do you mind squashing the commits into a single one?

pjfanning:
@julienrf is the adjusted PR ok now?

julienrf:
Perfect, thank you again!

PR #2087, opened 2021-06-20
scala-for-python-devs.md creating multiline string variable

abhi887:
Added an example for creating multiline string variable

### DIFF
--- a/_overviews/scala3-book/scala-for-python-devs.md
+++ b/_overviews/scala3-book/scala-for-python-devs.md
@@ -136,13 +136,19 @@ These examples demonstrate how to create variables in Python and Scala.
     <tr>
       <td class="python-block">
         <code>x = 1
-        <br>x = "Hi"</code>
+        <br>x = "Hi"
+        <br>x = """foo
+        <br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; bar
+        <br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; baz"""</code>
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -136,13 +136,19 @@ These examples demonstrate how to create variables in Python and Scala.
     <tr>
       <td class="python-block">
         <code>x = 1
-        <br>x = "Hi"</code>
+        <br>x = "Hi"
+        <br>x = """foo
```
julienrf, line 140: May I suggest to use a different name than `x` (e.g., `y`)?

abhi887: sounds good 👍🏻

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -186,7 +186,7 @@ These examples demonstrate how to create variables in Python and Scala.
     </tr>
     <tr>
       <td class="scala-block">
-        <code>val movies = Map(
+        <code>val x = Map(
```
julienrf, line 189: Sorry to be picky, but `movies` was a better name than `x`, IMHO. If we change something here, we should rather change the Python code to use `movies` as well.

abhi887: I thought that too, but  `x` was used throughout the documentation so I felt it looked weird.

julienrf (approved):

PR #2088, opened 2021-06-21
Suggestions on the Scala 3 book (part 5)

julienrf:
Last batch of suggestions on the Scala 3 book!

@alvinj if you have some time to review, it would be great!

### DIFF
--- a/_overviews/scala3-book/concurrency.md
+++ b/_overviews/scala3-book/concurrency.md
@@ -55,7 +55,8 @@ You’ll also see examples of methods that are used to handle the value in a fut
 
 > When you think about futures, it’s important to know that they’re intended as a one-shot, “Handle this relatively slow computation on some other thread, and call me back with a result when you’re done” construct.
 > As a point of contrast, [Akka](https://akka.io) actors are intended to run for a long time and respond to many requests during their lifetime.
-> While an actor may live forever, a future is intended to be run only once.
+> While an actor may live forever, a future eventually contains the result
+> of a computation that ran only once.
 
 
 
@@ -77,7 +78,7 @@ Now you’re ready to create a future.
 For this example, first define a long-running, single-threaded algorithm:
 
 ```scala
-def longRunningAlgorithm =
+def longRunningAlgorithm() =
   Thread.sleep(10_000)
   42
 ```
@@ -86,29 +87,29 @@ That fancy algorithm returns the integer value `42` after a ten second delay.
 Now call that algorithm by wrapping it into the `Future` constructor, and assigning the result to a variable:
 
 ```scala
-scala> val f = Future(longRunningAlgorithm)
-f: scala.concurrent.Future[Int] = Future(<not completed>)
+scala> val eventualInt = Future(longRunningAlgorithm())
+eventualInt: scala.concurrent.Future[Int] = Future(<not completed>)
 ```
 
-Right away your future begins running.
-If you immediately check the value of the variable `f`, you see that the future hasn’t completed yet:
+Right away, your computation---the call to `longRunningAlgorithm()`---begins running.
+If you immediately check the value of the variable `eventualInt`, you see that the future hasn’t been completed yet:
 
 ```scala
-scala> f
+scala> eventualInt
 val res1: scala.concurrent.Future[Int] = Future(<not completed>)
 ```
 
-But if you check again after ten seconds, you’ll see that it completes successfully:
+But if you check again after ten seconds, you’ll see that it is completed successfully:
 
 ```scala
-scala> f
+scala> eventualInt
 val res2: scala.concurrent.Future[Int] = Future(Success(42))
 ```
 
 While that’s a relatively simple example, it shows the basic approach: Just construct a new `Future` with your long-running algorithm.
 
 One thing to notice is that the `42` you expected is wrapped in a `Success`, which is further wrapped in a `Future`.
-This is a key concept to understand: the value in a `Future` is always an instance of one of the *scala.util.Try* types: `Success` or `Failure`.
+This is a key concept to understand: the value in a `Future` is always an instance of one of the `scala.util.Try` types: `Success` or `Failure`.
 Therefore, when you work with the result of a future, you use the usual `Try`-handling techniques.
 
 
@@ -118,7 +119,7 @@ Therefore, when you work with the result of a future, you use the usual `Try`-ha
 This is what the result looks like when you call `map` right after creating the variable `f`:
 
 ```scala
-scala> val a = f.map(_ * 2)
+scala> val a = eventualInt.map(_ * 2)
 a: scala.concurrent.Future[Int] = Future(<not completed>)
 ```
 
@@ -139,13 +140,13 @@ In addition to higher-order functions like `map`, you can also use callback meth
 One commonly used callback method is `onComplete`, which takes a *partial function* in which you handle the `Success` and `Failure` cases:
 
 ```scala
-f.onComplete {
+eventualInt.onComplete {
   case Success(value) => println(s"Got the callback, value = $value")
   case Failure(e) => e.printStackTrace
 }
 ```
 
-When you paste that code in the REPL you’ll see the result:
+When you paste that code in the REPL you’ll eventually see the result:
 
 ```scala
 Got the callback, value = 42
@@ -180,10 +181,11 @@ See the [Futures and Promises][futures] page for a discussion of additional meth
 
 ## Running multiple futures and joining their results
 
-To run multiple futures in parallel and join their results when all of the futures complete, use a `for` expression.
+To run multiple computations in parallel and join their results when all of the futures have been completed, use a `for` expression.
+
 The correct approach is:
 
-1. Create the futures
+1. Start the computations that return `Future` results
 2. Merge their results in a `for` expression
 3. Extract the merged result using `onComplete` or a similar technique
 
@@ -191,27 +193,27 @@ The correct approach is:
 ### An example
 
 The three steps of the correct approach are shown in the following example.
-A key is that you first create the futures and then join them in the `for` expression:
+A key is that you first start the computations that return futures, and then join them in the `for` expression:
 
 ```scala
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success}
 
-val startTime = System.currentTimeMillis
-def delta() = System.currentTimeMillis - startTime
+val startTime = System.currentTimeMillis()
+def delta() = System.currentTimeMillis() - startTime
 def sleep(millis: Long) = Thread.sleep(millis)
 
 @main def multipleFutures1 =
 
   println(s"creating the futures:   ${delta()}")
 
-  // (1) create the futures
+  // (1) start the computations that return futures
   val f1 = Future { sleep(800); 1 }   // eventually returns 1
   val f2 = Future { sleep(200); 2 }   // eventually returns 2
   val f3 = Future { sleep(400); 3 }   // eventually returns 3
 
-  // (2) run them simultaneously in a `for` expression
+  // (2) join the futures in a `for` expression
   val result =
     for
       r1 <- f1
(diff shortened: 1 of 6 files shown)


_overviews/scala3-book/concurrency.md
```diff
@@ -55,7 +55,8 @@ You’ll also see examples of methods that are used to handle the value in a fut
 
 > When you think about futures, it’s important to know that they’re intended as a one-shot, “Handle this relatively slow computation on some other thread, and call me back with a result when you’re done” construct.
 > As a point of contrast, [Akka](https://akka.io) actors are intended to run for a long time and respond to many requests during their lifetime.
-> While an actor may live forever, a future is intended to be run only once.
+> While an actor may live forever, a future eventually contains the result
+> of a computation that ran only once.
```
julienrf, line 59: I think it is important to not say things like “we run a `Future`” because that maintains a confusion on the purpose of `Future` (which only holds a future result, unlike a “task”, which is a computation that we can run, cancel, etc.)

_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -122,9 +123,9 @@ This section provides comparisons of features related to OOP-style classes and m
     <tr>
       <td class="java-block">
         <code>class Person {
-          <br>&nbsp; private String firstName;
-          <br>&nbsp; private String lastName;
-          <br>&nbsp; private int age;
+          <br>&nbsp; public String firstName;
+          <br>&nbsp; public String lastName;
+          <br>&nbsp; public int age;
```
julienrf, line 128: This is not idiomatic Java but since there are no getters/setters I think it is better to make them public, to be consistent with the Scala version (or, we could make them private in the Scala version as well).

_overviews/scala3-book/scala-tools.md
```diff
@@ -38,7 +38,20 @@ $ mkdir hello
...
+
+```bash
+$ mkdir project
+```
+
+Create a file named _build.properties_ in the directory `project`, with
+the following content:
+
+```text
+sbt.version=1.5.4
```
julienrf, line 51: I think it’s important to teach people to set the version of sbt, otherwise loading the build is undeterministic.

_overviews/scala3-book/concurrency.md
```diff
@@ -180,7 +181,7 @@ See the [Futures and Promises][futures] page for a discussion of additional meth
 
 ## Running multiple futures and joining their results
 
-To run multiple futures in parallel and join their results when all of the futures complete, use a `for` expression.
+To run multiple computations in parallel and join their results when all of the futures have been completed, use a `for` expression.
```
b-studios, line 184: This sentence was really confusing me. It sounds like it is suggesting:

```
for {
  a <- future { ... }
  b <- future { ... }
  c <- future { ... }
} yield ...
```
but IIUC this would actually run them in sequence, not in parallel, no?

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -192,7 +191,7 @@ Here are a few things to notice in that code:
 
 ## How to use Scala `Option` in Java
 
-When you need to use a Scala `Option` in your Java code, you can convert the `Option` to a Java `Optional` value using the `toJava` method of the Scala _scala.jdk.javaapi.OptionConverters_ object.
```
b-studios, line 195: do we now use ticks everywhere consistently? I know @alvinj was using underscore before, because O'Reilly did that.

_overviews/scala3-book/scala-for-java-devs.md
```diff
@@ -49,6 +49,7 @@ Also at a high level, the differences between Java and Scala are:
 - Scala has a full suite of immutable collections, including `List`, `Vector`, and immutable `Map` and `Set` implementations
 - Everything in Scala is an _expression_: constructs like `if` statements, `for` loops, `match` expressions, and even `try`/`catch` expressions all have return values
 - Scala idioms favor immutability by default: you’re encouraged to use immutable (`final`) variables and immutable collections
+- Idiomatic Scala code does not use `null`, and thus does not suffer from `NullPointerException`
```
b-studios, line 52: Maybe the comma is not necessary

_overviews/scala3-book/concurrency.md
julienrf: You are right, there is a subtlety here, maybe we should add a note about that (the fact that running the computations must be performed _outside_ of the `for` expression to possibly get parallelism)?

_overviews/scala3-book/interacting-with-java.md
julienrf: Ah, I tried to catch that as much as I could, but I might have missed some. I think we should use backticks everywhere for fully qualified names.

_overviews/scala3-book/scala-tools.md
b-studios: Yes, good point!

b-studios (approved):

b-studios:
Thanks @julienrf your changes look good to me. Maybe we can find a good solution to the `for` comprehension question above. Maybe I am just wrong :)

_overviews/scala3-book/concurrency.md
jxnu-liguobin: Consider describing it more clearly and writing a correct parallel example?

Philippus: There is a way to write it inside a for comprehension like this:

```
for
    _ <- Future.unit
    f1 = Future { sleep(800); 1 }   // eventually returns 1
    f2 = Future { sleep(200); 2 }   // eventually returns 2
    f3 = Future { sleep(400); 3 }   // eventually returns 3
    r1 <- f1
    r2 <- f2
    r3 <- f3
  yield
      println(s"in the 'yield': ${delta()}")
      (r1 + r2 + r3)  
```    

I lik this way because it keeps the code blocks together, less change of code parts (1) and (2) to drift apart over time.
Even Viktor uses it. https://viktorklang.com/blog/Futures-in-Scala-protips-2.html :)

julienrf: I have added a commit that clarifies things further and adds a notice about computations run sequentially within a `for` expression.

_overviews/scala3-book/concurrency.md
```diff
@@ -180,38 +181,39 @@ See the [Futures and Promises][futures] page for a discussion of additional meth
...
+1. Run the computations that return `Future` results
 2. Merge their results in a `for` expression
 3. Extract the merged result using `onComplete` or a similar technique
 
 
 ### An example
 
 The three steps of the correct approach are shown in the following example.
-A key is that you first create the futures and then join them in the `for` expression:
+A key is that you first run the computations that return futures, and then join them in the `for` expression:
```
alvinj, line 196: You could maybe say “you first start the computations” instead of “you first run the computations”, but I think either is fine, depending on your preference.

_overviews/scala3-book/concurrency.md
```diff
@@ -180,38 +181,39 @@ See the [Futures and Promises][futures] page for a discussion of additional meth
...
+val startTime = System.currentTimeMillis()
+def delta() = System.currentTimeMillis() - startTime
 def sleep(millis: Long) = Thread.sleep(millis)
 
 @main def multipleFutures1 =
 
   println(s"creating the futures:   ${delta()}")
 
-  // (1) create the futures
+  // (1) run the computations that return futures
```
alvinj, line 211: Again you can say “start” instead of “run”. I think the reason I prefer “start” is that the word “run” to me implies that you let it run completely.

_overviews/scala3-book/scala-for-java-devs.md
alvinj: I think I made a mistake on this. My intent was to show idiomatic Java, including the getters and setters, but forgot them for some reason. My thinking is Java developers will be reading this page, so I want to show correct Java code, i.e., what they’re used to seeing.

So I think I’d prefer to show private fields along with public getters and setters on the Java side. Then add a few words to explain the difference between that Java code and the Scala code, i.e., that Scala does not use `get` and `set` prefixes.

What do you think? (Also, if that makes sense, I can update the file if you prefer.)

_overviews/scala3-book/scala-for-javascript-devs.md
```diff
@@ -441,7 +441,7 @@ In both JavaScript and Scala, functions are objects, so their functionality is s
     </tr>
     <tr>
       <td class="scala-block">
-          <code>// technically this is a “method,” not a function
+          <code>// technically this is a “method”, not a function
```
alvinj, line 444: FWIW, it always looks funky to me, but the O’Reilly standard is to put the comma inside the closing quote. That being said, putting the comma after the closing quote looks better to me(!).

_overviews/scala3-book/scala-tools.md
```diff
@@ -92,31 +119,33 @@ A nice benefit of that is that once you’re comfortable with its structure, it
...
-   |-- scala/
-|-- test/
-   |-- java/
-   |-- resources/
-   |-- scala/
-target/
+```text
+.
+├── build.sbt
+├── project/
```
alvinj, line 125: Should we add _build.properties_ under the _project_ directory here as well?

_overviews/scala3-book/scala-tools.md
```diff
@@ -284,11 +313,18 @@ As with the previous lesson, create an sbt project directory structure for a pro
 $ mkdir HelloScalaTest
 $ cd HelloScalaTest
 $ mkdir -p src/{main,test}/scala
-$ mkdir project target
+$ mkdir project
 ```
 
 
-### Creating the build.sbt file
+### Creating the build.properties and build.sbt files
```
alvinj, line 320: I think I may need to look at this file when my brain is a little fresher, but at the moment it seems like I created a lot of `###` headers under the `## Building Scala projects with sbt` header, and today I find that confusing. I have another project I need to work on now, but if you’d like, I can look at this file again next week and see if it makes sense then, or reorganize it a little.

alvinj (changes requested):

_overviews/scala3-book/scala-tools.md
julienrf: Feel free to reorganize if you think it makes sense :blush: 

IMHO, the structure looks good:
Screenshot from 2021-06-23 09-09-02

_overviews/scala3-book/scala-tools.md
julienrf: Added.

_overviews/scala3-book/concurrency.md
julienrf: Changed for “start”

_overviews/scala3-book/concurrency.md
julienrf: Changed for “start”

_overviews/scala3-book/scala-for-java-devs.md
julienrf: What would be the Scala code?

_overviews/scala3-book/scala-tools.md
alvinj: Okay, thanks. I just created a reminder to look at it next Wednesday.

_overviews/scala3-book/scala-for-javascript-devs.md
Ichoran: Alas, the American "standard," such as it is, is to place commas and periods within quotes of all sorts, including definitional and "scare quotes," even though other punctuation goes outside unless the punctuation is being quoted...isn't this "lovely"?  (Contrast: He said, "yikes!".  I wasn't expecting him to say, "fine"!)

However, as language is a living thing (and the British style is already to put punctuation outside), I always recommend to use the more sensible formatting unless forced to do otherwise, and hope eventually it will grow to become the universal standard.

In this case, though, I wouldn't use quotes at all: `// technically, this is a method, not a function`

_overviews/scala3-book/scala-for-java-devs.md
alvinj: Okay, sorry to flip-flop here, but I worked through the JavaBeans style code, and I don’t like going down that road. Therefore, I think I would keep the current code, and in the *OOP style class, primary constructor* section I would precede it with a little introduction like this:

“Scala doesn’t follow the JavaBeans standard, so instead of showing Java code written in the JavaBeans style, here we show Java code that is equivalent to the Scala code that follows it.”

Does something like that sound okay?

_overviews/scala3-book/scala-for-javascript-devs.md
alvinj: I agree, that’s a nice change. (Thanks!)

julienrf:
```suggestion
          <code>// technically this is a method, not a function
```

_overviews/scala3-book/scala-for-java-devs.md
julienrf: Done.

alvinj:
@julienrf This looks much better to me this week. :) One question: What do you think about numbering the steps, like steps in a recipe? With that, the headers will look like this:

```
1) Create the project directory structure
2) Create the build.properties and build.sbt files
3) Create a Scala source code file
4) Create your first ScalaTest test
```

Personally I prefer this approach whenever there is a step-by-step process, but either way is fine.

julienrf:
I agree, that’s a good idea!

PR #2091, opened 2021-06-23
zh-cn/trait.md

zhangchuanben:
fix function result error


jxnu-liguobin:
I don't think it's wrong. `t` is not a reference type assignment here.

image

PR #2096, opened 2021-06-24
Old style Scala 2 If-else expression

Altair-Bueno:
`if (<condition>) <doThis> else <doThat>` syntax is from Scala 2. Use `if <condition> then <doThis> else <doThat>` instead or simplify it by removing the if-else expression. If-else is unnecessary so i removed it. I already tested this code using REPL. It does work as intended

```
$ scala3-repl
scala> def isNullOrEmpty(s: String): Boolean = s==null || s.trim.equals("")
def isNullOrEmpty(s: String): Boolean

scala> isNullOrEmpty (null)
val res0: Boolean = true

scala> isNullOrEmpty ("")                                                                            
val res3: Boolean = true

scala> isNullOrEmpty ("hello")                                                                     
val res1: Boolean = false
```

### DIFF
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -219,7 +219,7 @@ Here’s an example of a “string utilities” object that contains a set of me
 object StringUtils:
   def truncate(s: String, length: Int): String = s.take(length)
   def containsWhitespace(s: String): Boolean = s.matches(".*\\s.*")
-  def isNullOrEmpty(s: String): Boolean = s==null || s.trim.isEmpty
+  def isNullOrEmpty(s: String): Boolean = s == null || s.trim.isEmpty
 ```
 
 We can use the object as follows:
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -219,8 +219,7 @@ Here’s an example of a “string utilities” object that contains a set of me
 object StringUtils:
   def truncate(s: String, length: Int): String = s.take(length)
   def containsWhitespace(s: String): Boolean = s.matches(".*\\s.*")
-  def isNullOrEmpty(s: String): Boolean =
-    if s == null || s.trim.equals("") then true else false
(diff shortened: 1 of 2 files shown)


SethTisue:
nice catch.

also, while we're changing this example, `s.trim.equals("")` isn't idiomatic Scala. `s.trim == ""` would be a little better and `s.trim.isEmpty` would be better still

Altair-Bueno:
@SethTisue i changed `s.trim.equals(")` to s.`trim.isEmpty` as you suggested. Should be good to go

Altair-Bueno:
I found another `isNullOrEmpty`. This time it was using Scala's 3 syntax instead of Scala 2, but i changed it to match 'A taste of Scala' section

_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -219,8 +219,7 @@ Here’s an example of a “string utilities” object that contains a set of me
 object StringUtils:
   def truncate(s: String, length: Int): String = s.take(length)
   def containsWhitespace(s: String): Boolean = s.matches(".*\\s.*")
-  def isNullOrEmpty(s: String): Boolean =
-    if s == null || s.trim.equals("") then true else false
+  def isNullOrEmpty(s: String): Boolean = s==null || s.trim.isEmpty
```
Philippus, line 222:
```suggestion
  def isNullOrEmpty(s: String): Boolean = s == null || s.trim.isEmpty
```

_overviews/scala3-book/taste-objects.md
```diff
@@ -27,8 +27,7 @@ For example, this `StringUtils` object contains a small collection of string-rel
 
 ```scala
 object StringUtils:
-  def isNullOrEmpty(s: String): Boolean =
-    if (s==null || s.trim.equals("")) true else false
+  def isNullOrEmpty(s: String): Boolean = s==null || s.trim.isEmpty
```
Philippus, line 30:
```suggestion
  def isNullOrEmpty(s: String): Boolean = s == null || s.trim.isEmpty
```

Philippus (approved):

Philippus:
Thanks!

PR #2099, opened 2021-06-26
Fix a typo in Domain Modeling / Tools

genghis-yang:
I believe `have also have` is not grammatical. Should be `also have`


julienrf (approved):

julienrf:
Good catch, thank you!

PR #2100, opened 2021-06-26
Fix a typo in Scala 3 Book / Domain Modeling / FP Modeling

genghis-yang:
Duplicated `us` :hand_over_mouth:


Philippus:
I think the original sentence was correct. There's only one `us`, and one `use`.

genghis-yang:
I believe `Let's us use ...` = `Let us us use ...`
Have I misunderstood anything? 🤔

Philippus:
Ah yes, you're right, I missed the apostrophe. I think that should be removed so it becomes `It ... lets us use`.

Philippus (approved):

Philippus:
The build failure is unrelated to this PR.

Thanks @genghis-yang !

PR #2102, opened 2021-06-26
How to setup a virtual environment in Scala?

mariogalic:
To help with onboarding Python developers into Scala ecosystem.

### DIFF
--- a/_overviews/scala3-book/scala-for-python-devs.md
+++ b/_overviews/scala3-book/scala-for-python-devs.md
@@ -1287,6 +1287,30 @@ Follow the links below for more details:
 - Infix methods
 - Macros and metaprogramming
 
+## Scala and virtual environments
+
+In Scala, there is no need to explicitly set up the equivalent of a Python virtual environment. By default, Scala build tools manage project dependencies such that users do not have to think about manual package installation. For example, using the `sbt` build tool, we specify dependencies inside `build.sbt` file under `libraryDependencies` setting, then executing
+
+```
+cd myapp
+sbt compile
+```
+
+automatically resolves all dependencies for that particular project. The location of downloaded dependencies is largely an implementation detail of the build tool, and users do not have to interact with these downloaded dependencies directly. For example, if we delete the whole sbt dependencies cache, on the next compilation of the project, sbt simply resolves and downloads all the required dependencies again, automatically.
+
+This differs from Python, were by default dependencies are installed in system-wide or user-wide directories, so to obtain an isolated environment on a per-project basis one has to create a corresponding virtual environment. For example, using the `venv` module, we might create one for a particular project like so
+
+```
+cd myapp
+python3 -m venv myapp-env
+source myapp-env/bin/activate
+pip install -r requirements.txt
+```
+
+This installs all the dependencies under the project's `myapp/myapp-env` directory and alters the shell environmental variable `PATH` to look up dependencies from `myapp-env`.
+None of this manual process is necessary in Scala.
+
+
 [collections-classes]: {% link _overviews/scala3-book/collections-classes.md %}
 [concurrency]: {% link _overviews/scala3-book/concurrency.md %}
 [contextual]: {% link _overviews/scala3-book/ca-contextual-abstractions-intro.md %}


Jasper-M (commented):
I think you're getting a bit caught up in explaining python. I would leave that off, it's not a python tutorial. I'm also not sure I would say that with sbt dependencies are "installed". The fact that ivy or coursier keep local caches is pretty much orthogonal to how dependency management works, or how it is different from the python situation.

mariogalic:
Thanks @Jasper-M for feedback! I am not sure how to proceed without being able to compare it to Python. I tried to emulate side-by-side comparison approach taken in the rest of "Scala for Python Developers" page. Shall I rename "installed" to "downloaded"? IMO the most visible differences for newcomer dependencies are not duplicated across projects and environmental variables are not modified. Why is it orthogonal?

Jasper-M:
> I am not sure how to proceed without being able to compare it to Python.

I would start with a short description of how python and python modules are by default installed globally for the whole system, and how virtual environments are a solution to that. Then how Scala fundamentally doesn't have that issue because of how build tools and dependency management for JVM works. That's also a bit what I meant with the orthogonal comment. But I think you're probably right that it's definitely worth mentioning in this context as it's a big difference from python.

But that's all my opinion. Someone else may feel differently.

LPTK:
> Shall I rename "installed" to "downloaded"? IMO the most visible differences for newcomer are that dependencies are not duplicated across projects and environmental variables are not modified. Why is it orthogonal?

It's completely different. You can `rm -rf` SBT's cache, and the only difference you'll experience is that first-time build times after that will take longer as the artifacts will need redownloading. So SBT's cache is an implementation detail, which does not affect the semantics of the build (hence "orthogonal").

Compare that to _installing_ packages, which destructively modifies the global (or virtual) state of the build tool. Imagine what would happen if you started deleting files you have installed as part of a Python package.

I think this difference is fundamental and super important, even for beginners.

_overviews/scala3-book/scala-for-python-devs.md
```diff
@@ -1291,6 +1291,29 @@ Follow the links below for more details:
 - Infix methods
 - Macros and metaprogramming
 
+## How to setup a virtual environnement in Scala?
+
```
martijnhoekstra, line 1295: I'd love if there were a "most important takeaway" right on top before you get to the introduction about what python does.

In scala, there is no need for virtual environments.

Then after that comes a description of the python situation and how building with sbt differs from that. A reader could stop reading after that first sentence, or read more if they want to convince themselves it's true.

SethTisue:
@LPTK re-review?

LPTK (changes requested):
Ok, I read through it and found that the writing could be improved in many places. Please apply my patches, as there are many minor grammatical mistakes in the current version.

mariogalic:
I've applied the patches. Please feel free to make any direct changes. As a side note build is failing because GraalVM site seems to have removed Scala link so it results in 404.

LPTK:
I don't want to be a grammar nazi, but the writing still feels broken in many places. Just looking at the first sentence:

> In Scala there is no need to explicitly setup an equivalent of Python's virtual environment.

"setup" is not a verb; "set up" is. And "Python" does not have one virtual environment, so "Python's virtual environment" is the wrong formulation. That's why I proposed "set up the equivalent of a Python virtual environment".

I'm going over the rest, but you get the idea. I guess it's not that important though, so feel free to merge.

mariogalic:
I must have messed up interactive rebase because I thought I applied all your patches. I will fix it.

mariogalic:
@LPTK Your patches seem to be applied correctly. Maybe it was showing outdated github page for you? For example first sentence does have "set up" instead of "setup".

LPTK:
@mariogalic ok cool, I guess I was reading the wrong diff, thanks 👍

LPTK (approved):

mariogalic:
Any ideas for improving the title? Perhaps

- Virtual Environments for Scala
- Creating virtual environments for Scala
- Set up virtual environments for Scala
- Scala virtual environments

LPTK:
@mariogalic "Scala and Virtual Environments"?

SethTisue:
is this ready for merge?

mariogalic:
@SethTisue Is the question for me? I think I addressed all the comments. Do let me know if anything else is needed.

mariogalic:
Anyone interested in merging this? I do not have write access.

LPTK:
Me neither.

PR #2117, opened 2021-07-10
Update syntax-summary.md

jxnu-liguobin:


_overviews/quasiquotes/syntax-summary.md
```diff
@@ -120,7 +120,7 @@ permalink: /overviews/quasiquotes/:title.html
                               | Quasiquote                                                                                                                  | Type
 ------------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------
  [Val][401]                   | `q"$mods val $tname: $tpt = $expr"` or `q"$mods val $pat = $expr"`                                                          | ValDef
- [Var][401]                   | `q"$mods var $tname: $tpt = $expr"` or `q"$mods val $pat = $expr"`                                                          | ValDef
+ [Var][401]                   | `q"$mods var $tname: $tpt = $expr"` or `q"$mods var $pat = $expr"`                                                          | ValDef
```
Philippus, line 123: I'm guessing that the type should also become `VarDef`.
```suggestion
 [Var][401]                   | `q"$mods var $tname: $tpt = $expr"` or `q"$mods var $pat = $expr"`                                                          | VarDef
```

jxnu-liguobin: No, they are distinguished by `mods`, not exists `VarDef`.

This is doc:
```scala

  /** Broadly speaking, a value definition.  All these are encoded as ValDefs:
   *
   *   - immutable values, e.g. "val x"
   *   - mutable values, e.g. "var x" - the MUTABLE flag set in mods
   *   - lazy values, e.g. "lazy val x" - the LAZY flag set in mods
   *   - method parameters, see vparamss in [[scala.reflect.api.Trees#DefDef]] - the PARAM flag is set in mods
   *   - explicit self-types, e.g. class A { self: Bar => }
   *  @group Trees
   *  @template
   */
  type ValDef >: Null <: ValDefApi with ValOrDefDef
```

PR #2135, opened 2021-07-20
Fix a typo in a Java code example

Nicolapps:
This pull request fixes what looks like a typo in the [*Scala for Java Developers*](https://docs.scala-lang.org/scala3/book/scala-for-java-devs.html#auxiliary-constructors) page.

The Java code example isn’t currently compiling, because two constructors of `Person` have the same signature.


SethTisue:
review by @alvinj?

alvinj (approved):
Yes, this change is correct — thank you!

alvinj:
Oops, I also see that I have two issues in the first Java `Person` example, but I can fix those in a separate PR. (`int age` should be on a separate line, and `override` should be `public`.)

PR #2136, opened 2021-07-21
API documentation is now hosted on scala-lang.org/api/3.x/

julienrf:
@romanowski I generated the API documentation by removing the `-siteroot` argument, and I uploaded the content to https://scala-lang.org/api/3.0.0/, https://scala-lang.org/api/3.0.1/, and https://scala-lang.org/api/3.x/ (which is a symbolic link to the latest stable release). One thing that annoys me is that the URLs now contain “api” twice, e.g.: https://scala-lang.org/api/3.x/api/scala/util/CommandLineParser$.html

It is still time to change that, I believe. How would you recommend fixing this issue?

### DIFF
--- a/_overviews/scala3-macros/tutorial/reflection.md
+++ b/_overviews/scala3-macros/tutorial/reflection.md
@@ -50,9 +50,9 @@ the types in the API. For each type `Foo` in the tree:
  - the trait `FooModule` contains the static methods available on the object `Foo`. Most notably, constructors (`apply/copy`) and the `unapply` method which provides the extractor(s) required for pattern matching.
  - For all types `Upper` such that `Foo <: Upper`, the methods defined in `UpperMethods` are available on `Foo` as well.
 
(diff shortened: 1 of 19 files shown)


_overviews/scala3-macros/tutorial/reflection.md
```diff
@@ -230,6 +230,6 @@ We can make this printer the default if needed
 *Coming soon*
 
 [tasty inspection]: {{ site.scala3ref }}/metaprogramming/tasty-inspect.html
-[reflection doc]: https://dotty.epfl.ch/api/scala/quoted/Quotes$reflectModule.html?query=trait%20reflectModule
+[reflection doc]: https://scala-lang.org/api/3.x/api/scala/quoted/Quotes$reflectModule.html
```
julienrf, line 233: I’m not sure what the `query` parameter was supposed to do. It seems to do nothing on dotty.epfl.ch

romanowski: Scaladoc does not use query paramters at this moment.

romanowski:
I've got a fix for you in #13130

julienrf:
With the (huge) help of @BarkingBad and @romanowski, the issue mentioned above is fixed!

PR #2138, opened 2021-07-24
add transparent modifier to defaultValue

tkawachi:
`val dInt: Some[Int] = defaultValue[Int]` etc. fail to compile  without transparent.


julienrf (approved):
good catch, thanks!

PR #2139, opened 2021-07-24
Fix compile

tkawachi:
In scala-library 2.13.6, I confirmed that there are two argument lists, but in https://dotty.epfl.ch/api/scala/collection/immutable/TreeSet.html, it looks like there is only one argument list for the constructor.

I tried with `scalaVersion := "3.0.1"` then scala-libary 2.13.6 is used. I don't know where there's the updated stdlib.

### DIFF
--- a/_scala3-reference/metaprogramming/compiletime-ops.md
+++ b/_scala3-reference/metaprogramming/compiletime-ops.md
@@ -215,7 +215,7 @@ would use it as follows:
 import scala.compiletime.summonFrom
 
 inline def setFor[T]: Set[T] = summonFrom {
-  case ord: Ordering[T] => new TreeSet[T](using ord)
+  case ord: Ordering[T] => new TreeSet[T]()(using ord)
   case _                => new HashSet[T]
 }
 ```


SethTisue:
To aid migration, Scala 3.0.x uses the 2.13.6 standard library.

Perhaps it's a Scaladoc bug?

tkawachi:
Apparently so. It seems that empty argument lists, regardless of whether they are constructors or functions, are accidentally disappearing from scaladoc.
The following code confirms this.

```
class C1()(implicit i: Int)
class C2()(using i: Int)

class C {
    def f1()(implicit i: Int) = i
    def f2()(using i: Int) = i

    def f3(s: String)(implicit i: Int) = i
    def f4(s: String)(using i: Int) = i

    def f5()()(using i: Int) = i
}
```

The HTML generated by scala 3.0.1 looks like:

image

image

julienrf:
Thanks for reporting! Yes, that seems to be an issue in Scaladoc, so I’m closing the issue here.

julienrf:
Sorry, I didn’t see that there was an issue in the documentation as well… Thanks for the fix!

PR #2140, opened 2021-07-24
style guide: discourage infix

SethTisue:
plus minor assorted copyedits

### DIFF
--- a/_style/declarations.md
+++ b/_style/declarations.md
@@ -13,7 +13,7 @@ next-page: control-structures
 
 ## Classes
 
-Class/Object/Trait constructors should be declared all on one line,
+Class, object, and trait constructors should be declared all on one line,
 unless the line becomes "too long" (about 100 characters). In that case,
 put each constructor argument on its own line with
 [trailing commas](https://docs.scala-lang.org/sips/trailing-commas.html#motivation):
@@ -102,7 +102,7 @@ Local methods or private methods may omit their return type:
 
 #### Procedure Syntax
 
-Avoid the procedure syntax, as it tends to be confusing for very little gain in brevity.
+Avoid the (now deprecated) procedure syntax, as it tends to be confusing for very little gain in brevity.
 
     // don't do this
     def printBar(bar: Baz) {
@@ -160,7 +160,7 @@ or is of a non-functional nature (some mutable state, local or
 otherwise), the body must be enclosed in braces:
 
     def sum(ls: List[String]): Int = {
-      val ints = ls map (_.toInt)
+      val ints = ls.map(_.toInt)
       ints.foldLeft(0)(_ + _)
     }
 
@@ -197,7 +197,8 @@ There are three main reasons you should do this:
     Multiple parameter lists allow you to create your own "control
     structures":
 
-        def unless(exp: Boolean)(code: => Unit): Unit = if (!exp) code
+        def unless(exp: Boolean)(code: => Unit): Unit =
+          if (!exp) code
         unless(x < 5) {
           println("x was not less than five")
         }
@@ -223,7 +224,7 @@ There are three main reasons you should do this:
         List("").foldLeft(0, (b: Int, a: String) => b + a.length)
         List("").foldLeft[Int](0, _ + _.length)
 
-For complex DSLs, or with type-names that are long, it can be difficult
+For complex DSLs, or with type names that are long, it can be difficult
 to fit the entire signature on one line. For those cases there are several
 different styles in use:
 
--- a/_style/method-invocation.md
+++ b/_style/method-invocation.md
@@ -104,32 +104,16 @@ preference to their better-known names, `foldLeft` and `foldRight`.
 ### Higher-Order Functions
 
 Invoking higher-order functions may use parens or braces, but in
-either case, use dot notation:
+either case, use dot notation and omit any space after the method name:
 
     names.map{ _.toUpperCase }
 
-Not infix:
+These are not recommended:
 
     // wrong!
     names map { _.toUpperCase }
-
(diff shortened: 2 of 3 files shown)


SethTisue:
review by @bjornregnell ?

_style/method-invocation.md
```diff
@@ -92,25 +92,34 @@ gray area is short, operator-like methods like `max`, especially if commutative:
...
-invoke such methods in the following way:
+Invoking higher-order functions may use parens or braces, but in
+either case, use dot notation:
+
+    names.map{ _.toUpperCase }
+
+Not infix:
+
+    // wrong!
+    names map { _.toUpperCase }
```
smarter, line 114: Just below this you have:
```scala
    // right!
    names map { _.toUpperCase } filter { _.length > 5 }
```

At this point I suggest deleting the whole section since it confuses rather than enlightens.

SethTisue: Oops, good catch! I've pushed a fix.

I don't know about deleting the whole section, but my fix does shorten it considerably.

_style/method-invocation.md
```diff
@@ -92,35 +92,28 @@ gray area is short, operator-like methods like `max`, especially if commutative:
...
 
-As noted, methods which take functions as parameters (such as `map` or
-`foreach`) should be invoked using infix notation. It is also *possible* to
-invoke such methods in the following way:
+Invoking higher-order functions may use parens or braces, but in
+either case, use dot notation and omit any space after the method name:
 
-    // wrong!
-    names.map { _.toUpperCase }
+    names.map{ _.toUpperCase }
```
smarter, line 109: If this is supposed to be a "good" example then it should be `names.map(...)`, braces should only be used for multiline expressions.

smarter: also braces aren't even needed for multi-line expressions when the method takes a lambda in scala 3:
```scala
xs.map(x =>
  val a = x
  a + 1
)
```
because `=>` introduces an indentation region.

SethTisue: further adjustment pushed

Philippus (approved):

_style/declarations.md
```diff
@@ -160,7 +160,7 @@ or is of a non-functional nature (some mutable state, local or
 otherwise), the body must be enclosed in braces:
 
     def sum(ls: List[String]): Int = {
```
bjornregnell, line 162: The word "must" is perhaps too strong here without mentioning that it is a "must" in Scala 2 only... If there is no plan to make a specific style guide for Scala 3, perhaps it is best to include both Scala 2 and Scala 3 variations in style in the same guide; after all Scala 2 code will be around for almost eternity. I would thus suggest to include something about braces being optional in Scala 3 here.

_style/method-invocation.md
```diff
@@ -51,7 +51,7 @@ acceptable to omit parentheses when calling `queue.size`, but not when
 calling `println()`. This convention mirrors the method declaration
 convention given above.
 
-Religiously observing this convention will *dramatically* improve code
+Religiously observing this convention will dramatically improve code
```
bjornregnell, line 54: The word "religiously" seems religiously religious IMHO and does not have so much to do with earthly coding standards. I suggest using e.g. "Strictly" instead of "Religiously". I like the removal of italics. That makes the text feel less dogmatic and more factual. I'd even consider cancel "dramatically", as this is kind of hand-waving.

_style/method-invocation.md
```diff
@@ -92,35 +92,28 @@ gray area is short, operator-like methods like `max`, especially if commutative:
     // fairly common
     a max b
 
-Symbolic methods which take more than one parameter (they do exist!)
-may still be invoked using infix notation, delimited by spaces:
+Symbolic methods which take more than one parameter are discouraged.
```
bjornregnell, line 95: Good to remove exclamations in parenthesis - they are paradoxical (shouting with a  hushed voice!).

_style/method-invocation.md
```diff
@@ -92,35 +92,28 @@ gray area is short, operator-like methods like `max`, especially if commutative:
...
-invoke such methods in the following way:
+Invoking higher-order functions may use parens or braces, but in
+either case, use dot notation and omit any space after the method name:
 
-    // wrong!
-    names.map { _.toUpperCase }
-
-This style is *not* the accepted standard! The reason to avoid this style is for
-situations where more than one invocation must be chained together:
+    names.map(_.toUpperCase)
```
bjornregnell, line 109: Good to remove this as this was the most problematic part. Suggestions:
* Scala 3 is more liberal with allowing parenthesis where scala 2 requires braces - perhaps this should be mentioned here?
* In places where there are multiple lines in the lambda, perhaps we want to include an example in the style guide of how lines are broken after the rocket `=>` etc., as people reading this standard might wonder how that could/should look.

bjornregnell (changes requested):

bjornregnell:
great edit! See my comments inline. I clicked "request changes" but I could just as well have clicked "Approve" as well as my suggestions are not religious :)

_style/declarations.md
SethTisue: I'm considering Scala 3 out of scope for this PR, but I do hope somebody will tackle it in a future PR.

_style/method-invocation.md
SethTisue: I will omit both "religiously" and "dramatically"

_style/method-invocation.md
SethTisue: Yeah, but I'm going to consider it out of scope for this PR.

SethTisue:
Thanks Bjorn. I took one of your suggestions — the others are good as well but remain as future work for somebody.

PR #2144, opened 2021-07-27
update JDK compat page (add 18, other tweaks)

SethTisue:
I removed some of the information about JDK 9, 10, 12, 13, 14, and 15 as it seems like noise at this point. I kept 16 for now.


SethTisue:
@lrytz I went ahead and merged this, but you might still want to eyeball it

PR #2145, opened 2021-07-29
Nit: Change "Writer" to "Printer" in contravariance example 

mikkelmilo:
The contravariance example uses the `Printer` class, so the text should use this name instead of `Writer` to avoid confusion. Unless I am missing something?


SethTisue:
good catch, thanks!

PR #2148, opened 2021-07-31
add lazy for elemInstances in derivation.md

tkawachi:
I don't know why, but I seem to get a stack overflow error at compile time if I don't use lazy val.
In the code at the bottom of the page, it is lazy val.

If the elems argument of eqSum() is not a call-by-name, it seems to stack overflow.

tested with scala 3.0.1

### DIFF
--- a/_scala3-reference/contextual/derivation.md
+++ b/_scala3-reference/contextual/derivation.md
@@ -218,7 +218,7 @@ instance for the appropriate ADT subtype using the auxiliary method `check` (4).
 ```scala
 import scala.deriving.Mirror
 
-def eqSum[T](s: Mirror.SumOf[T], elems: List[Eq[_]]): Eq[T] =
+def eqSum[T](s: Mirror.SumOf[T], elems: => List[Eq[_]]): Eq[T] =
   new Eq[T]:
     def eqv(x: T, y: T): Boolean =
       val ordx = s.ordinal(x)                            // (3)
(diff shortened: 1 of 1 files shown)


_scala3-reference/contextual/derivation.md
```diff
@@ -218,7 +218,7 @@ instance for the appropriate ADT subtype using the auxiliary method `check` (4).
 ```scala
 import scala.deriving.Mirror
 
-def eqSum[T](s: Mirror.SumOf[T], elems: List[Eq[_]]): Eq[T] =
+def eqSum[T](s: Mirror.SumOf[T], elems: => List[Eq[_]]): Eq[T] =
```
nicolasstucki, line 221: This does not look like a good idea, each time we would do `check(elems(ordx))(x, y)` we would need to recompute the full list.

_scala3-reference/contextual/derivation.md
```diff
@@ -184,7 +184,7 @@ a `Mirror[T]`. Here is a possible implementation,
 import scala.deriving.Mirror
 
 inline given derived[T](using m: Mirror.Of[T]): Eq[T] =
-  val elemInstances = summonAll[m.MirroredElemTypes]           // (1)
+  lazy val elemInstances = summonAll[m.MirroredElemTypes]           // (1)
```
nicolasstucki, line 187: We need `elemInstances` in each branch exactly one, this would add unnecessary overhead?

nicolasstucki (commented):
If you get a stack overflow at compile-time, it might be a compiler bug. Could you report an issue in https://github.com/lampepfl/dotty/issues with the code that causes the stack overflow?

tkawachi:
I tried to reproduce the problem to report it, but was unable to do so. It's strange.

I'm going to close this PR.

The summary code at the bottom of the page uses lazy val and call-by-name arguments, is that OK?

PR #2151, opened 2021-08-08
Fix a typo in FP Modeling

Nicolapps:
This PR fixes a typo in the [*FP Modeling*](https://docs.scala-lang.org/scala3/book/domain-modeling-fp.html#modeling-the-operations) page. The current version doesn’t seem to make sense, since `toppingPrice` is a method and not an enumeration.

The sentence is reworded in a way that is coherent with the previous paragraph (*“since `Pizza` is a case class”*).


Philippus (approved):

joroKr21 (approved):
Thank you 👍

PR #2152, opened 2021-08-08
Fix a syntax error

Nicolapps:
This PR removes an extraneous space character in the [*Union Types*](https://docs.scala-lang.org/scala3/book/types-union.html) page of the Scala 3 Book. Currently, the code doesn’t compile, but not for the reason mentioned in the paragraph above the code block.


Philippus (approved):

joroKr21 (approved):
Thank you 👍

PR #2159, opened 2021-08-18
Clearer grammar when talking about type variance

martinjaime:


julienrf (approved):
That’s better indeed, thanks @martinjaime!

PR #2161, opened 2021-08-21
Fix deprecation warning in scala-for-java-programmers.md

rhysd:
Today I was learning Scala with the text and found `scalac` reported deprecation warning:

```
generics.scala:3: warning: procedure syntax is deprecated: instead, add `: Unit =` to explicitly declare `set`'s return type
  def set(value: T) { contents = value }
                    ^
1 warning
```

This PR fixes the warning.


julienrf (approved):
Thank you @rhysd!

PR #2166, opened 2021-08-29
Update matchable.md

bblfish:
The type has to of the method has to be specified or else the return types will just be a Double in both cases and the tests will pass. 
Strict Equality has to be set for the compiler to fail.


julienrf (approved):
Thank you @bblfish!

PR #2168, opened 2021-08-30
Update sample code for classes

calindragomir:
I think you might have a slight syntax error in the example.

### DIFF
--- a/_style/declarations.md
+++ b/_style/declarations.md
@@ -28,7 +28,7 @@ put each constructor argument on its own line with
       birthdate: Date,
       astrologicalSign: String,
       shoeSize: Int,
-      favoriteColor: java.awt.Color,
+      favoriteColor: java.awt.Color
     ) {
       def firstMethod: Foo = …
     }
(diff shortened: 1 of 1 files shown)


calindragomir:
Or is this intentional , based on https://docs.scala-lang.org/sips/trailing-commas.html ?
If so, I apologize for the noise.

julienrf:
Yes, the code shown on the page is valid with Scala 2.12 or 2.13 (I don’t remember which one). I am not sure if we have a recommendation regarding the usage of trailing commas…

calindragomir:
Ok, I just thought it looked a bit odd, even though it's valid.
I'm guessing the preferred choice for coders would be to omit the trailing comma, even if it works with it. And for that reason the sample code would have the clean version. But , it's just a guess 😄 .

Thanks for the reply.

SethTisue:
Closing, but I think a new PR addressing the question about trailing commas would be welcome. I don't think there is a consensus on whether to use them in cases like this, so I don't think we should make a clear recommendation one way or the other; but it would be appropriate to note that the choice was conscious here but that tastes may differ.

PR #2169, opened 2021-08-31
Tutorials as a new category on the overviews page

makingthematrix:
Currently some of the learning materials are accessed through the "overviews" page while others are under the "Learn" meanu in the header. The header is not intuitive. It's not visible on the main page, so the user needs to click on the overviews to be able to access the materials under the header, which defeats the purpose of having them there if there ever was one. (I suspect the current situation is a side-effect of some other changes, not a conscious decision). On top of that, "Scala with Maven" and "Binary Compatibility for Library Authors" are nowhere to be found.

My proposal is to remove the "Learn" menu from the header and instead to have a new category in the "overviews" page, called "Tutorials", directly under "Language", so it's pretty high on the webpage and easy to find. I discussed it on the Scala discord and I got an impression that even though it's not a perfect solution, it's at least a step in the right direction.

I moved all the links from the "Learn" menu to that new category and added short descriptions and authors' names in places where I was pretty sure what to do. Please review also from this angle: maybe there should more names under the author's tag or it's better to remove them completely, and maybe someone could update the descriptions.

### DIFF
--- a/_data/doc-nav-header.yml
+++ b/_data/doc-nav-header.yml
@@ -5,21 +5,6 @@
       url: https://www.scala-lang.org/api/current/
     - title: All Versions
       url: "/api/all.html"
-- title: Learn
-  url: "#"
-  submenu:
-    - title: Getting Started
-      url: "/getting-started/index.html"
-    - title: Tour of Scala
-      url: "/tour/tour-of-scala.html"
-    - title: Scala Book
-      url: "/overviews/scala-book/introduction.html"
-    - title: Scala for Java Programmers
-      url: "/tutorials/scala-for-java-programmers.html"
-    - title: Scala on Android
-      url: "/tutorials/scala-on-android.html"
-    - title: Online Resources
-      url: "/learn.html"
 - title: Reference
   url: "#"
   submenu:
--- a/_data/overviews.yml
+++ b/_data/overviews.yml
@@ -78,6 +78,48 @@
       icon: diamond
       url: "core/value-classes.html"
 
+- category: Tutorials
+  description: "Guides and overviews covering various topics, from basics of the Scala language to usage in non-standard projects."
+  overviews:
+    - title: Getting Started
+      by:  Meriam Lachkar
+      icon: space-shuttle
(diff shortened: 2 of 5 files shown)


julienrf:
I think the main issue is that the “Learn” menu is not anymore visible on the front page (you have to click on any of the Scala 2 pages to see it). This is because we now have two different navigation menus, depending on the Scala version.

Although I agree that the visibility of the tutorials is terrible, I am not convinced that we should put more things into the “overviews”. I don’t have a better solution to propose though, but let me think about it…

makingthematrix:
@julienrf : I'm not 100% sure about this too, but I decided that it'd be better if I come up with something and then maybe we can improve it. Otherwise probably nobody would ever touch it :)

Anyway, at this moment I already have Jekyll installed on my laptop, and I looked a bit through the code, so if we discuss it and come up with ideas how to make those tutorials more visible, I can do it.

SethTisue:
In general I welcome seeing the site structure get flatter. I find it a bit annoying to remember which page or menu something is on, and annoying to have to view multiple pages and menus to see a list of all the guides/tutorials/etc that are available. As a user of the site, I'd rather just run my eyes over a big list of what's available. Eyes are good at quickly picking out relevant things, fingers less so.

makingthematrix:
@julienrf : Any update on this?

julienrf:
@makingthematrix Thank you for pinging me! I believe we should give tutorials more visibility on the front page rather than making them visible from within the Guides page. I mentioned the issue [here](https://contributors.scala-lang.org/t/scala-documentation-website-roadmap/5173/12), with ideas of possible approaches to fix it.

makingthematrix:
@julienrf : Alright, so it looks like this PR is not going to make it. But that's ok. I'm just interested in improving the website. And I don't want this topic to just die out, so if that Courier issue you mentioned is finally resolved, and you need help with implementing the proposed changes, please let me know. I will do my best.

julienrf:
@makingthematrix I have submitted #2206 which addresses your concern I believe, what do you think?

PR #2170, opened 2021-08-31
Fixed code example for derivation with shapeless 3

i-surkov:
While running code examples form the derivation page, I noticed that the extension method `derive` in shapeless 3 has a different order of arguments. Here's it's current signature (version `3.0.2`):
```scala
extension [F[_], T](gen: Generic[T])
  inline def derive(f: => (ProductGeneric[T] & gen.type) ?=> F[T], g: => (CoproductGeneric[T] & gen.type) ?=> F[T]): F[T] =
    inline gen match {
      case p: ProductGeneric[T]   => f(using p.asInstanceOf)
      case c: CoproductGeneric[T] => g(using c.asInstanceOf)
    }
```


julienrf (approved):
Thank you!

PR #2172, opened 2021-08-31
Simplify the example used for case objects.

vlulla:
Also, fixed a small typo.

### DIFF
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -685,14 +685,14 @@ case class DecreaseVolume(amount: Int) extends Message
 case object StopPlaying extends Message
 ```
 
-Then in other parts of your code you can write methods like this, which use pattern matching to handle the incoming message:
+Then in other parts of your code you can write methods like this, which use pattern matching to handle (possibly by some methods similar to the ones mentioned as comments) the incoming message:
 
 ```scala
-def handleMessages(msg: Message) = message match
-  case PlaySong(name)      => playSong(name)
-  case IncreaseVolume(amt) => changeVolume(amt)
-  case DecreaseVolume(amt) => changeVolume(-amt)
-  case StopPlaying         => stopPlayingMusic
+def handleMessages(msg: Message): String = msg match
+  case PlaySong(name)      => s"Playing song ${name}"      // playSong(name)
+  case IncreaseVolume(amt) => s"Increase volume by ${amt}" // changeVolume(amt)
+  case DecreaseVolume(amt) => s"Decrease volume by ${amt}" // changeVolume(-amt)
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -685,14 +685,14 @@ case class DecreaseVolume(amount: Int) extends Message
 case object StopPlaying extends Message
 ```
 
-Then in other parts of your code you can write methods like this, which use pattern matching to handle the incoming message:
+Then in other parts of your code you can write methods like this, which use pattern matching to handle (possibly by some methods similar to the ones mentioned as comments) the incoming message:
```
julienrf, line 688:
```suggestion
Then in other parts of your code, you can write methods like this, which use pattern matching to handle the incoming message (assuming the methods `playSong`, `changeVolume`, and `stopPlayingSong` are defined somewhere else):
```

_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -685,14 +685,14 @@ case class DecreaseVolume(amount: Int) extends Message
...
-def handleMessages(msg: Message) = message match
-  case PlaySong(name)      => playSong(name)
-  case IncreaseVolume(amt) => changeVolume(amt)
-  case DecreaseVolume(amt) => changeVolume(-amt)
-  case StopPlaying         => stopPlayingMusic
+def handleMessages(msg: Message): String = msg match
+  case PlaySong(name)      => s"Playing song ${name}"      // playSong(name)
+  case IncreaseVolume(amt) => s"Increase volume by ${amt}" // changeVolume(amt)
+  case DecreaseVolume(amt) => s"Decrease volume by ${amt}" // changeVolume(-amt)
+  case StopPlaying         => s"Stopped playing music"     // stopPlayingMusic
```
julienrf, line 695:
```suggestion
def handleMessages(message: Message): Unit = message match
  case PlaySong(name)         => playSong(name)
  case IncreaseVolume(amount) => changeVolume(amount)
  case DecreaseVolume(amount) => changeVolume(-amount)
  case StopPlaying            => stopPlayingMusic()
```

julienrf (commented):
Personally, I prefer leaving the method names and adding a comment saying that they must be defined somewhere else. What do you think?

_overviews/scala3-book/domain-modeling-tools.md
vlulla: This sounds great! I like your changes. Thank you.

Since I'm fine with this change, but new at doing PR, should I hit the "Resolve conversation" button now? Please let me know if I need to do anything else.

SethTisue (approved):

julienrf (approved):

SethTisue:
Thank you!

Issue #1984, opened 2021-04-12
"Generic Classes" page in Tour of Scala uses an unnecessarily complicated example


knight-of-zero:
I'm new to scala, but have a deep background in Java. I've been going through the "tour of scala" to learn the language. Overall  this guide is super helpful and well written... but I did hit one snag on the [Generic Classes](https://docs.scala-lang.org/tour/generic-classes.html) page:

```
  def push(x: A): Unit =
    elements = x :: elements
```

At first I assumed `::` was some kind of built-in part of the language (like `+` on ints or strings). Eventually I hit [the Operators page](https://docs.scala-lang.org/tour/operators.html) which set me straight and described infix notation.

This was puzzling, though. If infix notation still goes `{object} {op} {argument}`, does that mean `Int` has a `def ::(to: List[Int])` on it? That seems a bit crazy given that `List` is generic, so you'd have to add `::` to every type in the stdlib to make things consistent.

I went down quite a rabbit hole (1-2 hours?) googling and running little experiments to try to figure out what was going on. I _eventually_ found my way to [the language spec](https://scala-lang.org/files/archive/spec/2.13/06-expressions.html), which briefly mentioned:

> The associativity of an operator is determined by the operator's last character. Operators ending in a colon ‘:’ are right-associative. All other operators are left-associative.

-----------------------------

All that said, I have two suggestions to improve the tour:

1. Avoid method names that end in colons throughout the other examples. The extra complexity just distracts from the other pages' main points. 
2. Add a small section to "operators" page so people know that infix notation doesn't _always_ behave like you'd expect.

SethTisue:
Agreed. Pull requests with these improvements would be welcome.

gkepka:
Completed by #2178

PR #2178, opened 2021-09-13
Update generics: use `xs.prepended(x)` instead of `x :: xs`

AesaKamar:
Addresses a simple issue https://github.com/scala/docs.scala-lang/issues/1984

Using a more [human readable method name](https://scala-lang.org/api/3.x/scala/collection/immutable/List.html#prepended-c0e), instead of a [prefix notation cons operator](https://scala-lang.org/api/3.x/scala/collection/immutable/::.html) which might be confusing to newcomers


julienrf (approved):
Thank you!

SethTisue:
thank you!

PR #2180, opened 2021-09-14
Update union-types.md

danecek:
// script.scala
type Hash=Int
case class UserName(name: String)
case class Password(hash: Hash)
val password = Password(123)
val name = UserName("Eve")
if true then name else password
val either: Password | UserName = if true then name else password

scala> :load script.scala
// defined alias type Hash = Int
// defined case class UserName
// defined case class Password
val password: Password = Password(123)
val name: UserName = UserName(Eve)
val res0: Object = UserName(Eve)
val either: Password | UserName = UserName(Eve)


julienrf (approved):
Thank you @danecek

PR #2181, opened 2021-09-14
Scala 2.12.15 (and JDK 17)

SethTisue:


SethTisue:
CI is failing because the API docs haven't synced to the site yet

PR #2185, opened 2021-09-21
显示 -> 显式

fanofxiaofeng:
Hello, I am a chinese. We usually use “显式”(or "显式地") instead of “显示” when we want to express the word "explicitly".
“显示” in chinese means something like "display" or "show"
 I also tried with google translate (https://translate.google.cn/?sl=en&tl=zh-CN&text=explicitly&op=translate)
 
image

The modification is shown as follows
image

### DIFF
--- a/_zh-cn/tour/basics.md
+++ b/_zh-cn/tour/basics.md
@@ -58,7 +58,7 @@ println(x) // 2
 x = 3 // This does not compile.
 ```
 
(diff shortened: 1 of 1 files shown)


liufengyun (approved):

eed3si9n (approved):
Thanks guys! 🥮

PR #2187, opened 2021-09-22
Update docs Scala for java devs

15DV:


SethTisue:
@15DV care to revise in response to the review feedback?

PR #2188, opened 2021-09-23
docs: reword pattern matching comment

JurgisSi:
Hey, I feel that `returns` fit better here since the method does not print the message.

Or if you want to have `prints` I have another pr. :)
https://github.com/scala/docs.scala-lang/pull/2189


SethTisue:
thank you!

PR #2189, opened 2021-09-23
docs: fix pattern matching example

JurgisSi:
Hey, I think the current comment is incorrect, since the method does not print the message.

I have another pr with comment rewording, if you feel it would be better:
https://github.com/scala/docs.scala-lang/pull/2188


Philippus (approved):

SethTisue:
I merged #2188 instead. Thank you!

PR #2195, opened 2021-09-30
FAQ: small improvement in 'this type' answer

SethTisue:
This comes up all the time, most recently at https://users.scala-lang.org/t/guarantees-of-this-type/7822

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -229,10 +229,12 @@ For an in-depth treatment of types vs. classes, see the blog post
 
 ### How can a method in a superclass return a value of the “current” type?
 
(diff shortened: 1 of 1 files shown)


BalmungSan (commented):
I would rephrase it a bit and maybe move it below the current paragraphs as a _"note"_.

> **Note:** Using `this.type` won't work; **unless your method will always return `this`**.
> People often try that **to refer to the same type of the current instance**, but `this.type` means "the singleton type of this instance", a different and too-specific meaning.

_(bold to emphasize my modifications)_

SethTisue:
I tweaked it a bit but I like keeping `this.type` first. 🤷, nbd either way I guess

PR #2197, opened 2021-10-04
Blend Scala 2 and Scala 3 “getting started” pages into a single page

julienrf:
- A lot of content is duplicated between those pages
- Several people got confused by the fact that `cs setup` installs the Scala 2 compiler (`scalac`), but that command is enough to work with Scala 3 projects (since they usually use a build tool, which automatically uses the appropriate compiler version).

With this PR, both pages /scala3/getting-started.html and /getting-started show exactly the same content, which is the result of merging the information that was present in both pages (except that now, it is defined in a single place), and tweaking to highlight the differences between Scala 2 and Scala 3 when necessary.

Result:

Screenshot 2021-10-06 at 13-56-34 Getting Started

The Japanese translation of /getting-started needs to be updated accordingly (maybe @xuwei-k could help?)


_getting-started/index.md
```diff
@@ -4,21 +4,32 @@ title: Getting Started
 partof: getting-started
 languages: [ja]
 includeTOC: true
+scala3: true
 
-redirect_from: "/getting-started.html"
+redirect_from:
+  - "/getting-started.html"
+  - "/scala3/getting-started.html"
```
julienrf, line 11: The old URL, https://docs.scala-lang.org/scala3/getting-started.html will still work (it will redirect to /getting-started).

julienrf:
Hey reviewers, may I draw your attention to this PR, please?

_getting-started/index.md
```diff
@@ -4,21 +4,32 @@ title: Getting Started
 partof: getting-started
 languages: [ja]
 includeTOC: true
+scala3: true
```
sjrd, line 7: I think shouldn't be here. It causes the "Work in Progress" banner to appear, and we probably don't want that for this page.

_getting-started/index.md
```diff
@@ -71,42 +106,69 @@ create a project called "hello-world".
...
+
 1. `cd` into `hello-world`.
-1. Run `sbt`. This will open up the sbt console.
+1. Run `sbt`. This opens up the sbt console.
 1. Type `~run`. The `~` is optional and causes sbt to re-run on every file save,
 allowing for a fast edit/run/debug cycle. sbt will also generate a `target` directory
 which you can ignore.
 
+When you’re finished experimenting with this project, press `[Enter]` to interrupt the `run` command.
+Then type `exit` or press `[Ctrl][d]` to exit sbt and return to your command line prompt.
```
sjrd, line 161:
```suggestion
Then type `exit` or press `[Ctrl+D]` to exit sbt and return to your command line prompt.
```
?

_getting-started/index.md
```diff
@@ -115,11 +177,11 @@ There are a multitude of mailing lists and real-time chat rooms in case you want
  -- This is handled in `resources/js/functions.js`.
  --> 
 <div style="display:none" id="stepOne-linux">
-       <code class="hljs">$ curl -Lo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup </code> <br>
+       <code class="hljs">$ curl -fLo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup </code> <br>
```
sjrd, line 180: Why this change? What's the difference?

sjrd (approved):
LGTM overall.

_getting-started/index.md
sjrd: > Maybe we should show neither the Scala 2 nor the Scala 3 banner?

Ideally, yes, indeed. Not sure how easy that is to do given the current infrastructure.

julienrf: It turns out it’s not so simple because that banner not only shows “Scala 2” or “Scala 3”, but it also contains a navigation menu with several items (Learn, Reference, Glossary, etc.), and that navigation menu is different for Scala 2 and Scala 3.

So, I am tempted to go back to having two URLs (`/getting-started`, and `/scala3/getting-started`) showing the same content except the banner, which would be specialized.

mlachkar (approved):

PR #2205, opened 2021-10-15
restore stickyness for scala 3 docs

bishabosha:
I was annoyed that the sidebar did not stick to the top of the page in the Scala 3 docs, this also fixes the height and allows scrolling of the sidebar independently of the main body.

- look at scala 3 book, scala 2 book, and perhaps the style guide

### DIFF
--- a/_sass/layout/toc.scss
+++ b/_sass/layout/toc.scss
@@ -6,6 +6,8 @@
   position: -webkit-sticky;
   position: sticky;
   top: 0;
+  max-height: 85vh;
+  overflow-y: auto;
 
   @include bp(large) {
     position: relative;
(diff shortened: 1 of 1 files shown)


bishabosha:
@julienrf here are some previews as GIFs:

### Getting Started (Scala 2)
capture-getting-started

### Scala 3 Book
capture-scala-3-book

### Tour of Scala
capture-tour-of-scala

SethTisue:
can we try and close this out, one way or the other?

julienrf:
After thinking more about this, I think one way to move forward would be to have two navigation panels (like on https://kotlinlang.org/docs/object-declarations.html), and to support collapsible sections on the navigation panel that links the pages together.

bishabosha:
I like Julien's idea

PR #2215, opened 2021-11-06
remove @migration ?

performantdata:
Does this still exist? Not seeing it in the API docs. @SethTisue ?


SethTisue:
It exists but is `private[scala]` these days.

SethTisue:
Thanks!

PR #2220, opened 2021-11-10
Translated the new getting-started page into Japanese.

rysh:
The new getting-started page was great, so I quickly translated it into Japanese.
I was careful not to change the overall meaning of the translation.
At the bottom of the page, there is Getting Help, where I added the current best way to ask a question in Japanese.

### DIFF
--- a/_ja/getting-started/index.md
+++ b/_ja/getting-started/index.md
@@ -6,47 +6,172 @@ language: ja
 includeTOC: true
 ---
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
some minor comments, but overall LGTM

rysh:
Hi @eed3si9n,
Thanks for the advice.
I have reflected them in the commit.

eed3si9n (approved):
Thanks!

PR #2222, opened 2021-11-12
Fix problem that language dropdown lost when change to ja language.

dongxuwang:
Fix problem:
On the home page of docs, change language to `日本語`, and on the new page, language dropdown would disappear.


SethTisue:
Thank you!

PR #2223, opened 2021-11-12
Updated example for filtering keys from Map

mukesh210:
This is regarding below example:
"val twoAndThree = m.view.filterKeys(Set(2,3)).toMap"

`filterKeys` is not a member of `scala.collection.IterableView`, so the example mentioned gives compile time error. Therefore, I updated the example to not use "view" and "toMap".

### DIFF
--- a/_overviews/scala-book/collections-maps.md
+++ b/_overviews/scala-book/collections-maps.md
@@ -55,7 +55,7 @@ scala> val ucMap = m.transform((k,v) => v.toUpperCase)
 ucMap: scala.collection.immutable.Map[Int,String] = Map(1 -> A, 2 -> B, 3 -> C, 4 -> D)
 
 // how to filter a Map by its keys
-scala> val twoAndThree = m.view.filterKeys(Set(2,3)).toMap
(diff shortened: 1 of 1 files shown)


SethTisue:
I think you must be using Scala 2.11 or 2.12?

The Scala collections changed a lot between 2.12 and 2.13.

The example works as written in both Scala 2.13 and Scala 3.

https://docs.scala-lang.org/overviews/scala-book/collections-maps.html aims to cover 2.13. At least, I don't think it's necessarily intended to work in 2.12 too? @alvinj?

mukesh210:
@SethTisue Thanks for the info... I updated Scala version in local to 2.13 and it's working fine.
Can we somehow mention that "example would work with 2.13 or later" in documentation.

SethTisue:
I believe a pull request adding that to https://docs.scala-lang.org/overviews/scala-book/introduction.html and/or https://docs.scala-lang.org/overviews/scala-book/prelude-taste-of-scala.html would be accepted. (I don't think we want to scatter such disclaimers throughout the book.)

Philippus:
I think we can close this PR.

PR #2227, opened 2021-11-16
FAQ: add sbt bare settings

SethTisue:
"I set a setting in sbt but nothing happened. Why?"

I can vouch for how frequently this is actually asked, in some form or another.

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -297,7 +297,6 @@ In many cases one should simply write instead:
 
 Other possibilities include:
 
-* `Global`
 * the common settings pattern, where you put shared settings
   in a `val`, typically named `commonSettings`, and then
   `.settings(commonSettings)` in every project you want to
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -277,3 +277,31 @@ give the compiler, then it's a compiler bug. Please report it on the
 [Scala 2 bug tracker](https://github.com/scala/bug/issues) or [Scala 3
 bug tracker](https://github.com/lampepfl/dotty/issues), but check
 first if it's a duplicate of an existing ticket.
+
+### I set a setting in sbt but nothing happened. Why?
+
+There could be a lot of reasons.  An extremely common one, that
+almost everyone runs into sooner or later, is that you have a bare
+setting in a multi-project build.
+
+For example, if you add this to your `build.sbt`:
+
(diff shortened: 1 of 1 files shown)


_overviews/FAQ/index.md
```diff
@@ -277,3 +277,31 @@ give the compiler, then it's a compiler bug. Please report it on the
...
+
+### I set a setting in sbt but nothing happened. Why?
+
+There could be a lot of reasons.  An extremely common one, that
+almost everyone runs into sooner or later, is that you have a bare
+setting in a multi-project build.
+
+For example, if you add this to your `build.sbt`:
+
+    scalaVersion := "2.13.7"
```
eed3si9n, line 289: For `scalaVersion` in particular, could we suggest

```scala
ThisBuild / scalaVersion := "2.13.7"
```

up front? If that fixes 80% of the users, they don't really need to read further explanation.

Also this is an own-goal problem created by hello-world.g8
https://github.com/scala/hello-world.g8/blob/e65d2a3d300af1c32bbd34037614daf62cd58686/src/main/g8/build.sbt#L4

SethTisue: pushed an edit. wdyt?

eed3si9n: looks good.

_overviews/FAQ/index.md
```diff
@@ -277,3 +277,34 @@ give the compiler, then it's a compiler bug. Please report it on the
...
+that's a "bare" setting, and you might expect it to apply build-wide.
+But it doesn't. _It only applies to the root project._
+
+In many cases one should simply write instead:
+
+    ThisBuild / scalaVersion := "2.13.7"
+
+Other possibilities include:
+
+* `Global`
```
dwijnand, line 300: This isn't a good suggestion.  Global is only for defaults (and some exceptional keys that operate outside of all builds).  If you want to change a key across all the projects in the build use ThisBuild.

But you could suggest a local auto-triggering plugin as an alternative to the `.settings(commonSettings)` spammy option.

SethTisue: Okay, I removed the `Global` bullet.

Scala FAQ answers intend to be very brief, and any in-depth information is provided only as links. I think the auto-triggering-plugin thing seems beyond the scope of what's necessary to include, it seems like pretty advanced information to me, and I presume it's covered by the doc page we link to.

dwijnand (approved):

PR #2228, opened 2021-11-20
[zh-cn] Chinese Simple translation for the new overview page.

dongxuwang:


eed3si9n (approved):
Thanks!

_data/overviews-zh-cn.yml
```diff
@@ -0,0 +1,324 @@
+- category: 标准库
+  description: "涵盖Scala标准库的参考与概览"
+  overviews:
+    - title: Scala 容器
```
eed3si9n, line 4: The current page says "Scala的容器类" so changing this to "Scala 容器" hopefully is not controversial because it's closer to the original "Scala Collections" in https://docs.scala-lang.org/overviews/index.html

The only potential thing to note is the whitespace after Scala, which I personally prefer but does not seem to be the convention applied consistently.

dongxuwang: Thanks for the review.
1. The reason that keeping "Scala 容器" is because not just what you're referring, but also there has been an additional description "Scala 的容器库" which is demonstrating the title in detail. Thus the title needs to be simpler and conciser.

2.  Yes, totally agree the translation needs to be kept consistent. I added spaces to wrap the English terminologies and numbers in the translation and re-pushed.

PR #2233, opened 2021-11-23
add fr traduction for the page basics

apointeau:


julienrf (commented):
I think this one should be put on hold, since scalafiddle is not running anymore. (the problem also exists for the English version…)

julienrf:
Hey Antoine, would you be interested in going back to this work? You would have to sync your translation with the current state of the English version.

PR #2237, opened 2021-11-27
Improve consistency

ossScharom:
The name of the new Secret could be only inferred from the the contents of the file `.github/workflows/publish.yml`, which is also located below the edited section. Stating the name of the new Secret is more consistent to the expectation of the reader.


SethTisue:
Thank you!

PR #2240, opened 2021-11-29
Add the match expressions and switch/match content for Python part

### LINKED ISSUE #2232: Scala 3 for python developers: switch/match/case
From python 3.10 (see [PEP 634](https://www.python.org/dev/peps/pep-0634/)) there is already structural pattern matching.

So comparision of "match expressions:" and "switch/match:" is probably no longer "N/A" for python.

I'm not using python >=3.10 nor scala so I'm not able to suggest proper python alternative.

byyue:
Close #2232.
1. Structural pattern matching was added in Python 3.10, which can be reflected in the Scala for Python Developers chapter. 
2. Another subtle change is to replace "day" with "month" so that the code snippet can be more intuitive.

### DIFF
--- a/_overviews/scala3-book/scala-for-python-devs.md
+++ b/_overviews/scala3-book/scala-for-python-devs.md
@@ -650,12 +650,20 @@ Scala also has `match` expressions.
   <tbody>
     <tr>
       <td class="python-block">
-        N/A (but you can use dictionaries for basic “switch” functionality)
+        <code># From 3.10, Python supports structural pattern matching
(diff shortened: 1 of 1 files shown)


galuszkak:
👍 I was actually wanting to contribute that change. Thanks @byyue!

byyue:
> 👍 I was actually wanting to contribute that change. Thanks @byyue!

Thanks! Maybe you could help ping maintainers if you want as they probably overlooked this PR.

julienrf (approved):
Thank you!

PR #2241, opened 2021-11-29
JDK compat page: update JDK 17 wordings

SethTisue:


SethTisue:
@eed3si9n you might want to review the sbt wording

_overviews/jdk-compatibility/overview.md
```diff
@@ -73,7 +75,9 @@ Scala 2.13.6 and 2.12.15 support JDK 17.
 
 The Scala test suite and Scala community build are green on JDK 17.
 
-For possible issues, see the [jdk11](https://github.com/scala/bug/labels/jdk11) and [jdk17](https://github.com/scala/bug/labels/jdk17) labels in the Scala 2 bug tracker.
+For sbt users, sbt 1.6.0-RC1 is the first version to fully support JDK 17, but 1.5.5 works fine too. (It will print a warning on startup about `TrapExit` that you can ignore.)
```
eed3si9n, line 78:
```suggestion
For sbt users, sbt 1.6.0-RC1 is the first version to support JDK 17, but in practice sbt 1.5.5 may also work. (It will print a warning on startup about `TrapExit` that you can ignore.)
```

"work fine" written on the official compatibility documentation sounds like an effective statement of support. Just a few paragraph above it says Scala 2.12.15 supports JDK 17. Note that in addition to dropping `TrapExit` sbt 1.6.0 bumps the metabuild to use Scala 2.12.15, so saying "works fine" (even though in practice it would for 90%) is sort of saying Scala 2.12.14 is also fine.

eed3si9n (approved):
Thanks Seth!

PR #2252, opened 2021-12-02
add fr tour mixin class composition

apointeau:


SethTisue:
@sjrd review, or suggest a reviewer?

PR #2265, opened 2021-12-06
taste-repl.md added 'scala3'-command to start REPL

A-Roso:
Added 'scala3' as command to start REPL as only 'scala' doesn't work depending on the installation (e.g. with Coursier on Windows 10).


SethTisue:
Thank you!

julienrf:
Actually, this has been solved and now with Coursier you should be able to run the Scala 3 REPL by invoking the `scala` command. Can a Windows user confirm this?

PR #2272, opened 2021-12-13
[zh-cn] Initialization and Introduction

byyue:
Translate Scala 3 Book into Chinese step by step.
1. Add a copy of scala3-book to _zh-cn.
2. Complete the translation of the Introduction chapter.


byyue:
@SethTisue @julienrf @eed3si9n @Philippus 
Sorry to interrupt, but could you please take a look at this PR at your convenience?

julienrf:
Hello @byyue, thank you for the pull request (and for your patience!).

I didn’t have a look at how it works with the other translations, but we should not copy all the pages, if possible. Instead, we should only copy the translated pages. Did you try that approach?

byyue:
> Hello @byyue, thank you for the pull request (and for your patience!).
> 
> I didn’t have a look at how it works with the other translations, but we should not copy all the pages, if possible. Instead, we should only copy the translated pages. Did you try that approach?

Hi @julienrf, thank you so much for your prompt reply. 

Yes I have already considered what you have mentioned. The key point is the parts I would like to translate are under a different file structure than past translations related to Scala 2. It looks like I'm the first person to try to do this.

Another point I considered was to provide a smooth reading experience for the reader: If a chapter has not yet been translated, they can read the original English version in the current directory structure instead of jumping to another link. That's why I copied all the pages.

I did spend some time reading the [contribution doc](https://docs.scala-lang.org/contribute.html) and trying to find better solution but to no avail. If you have a preferable solution, please let me know!

julienrf:
> Yes I have already considered what you have mentioned. The key point is the parts I would like to translate are under a different file structure than past translations related to Scala 2. It looks like I'm the first person to try to do this.
> 
> I did spend some time reading the [contribution doc](https://docs.scala-lang.org/contribute.html) and trying to find better solution but to no avail. If you have a preferable solution, please let me know!

Sorry about that. Someone needs to check how to translate those pages then, but eventually there should be no need to copy the old files. (I can’t do that for now, unfortunately)

> Another point I considered was to provide a smooth reading experience for the reader: If a chapter has not yet been translated, they can read the original English version in the current directory structure instead of jumping to another link. That's why I copied all the pages.

I understand but then the risk is that we end up with different English versions of the pages, which might bring a lot of confusion.

byyue:
> > Yes I have already considered what you have mentioned. The key point is the parts I would like to translate are under a different file structure than past translations related to Scala 2. It looks like I'm the first person to try to do this.
> > I did spend some time reading the [contribution doc](https://docs.scala-lang.org/contribute.html) and trying to find better solution but to no avail. If you have a preferable solution, please let me know!
> 
> Sorry about that. Someone needs to check how to translate those pages then, but eventually there should be no need to copy the old files. (I can’t do that for now, unfortunately)
> 
> > Another point I considered was to provide a smooth reading experience for the reader: If a chapter has not yet been translated, they can read the original English version in the current directory structure instead of jumping to another link. That's why I copied all the pages.
> 
> I understand but then the risk is that we end up with different English versions of the pages, which might bring a lot of confusion.

As you suggested, I spent some time figuring out how to make this process more elegant. The current plan is to translate each chapter/section step by step and commit separately. Before the translation is complete, the work-in-progress would not be visible to the reader (but can be accessed directly via the specific link). Cloud you please re-review this? @julienrf

byyue:
Could you please have a look at this as well? @eed3si9n

eed3si9n:
@byyue I'd love to but my Mandarin is not great.

@scala/docs-zh 祝大家新年快乐身体健康! could you plz review this?

_zh-cn/overviews/scala3-book/introduction.md
```diff
@@ -0,0 +1,30 @@
+---
+title: 导言
+type: chapter
+description: This page begins the overview documentation of the Scala 3 language.
```
julienrf, line 4: Do you want to also translate this one?

_zh-cn/overviews/scala3-book/introduction.md
```diff
@@ -0,0 +1,30 @@
...
+num: 1
+previous-page: 
+next-page:
```
julienrf, line 7: How is the page rendered with these empty values? Should we temporarily remove these properties?

_zh-cn/overviews/scala3-book/introduction.md
```diff
@@ -0,0 +1,30 @@
...
+description: This page begins the overview documentation of the Scala 3 language.
+num: 1
+previous-page: 
+next-page: 
+
+scala3: true
+partof: scala3-book
+overview-name: "Scala 3 — Book"
+layout: multipage-overview
+permalink: "/zh-cn/scala3/book/:title.html"
```
julienrf, line 13: Is this property needed?

julienrf (commented):
Thank you! The front matter seems good, but could you please post a screenshot to show that it works as expected?

_zh-cn/overviews/scala3-book/introduction.md
liufengyun: The original [file](https://raw.githubusercontent.com/scala/docs.scala-lang/main/_overviews/scala3-book/introduction.md):
```
---
title: Introduction
type: chapter
description: This page begins the overview documentation of the Scala 3 language.
num: 1
previous-page: 
next-page: scala-features
---
```

liufengyun (approved):
LGTM

_zh-cn/overviews/scala3-book/introduction.md
byyue: Type and description are invisible to readers, so we'd better leave them as they are.

_zh-cn/overviews/scala3-book/introduction.md
byyue: We can regard each chapter/section as a node in a doubly linked list. The properties can be null but should be there.

_zh-cn/overviews/scala3-book/introduction.md
byyue: These were deliberately added, which are related to the rendering of the page. You may notice:

https://github.com/scala/docs.scala-lang/blob/3b614bee449537d8e0b9b567ce7e36137eb7416f/_config.yml#L96-L106

Without messing up the global config file, I decided to add this snippet to each translation page, which should be more maintainable.

julienrf:
Thank you!

byyue:
> Thank you! The front matter seems good, but could you please post a screenshot to show that it works as expected?

intro

_zh-cn/overviews/scala3-book/introduction.md
byyue: Clarification: type is related to the rendering of the page, and description is invisible to readers.

PR #2274, opened 2021-12-16
[doc] small improvements

robstoll:
IMO better to write out `'ll` because english beginners sometimes struggle with it


robstoll:
Can someone re-trigger the build please, I doubt this has something to do with my change

SethTisue:
Yeah you can ignore the CI failure (though I believe it would go away if you rebase onto current `main`).

About your second edit though, you've rewritten it from "Because A, B" to "A, because B", which reverses the meaning. I would suggest "thus the dozens" instead of "because the dozens", but it's sort of a nitpick, wdyt?

robstoll:
I have changed the meaning because IMO it was wrong  (I guess one knew what the author meant though). Maybe I get it wrong but IMO it's not that there are dozens of stdlib functions because you rarely write a for loop, you rarely write a for loop because there are dozens of stdlib functions. I also think that `thus` does not reflect this aspect as nice as `because` but I am not a native speaker.

Let's see if dependabot can rebase this (I guess its not configured in this repo but could be useful for such a case)
@dependabot rebase

SethTisue:
Thank you!

PR #2279, opened 2021-12-19
scala3-book: Fix abstract class example

### LINKED ISSUE #2278: domain-modelling-tools: Use `name` argument in abstract class
In Scala 3 Book, under [Abstract classes](https://docs.scala-lang.org/scala3/book/domain-modeling-tools.html#abstract-classes), the example for an abstract class that can take arguments doesn't really use the argument.

```scala
abstract class Pet(name: String):
  def greeting: String
  def age: Int
  override def toString = s"I say $greeting, and I’m $age"
```

However, the immediate example that shows "traits with parameters" use them.

I'd be happy to send a PR.

arkivm:
Use `name` parameter inside the abstract class

Fixes #2278


Philippus (approved):

Philippus:
Thanks @arkivm !

PR #2292, opened 2022-01-11
Tweak implicit classes overview

som-snytt:
Per https://github.com/lampepfl/dotty/issues/3654

### DIFF
--- a/_overviews/core/implicit-classes.md
+++ b/_overviews/core/implicit-classes.md
@@ -80,9 +80,9 @@ While it's possible to create an implicit class with more than one non-implicit 
 aren't used during implicit lookup.
 
 
-**3. There may not be any method, member or object in scope with the same name as the implicit class.**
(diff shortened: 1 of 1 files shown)


SethTisue:
I've never seen the backquote trick in a real codebase. Do people really do that?

I think it might be better to relegate it to a footnote than to actually suggest it in the main text?

som-snytt:
I guess by real codebase you mean Akka. I've seen it recommended in public chat, and I was like, hey I thought that was my idea.

I'm not invested in overview examples for obsolete features, so I pushed the minimum viable correction.

There should be a big label somewhere that says, "Overviews are not the spec."

Scala 3 is still loose about where the holiest words are kept.

som-snytt:
I forgot the preserve the corrected sip URL, but it forwards anyway.

SethTisue (approved):

PR #2295, opened 2022-01-21
Add missing function invocation syntax

andrewcameronsims:
Hi! 👋 

While reading the documentation of Futures I noticed that one of the code examples is missing the function invocation syntax (you'll see it is present in the previous code example, but not this one).

Let me know if I'm mistaken here, but I think it's supposed to be in this example too.

Thanks!


Philippus (approved):

julienrf:
Good catch, thank you Andrew!

PR #2301, opened 2022-01-28
Add reference docs for forward compatibility feature

prolativ:
Adapted from https://github.com/lampepfl/dotty/pull/14156


_scala3-reference/language-versions.md
```diff
@@ -1,36 +1,11 @@
...
-
-```scala
-package p
-import scala.language.`future-migration`
-
-class C { ... }
-```
-
-Language imports supersede command-line settings in the source files where they are specified. Only one language import specifying a source version is allowed in a source file, and it must come before any definitions in that file.
+Additional information on interoperability and migration between Scala 2 and 3 can be found [here](https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html).
```
julienrf, line 11: Please use the `{% link ... %}` directives to create internal links.

_scala3-reference/language-versions.md
```diff
@@ -1,36 +1,11 @@
...
 type: chapter
-description: This page lists the different flavours of language supported by the Scala 3 compiler.
 num: 100
-previous-page: overview
+previous-page: /scala3/reference/syntax
+next-page: /scala3/reference/language-versions/source-compatibility
 ---
 
-The default Scala language version currently supported by the Dotty compiler is `3.0`. There are also other language versions that can be specified instead:
+This chapter explains how different versions of the language interact with each regarding their inputs and outputs.
```
julienrf, line 9: I am not an english native, but “how they interact with each” does not sound correct to me, could you please double-check?

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Binary Compatibility
+type: section
+num: 102
+previous-page: /scala3/reference/language-versions/source-compatibility
+---
+
+In Scala 2 different minor versions of the compiler were free to change the way how they encode different language features in JVM bytecode so each bump of the compiler's minor version resulted in breaking binary compatibility and if a project had any Scala dependencies they all needed to be (cross-)compiled to the same minor Scala version that was used in that project itself. On the contrary, Scala 3 has a stable encoding into JVM bytecode.
```
julienrf, line 8: Minor suggestions:

```suggestion
In Scala 2, different minor versions of the compiler were free to change the way they encoded different language features in JVM bytecode. So, each bump of the compiler's minor version resulted in breaking the binary compatibility, and if a project had any Scala dependencies they all needed to be (cross-)compiled to the same minor Scala version that was used in that project itself. On the contrary, Scala 3 has a stable encoding into JVM bytecode.
```

However, I am not sure we want to guarantee that “Scala 3 has a stable encoding into JVM bytecode”? Has this been discussed previously? I thought that, at some point we would like to leverage new features of the JVM, and we could recompile TASTy artifacts to still be able to use Scala libraries compiled before those new features of the JVM were used by the compiler, without having to recompile everything. But, I am not sure what are the plans now.

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
...
+
+In addition to classfiles the compilation process in Scala 3 also produces files with `.tasty` extension. The [TASTy](https://docs.scala-lang.org/scala3/guides/tasty-overview.html) format is an intermediate representation of Scala code containing full information about sources together with information provided by the typer. Some of this information is lost during generation of bytecode so Scala 3 compilers read TASTy files during compilation in addition to classfiles to know the exact types of values, methods, etc. in already compiled classes (although compilation from TASTy files only is also possible). TASTy files are also typically distributed together with classfiles in published artifacts.
```
julienrf, line 10: Maybe introduce the wording “fully elaborated Scala programs”? TASTy contains fully elaborated Scala programs in the sense that it contains all the information (context parameters and type parameters) that have been inferred by the compiler during the compilation.

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
...
+As this feature is experimental it does not have any special support in build tools yet (at least not in sbt 1.6.1 or lower).
+E.g. when a project gets compiled with Scala compiler `3.x1.y1` and `-Yscala-release 3.x2` option and then published using sbt
+then the standard library in version `3.x1.y1` gets added to the project's dependencies instead of `3.x2.y2`.
+When the dependencies are added to the classpath during compilation with Scala `3.x2.y2` the compiler will crash while trying to read TASTy files in the newer format.
+A currently known workaround is to modify the build definition of the dependent project by explicitly overriding the version of Scala standard library in dependencies, e.g.
+
+```scala
+dependencyOverrides ++= Seq(
+  scalaOrganization.value %% "scala3-library" % scalaVersion.value,
+  scalaOrganization.value %% "scala3-library_sjs1" % scalaVersion.value // for Scala.js projects
```
julienrf, line 34: Would this work?

```suggestion
  scalaOrganization.value %% "scala3-library" % scalaVersion.value,
  scalaOrganization.value %%% "scala3-library" % scalaVersion.value // for Scala.js projects
```

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
...
+* the code contains no references to parts of the standard library which were added to the API after `<release-version>` and would crash at runtime when a program is executed with the older version of the standard library on the classpath
+* no dependency found on the classpath during compilation (except for the standard library itself) contains TASTy files produced by a compiler newer than `<release-version>` (otherwise they could potentially leak such disallowed references to the standard library).
+
+If any of the checks above is not fulfilled or for any other reason older TASTy cannot be emitted (e.g. the code uses some new language features which cannot be expressed the the older format) the entire compilation fails (with errors reported for each of such issues).
+
+As this feature is experimental it does not have any special support in build tools yet (at least not in sbt 1.6.1 or lower).
```
julienrf, line 25: Just to understand better: we’d need build tools to special-handle modules compiled with `-Yscala-release` to change their dependency to the scala3-library that is shown in their pom.xml?

Do you plan to do this work?

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
...
+E.g. when a project gets compiled with Scala compiler `3.x1.y1` and `-Yscala-release 3.x2` option and then published using sbt
+then the standard library in version `3.x1.y1` gets added to the project's dependencies instead of `3.x2.y2`.
+When the dependencies are added to the classpath during compilation with Scala `3.x2.y2` the compiler will crash while trying to read TASTy files in the newer format.
+A currently known workaround is to modify the build definition of the dependent project by explicitly overriding the version of Scala standard library in dependencies, e.g.
```
julienrf, line 29: Is there a known workaround that could be applied on the library itself, instead of the modules that depend on the library? Maybe a way to override the `package` task in sbt to override the version of the dependency to `scala3-library`? Maybe @eed3si9n knows a trick?

julienrf (commented):
Thank you @prolativ for the update of documentation.

The page `binary-compatibility` contains useful information, but I think it would be good to summarize somewhere what are the relevant action points for library maintainers, with a concrete example.

_scala3-reference/language-versions.md
prolativ: Good catch. This was supposed to be `with each other`

_scala3-reference/language-versions/binary-compatibility.md
prolativ: https://github.com/lampepfl/dotty/pull/14156#discussion_r786765757

_scala3-reference/language-versions/binary-compatibility.md
prolativ: I tried this but for some reason this didn't seem to work

_scala3-reference/language-versions/binary-compatibility.md
prolativ: Yes, I'm currently trying to make this work for sbt

prolativ:
We can add more documentation later when we set some best practices how to make this work smoothly. For now it's still experimental but hopefully we should get some support for this feature in build tools.

_scala3-reference/language-versions/binary-compatibility.md
```diff
@@ -0,0 +1,38 @@
...
+
+In addition to classfiles the compilation process in Scala 3 also produces files with `.tasty` extension. [TASTy]({% link scala3/guides/tasty-overview.md %}) files are an intermediate representation of Scala programs containing fully elaborated Scala code in the sense that they contain all the information that have been inferred by the compiler during the compilation (like contextual parameters and type parameters). Some of this information is lost during generation of bytecode so Scala 3 compilers read TASTy files during compilation in addition to classfiles to know the exact types of values, methods, etc. in already compiled classes (although compilation from TASTy files only is also possible). TASTy files are also typically distributed together with classfiles in published artifacts.
```
julienrf, line 10: Sorry to be picky, but:

```suggestion
In addition to classfiles, the compilation process in Scala 3 also produces files with `.tasty` extension. [TASTy]({% link scala3/guides/tasty-overview.md %}) files represent fully elaborated Scala programs in the sense that they contain all the information that has been inferred by the compiler during the compilation (like contextual parameters and type parameters). Some of this information is lost during the generation of bytecode so Scala 3 compilers read TASTy files during compilation in addition to classfiles to know the exact types of values, methods, etc. in already compiled classes (although compilation from TASTy files only is also possible). TASTy files are also typically distributed together with classfiles in published artifacts.
```

julienrf (approved):

PR #2303, opened 2022-02-01
[Blocker] Fix an invalid link to pass the HTML Proofer

### LINKED ISSUE #2302: A link is invalid
https://dotty.epfl.ch/docs/reference/other-new-features/explicit-nulls.html is not available anymore, which will fail the build.

byyue:
Close #2302.


eed3si9n (approved):
Thanks!

PR #2306, opened 2022-02-02
Update which courses use Scala 2 and which courses use Scala 3

julienrf:


mlachkar (approved):

vincenzobaz (approved):
:sparkles:

byyue:
The link is invalid, and I think the Spark course is still using Scala 2 actually.

julienrf:
True, the link is not yet valid, sorry for the mess regarding the CI, I was too eager to merge it! It should become public soon (within a couple of days), I propose that we ignore that specific error in the meantime.

PR #2317, opened 2022-02-06
Fix a broken link

komainu8:
"Big Data Analysis with Scala and Spark (Scala 2 version)" doesn't exist and the link to this page has been already invalid.

Currently, It seems that "Big Data Analysis with Scala and Spark" exists instead of "Big Data Analysis with Scala and Spark (Scala 2 version)" on Coursera.

In addition, the test on GitHub Actions is failing by the link to "Big Data Analysis with Scala and Spark (Scala 2 version)" is invalid.


julienrf:
Hello, thank you for your initiative. However, the link should be active within a couple of days so I suggest we simply ignore the CI failure for now. I am sorry about this mess, I was expecting the link to work earlier, I should have waited longer.

komainu8:
@julienrf Thank you for sharing the information! I understand about we ignore the CI failure for now.

PR #2324, opened 2022-02-09
Some fp-wise reformatting of several (namely 6) examples.

ac2epsilon:
Subject, plus minor changes to make code more FP and Copy-n-Paste friendly. Tested thoroughly in Scastie


alvinj:
> is there a policy around this in the book generally

No, we didn’t explicitly talk about a policy, though I think I may have talked to @julienrf a little bit about trying to make as much of the code as interactive as possible. To that end, I like these changes (more expression-oriented and interactive). I would just change the one `val num =...` reference to a different variable name, because that expression doesn’t return a number.

SethTisue:
@ac2epsilon care to update this PR based on Alvin's feedback?

ac2epsilon:
oops. Yes-yes I fixed renaming num= to code=. I do not know if I should re- PR or something

PR #2328, opened 2022-02-10
Update the Getting Started page(s) for the latest `cs` behavior.

sjrd:
The default installation of coursier now installs Scala 3 by default under `scala` and `scalac`. However, it also allows to override that with `cs launch scala:2.13.8` or `cs install scala:2.13.8`, as Scala 2 and Scala 3 have been unified.

---

It seems that this is the only file that actually needs to be updated. :thinking:  @julienrf Did I miss anything?


julienrf (approved):
Great, thank you!

PR #2332, opened 2022-02-15
Fix "Matching types of expression" example in quotes.md

Bersier:
The definition of exprOfOption seems broken, returning values of a different type than declared. This is an attempt to fix it.

Note that, to get this function to compile in Scala 3.1.1, I had to add an explicit cast:

```scala
import scala.quoted.{Expr, Quotes, Type}

def exprOfOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Expr[T]] = x match {
  case '{ Some($x) } => Some(x.asInstanceOf[Expr[T]])
  case '{ None } => None
}
```

Otherwise, I get:

```
Found:    (x : scala.quoted.Expr[Any])
Required: quoted.Expr[T]
  case '{ Some($x) } => Some(x)
```

### DIFF
--- a/_overviews/scala3-macros/tutorial/quotes.md
+++ b/_overviews/scala3-macros/tutorial/quotes.md
@@ -207,8 +207,7 @@ Below, we match against `$x` of type `T` and we get out an `x` of type `Expr[T]`
 def exprOfOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Expr[T]] =
   x match
     case '{ Some($x) } => Some(x) // x: Expr[T]
-    case '{ None } => Some(None)
-    case _ => None
+    case '{ None } => None
 ```
 
 We can also check for the type of an expression:


SethTisue:
@nicolasstucki ping?

_overviews/scala3-macros/tutorial/quotes.md
```diff
@@ -207,8 +207,7 @@ Below, we match against `$x` of type `T` and we get out an `x` of type `Expr[T]`
 def exprOfOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Expr[T]] =
```
nicolasstucki, line 207: The original code should have been

```scala
def exprOfOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Option[Expr[T]]] =
  x match
    case '{ Some($x) } => Some(Some(x)) // x: Expr[T]
    case '{ None } => Some(None)
    case _ => None
```

Not sure if the nested option might be confusing. The idea is that if it does not contain `Some(..)` or `None` it returns `None`, otherwise it return `Some` of the partially unlifted option it contains.

julienrf: What can it be other than `Some` or `None`?

nicolasstucki: The explicitness was to differentiate between `Some(Some(..))`, `Some(None)` and `None`.

Bersier: But when would `x` ever match neither `'{ Some($x) }` nor `'{ None }`? Aren't these the only possible options for `x: Expr[Option[T]]`?

nicolasstucki: If you would do 

```scala
def f[T: Type](x: Expr[Option[T]])(using Quotes) =
  x match
    case '{ $y: Some[T] } => // y: Expr[Some[T]]
    case '{ $y: None.type } => 
```

then you would cover all possible expressions that have `Some` or `None` type.

julienrf: OK, thank you for clarifying. I suggest also renaming the method to `exprOfLiteralOption`.

Bersier: Looking at the rest of this page, I think this needs a bigger fix, as this example gets discussed and modified further in the page, and it wouldn't quite make sense anymore if I changed it to

```
def exprOfLiteralOption[T: Type](x: Expr[Option[T]])(using Quotes): Option[Option[Expr[T]]] =
  x match
    case '{ Some($x: T) } => Some(Some(x)) // x: Expr[T]
    case '{ None } => Some(None)
    case _ => None
```

I'd be happy to have a call to discuss further.

PR #2335, opened 2022-02-21
Fix deprecated warning in scala for java programer

yechs:
```
warning: method + in class Int is deprecated (since 2.13.0):
Adding a number and a String is deprecated. Use the string interpolation `s"$num$str"`
override def toString(): String = year + "-" + month + "-" + day
```


Philippus (approved):
LGTM

yechs:
The CI error seem to be the same as main branch and not relevant to the changes made in the PR.

If it is preferred, I can rebase onto main branch as soon as it is fixed in main.

Philippus:
Not needed, thanks for the fix!

PR #2344, opened 2022-03-14
list scala 3 releases in api/all

bishabosha:
I propose that we also list the scala 3 APIs in this documentation page for new releases.

paging @SethTisue @Kordyjan 

The plan is to then link to this from the scala-lang homepage where it currently says "API Docs (All Releases)"

### DIFF
--- a/api/all.md
+++ b/api/all.md
@@ -6,6 +6,8 @@ includeTOC: true
 
 ## Latest releases
 
+* Scala 3.1.1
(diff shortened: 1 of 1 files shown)


julienrf (approved):
The top banner of the page says “Scala 2”, but I propose that we address this more general issue in a separate PR.

@Kordyjan could you please make sure to include updating this file in your release process?

SethTisue (approved):

PR #2345, opened 2022-03-15
Even more section literally doubles chapters Summary

ac2epsilon:
Former one removed

// BTW who knows how to clean commit history to see only (last) commit of interest?


_overviews/scala3-book/methods-most.md
```diff
@@ -279,16 +288,15 @@ val d = new Dog
 d.speak()   // prints "Woof"
 ```
 
-Methods can also be marked as `private`.
-This makes them private to the current class, so they can’t be called nor overridden in subclasses:
+Methods can also be marked as `private`. This makes them private to the current class, so they can’t be called nor overridden in subclasses:
```
som-snytt, line 291: Making it one line doesn't change rendering, but "nor" should be "or".

_overviews/scala3-book/methods-most.md
```diff
@@ -269,7 +278,7 @@ For more details on the `Matchable` trait, see the [Reference documentation][ref
 
 ## Controlling visibility in classes
 
-In classes, objects, traits, and enums, Scala methods are public by default, so the `Dog` instance created here can access the `speak` method:
+In classes, objects, traits, and enums, Scala methods are public by default, so world can access the `speak` method of `Dog` instance created here:
```
som-snytt, line 281: yes, the previous version was wrong. But I think the whole section needs a rewrite, not tweaks about what a cat says.

_overviews/scala3-book/methods-most.md
```diff
@@ -383,23 +391,6 @@ See the [Extension methods section][reference_extension_methods] of this book, a
...
-- Create inline methods
-- Handle exceptions
-- Use vararg input parameters
-- Write methods that have multiple parameter groups (partially-applied functions)
-- Create methods that have generic type parameters
-
-See the [Reference documentation][reference] for more details on these features.
-
-
-
```
som-snytt, line 402: I agree placeholder text isn't helpful unless it is meant to be expanded in place. I'm not actually sure if this is meant to be closer to the Scala tour than an overview.

_overviews/scala3-book/control-structures.md
```diff
@@ -357,8 +357,9 @@ Notice that this really is an expression, as it evaluates to a result:
 ```scala
 import scala.annotation.switch
 
-// `i` is an integer
-val day = i match
+val i = 5 // `i` is an integer
+
+val day = (i : @switch) match
```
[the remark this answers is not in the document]
ac2epsilon: What do You mean by "wonky"? It is the most @switch-able case. As for me. If we are taking these tricks as given at all. And what is wrong with the colons? Anyway, no @switch - no problem. So, what to do? Should we drop the whole @switch-ing part? I believe this is something one can live a happy life without.

_overviews/scala3-book/methods-most.md
```diff
@@ -219,7 +218,17 @@ speak     // error: "method speak must be called with () argument"
...
+The main intention behind this convention is to make transition field-to-method and visa-versa easy for developers and transparent to method consumers. For example, given the following code, no one can tell if `speak` is a field or method:  
+
+```scala
+val alleyCat = Cat("Oliver")
+println(alleyCat.speak)
+```
+
+In FP terms "call of pure function" and "result of pure function" (with given arguments) are totally the same thing. While code does not change any existing state we can substitute field instead of method, and method instead of the field. The result must be the same. 
+
+However, in the case of impure code consumers must be warned and alarmed at least, because impure code tends to make dangerous things, like throwing `IOException`. That is why there is a difference between an absent list and an empty argument list.
```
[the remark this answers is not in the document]
ac2epsilon: Replaced by some short mention of UAP

_overviews/scala3-book/methods-most.md
ac2epsilon: "Or" done! "This" is not a question of rendering, but "This" implies some idea development in the scope of a paragraph. Anyway...

_overviews/scala3-book/methods-most.md
ac2epsilon: and?

_overviews/scala3-book/methods-most.md
ac2epsilon: and?

_overviews/scala3-book/control-structures.md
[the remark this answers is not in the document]
som-snytt: I just meant that if the material is "introductory", the switch adds an "advanced concept" and syntax.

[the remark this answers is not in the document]
ac2epsilon: Well, then. I leave everything in my, at least working, edition.

ac2epsilon:
Sorry, kind of bombing here. Literally. Distracts. A little. Kind of.

bishabosha:
Hey please could you try to rebase this, since you opened the PR we added code tabs with syntax for both scala 2 and 3

PR #2347, opened 2022-03-17
Potential Semantic Clarity

0xrutvij:
Hello! Firstly, I would like to thank all the contributors for the awesome tour of Scala Lang. While going through the example, the usage of `Node[Bird]` type for birdList threw me in a loop for a moment. The location of usage has been proposed as a change for both demonstration purposes and as a potential change.

I understand traits can be used to describe a type, but in this instance wouldn't it be clearer to say `ListNode[Bird]` instead? Seeing that the value name –birdList– indicates we are going for a list. From what I can surmise, the Node trait can be implemented by say, a GraphNode which uses the prepend method in some way (one might argue it to be a convoluted use case). For the sake of clarity and consistency with the semantic meaning of the example, I believe `ListNode[Bird]` would be a better choice.

### DIFF
--- a/_tour/lower-type-bounds.md
+++ b/_tour/lower-type-bounds.md
@@ -61,7 +61,7 @@ case class EuropeanSwallow() extends Bird
 
 
 val africanSwallowList = ListNode[AfricanSwallow](AfricanSwallow(), Nil())
-val birdList: Node[Bird] = africanSwallowList
+val birdList: ListNode[Bird] = africanSwallowList
 birdList.prepend(EuropeanSwallow())
 ```
-The `Node[Bird]` can be assigned the `africanSwallowList` but then accept `EuropeanSwallow`s.
+The `ListNode[Bird]` can be assigned the `africanSwallowList` but then accept `EuropeanSwallow`s.


som-snytt:
The section is demonstrating why you need a type parameter for the method and give it a bound.

It doesn't quite demonstrate that.

This shows that you can prepend a "disparate" swallow:
```
  val africanSwallowList: Node[AfricanSwallow] = ListNode[AfricanSwallow](AfricanSwallow(), Nil())
  val birdList: Node[Bird] = africanSwallowList
  val s0 = birdList.prepend(EuropeanSwallow())
  val s1 = africanSwallowList.prepend(EuropeanSwallow())
```
(This is that thing where `s0` is still `Node[Bird]` but `s1` is `Node[Product with Bird with Serializable]`.)

The last line of text is pointing out that the assignment to `birdList` works and `s0` also works. But I think it is more surprising that `s1` works.

Here is an example of why `Node` is useful, namely, you don't know you have a non-empty list until you prepend:
```
val swallowsFromAntarctica: Node[Bird] = Nil()
```
Normally you'd use pattern matching to inspect a `Node`. And now I see that `ListNode` is a `case class` but has extra members.

som-snytt:
Possibly superseded by https://github.com/scala/docs.scala-lang/pull/2355 which attempts to shake it out a bit.

som-snytt:
Expanding this recommendation at https://github.com/scala/docs.scala-lang/pull/2355 to avoid `Node` and say `List`.

PR #2349, opened 2022-03-19
Fix compiling warnings in a code example of tour/classes

TSLsun:
Hi,

While following the code examples in Classes (Tour of Scala) in IDE, I noticed some compile warnings as follows:
* Type annotation required for public member (`x` and `y`)
* Method with Unit result type is parameterless (`printWarning`)
* Type annotation required for Unit definition (`printWarning`)

I tried to fix them in this PR :)

Thanks!

### DIFF
--- a/_tour/classes.md
+++ b/_tour/classes.md
@@ -86,7 +86,7 @@ class Point {
     if (newValue < bound) _y = newValue else printWarning()
   }
 
-  private def printWarning(): Unit = println("WARNING: Out of bounds")
+  private def printWarning() = println("WARNING: Out of bounds")
 }
 
 val point1 = new Point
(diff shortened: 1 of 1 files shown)


_tour/classes.md
```diff
@@ -76,17 +76,17 @@ class Point {
...
 
-  def y = _y
+  def y: Int = _y
   def y_= (newValue: Int): Unit = {
-    if (newValue < bound) _y = newValue else printWarning
+    if (newValue < bound) _y = newValue else printWarning()
   }
 
-  private def printWarning = println("WARNING: Out of bounds")
+  private def printWarning(): Unit = println("WARNING: Out of bounds")
```
som-snytt, line 89:
```suggestion
  private def printWarning() = println("WARNING: Out of bounds")
```
Since it is private, it is more usual just to infer a trivial type.

som-snytt (approved):
Thanks!

PR #2351, opened 2022-03-21
update status of JDK 8

SethTisue:
* also change AdoptOpenJDK->Temurin
* also move some GraalVM text to a more appropriate location

note that I chose not to add coverage of `-release` and `-target` and such.  I feel like it would be out of scope for this page to go into detail about that. (a separate PR that added a brief reference and link of some kind to information on that would be welcome.)

### DIFF
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -26,13 +26,11 @@ Lightbend offers [commercial support](https://www.lightbend.com/lightbend-platfo
 
 ## Running versus compiling
 
-JDK 8, 11, and 17 are all good choices both for *compiling* and *running* Scala code.
+JDK 8, 11, and 17 are all reasonable choices both for *compiling* and *running* Scala code.
 
 Since the JVM is normally backwards compatible, it is usually safe to use a newer JVM for *running* your code than the one it was compiled on, especially if you are not using JVM features designated "experimental" or "unsafe".
 
-If you compile on JDK 11+ but have users on an older JDK version, additional care is needed to avoid using APIs and features that don't exist in 8. Therefore, *compiling* on 8 may be the safest choice in this scenario. (Some Scala developers use JDK 11 or 17 for their daily work, but do release builds on JDK 8.)
-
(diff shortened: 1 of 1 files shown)


som-snytt:
As you may not yet be aware, I spent a few hours with `-release` yesterday. It's so cool! Let us give thanks to retronym, rest his soul, I enjoyed reading his old email thread to the JVM folks.

Since there is no longer any "and such", but only `-release`, I think, "Consider using `-release:8`!" is warranted.

Also some of your periods could be exclamation marks, as in "New! and improved!"

It seems to me that you say "reasonable" when a more precise word might be "passable."

I'm reading Pickwick Papers, so soon I will be speaking only as Mr Pickwick does, or rather, as only he _can_.

som-snytt (approved):
Not disapproving.

julienrf (approved):

_overviews/jdk-compatibility/overview.md
```diff
@@ -26,13 +26,11 @@ Lightbend offers [commercial support](https://www.lightbend.com/lightbend-platfo
...
 
 Since the JVM is normally backwards compatible, it is usually safe to use a newer JVM for *running* your code than the one it was compiled on, especially if you are not using JVM features designated "experimental" or "unsafe".
 
-If you compile on JDK 11+ but have users on an older JDK version, additional care is needed to avoid using APIs and features that don't exist in 8. Therefore, *compiling* on 8 may be the safest choice in this scenario. (Some Scala developers use JDK 11 or 17 for their daily work, but do release builds on JDK 8.)
-
-Additionally, you can also run your Scala application on GraalVM which is a JVM. GraalVM performs well on the Scala benchmarks, and it benefits from GraalVM runtime and runs faster too.
+JDK 8 remains in use at many shops (as of early 2022), but usage is declining and some projects are dropping support. If you compile on JDK 11+ but want to allow your users to stay on 8, additional care is needed to avoid using APIs and features that don't exist in 8. (For this reason, some Scala developers use JDK 11 or 17 for their daily work but do release builds on JDK 8.)
```
ijuma, line 33: I understand that we intentionally avoided mentioning --release, but I think this wording is nearly misleading as it makes it sound like one has to do this manually.

SethTisue: I see what you mean. Would you like to PR a suggested wording change?

PR #2355, opened 2022-03-27
Lower bound in tour includes dangers

som-snytt:
Challenging to fit on a single tour slide, but tries to emphasize that there are two concepts in play: covariance and need for lower bound. The naming of the list example classes may or may not be a distraction.

### DIFF
--- a/_tour/lower-type-bounds.md
+++ b/_tour/lower-type-bounds.md
@@ -16,33 +16,30 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
 Here is an example where this is useful:
 
 ```scala mdoc:fail
-trait List[+B] {
-  def prepend(elem: B): NonEmptyList[B] = NonEmptyList(elem, this)
+trait List[+A] {
+  def prepend(elem: A): NonEmptyList[A] = NonEmptyList(elem, this)
 }
 
-case class NonEmptyList[+B](head: B, tail: List[B]) extends List[B]
+case class NonEmptyList[+A](head: A, tail: List[A]) extends List[A]
 
 object Nil extends List[Nothing]
 ```
 
-This program implements a singly-linked list. `Nil` represents an empty list with no elements. `class NonEmptyList` is a node which contains an element of type `B` (`head`) and a reference to the rest of the list (`tail`). The `trait List` and its subtypes are covariant because we have `+B`.
+This program implements a singly-linked list. `Nil` represents an empty list with no elements. `class NonEmptyList` is a node which contains an element of type `A` (`head`) and a reference to the rest of the list (`tail`). The `trait List` and its subtypes are covariant because we have `+A`.
 
-However, this program does _not_ compile because the parameter `elem` in `prepend` is of type `B`, which we declared *co*variant. This doesn't work because functions are *contra*variant in their parameter types and *co*variant in their result types.
+However, this program does _not_ compile because the parameter `elem` in `prepend` is of type `A`, which we declared *co*variant. This doesn't work because functions are *contra*variant in their parameter types and *co*variant in their result types.
 
-To fix this, we need to flip the variance of the type of the parameter `elem` in `prepend`. We do this by introducing a new type parameter `U` that has `B` as a lower type bound.
+To fix this, we need to flip the variance of the type of the parameter `elem` in `prepend`. We do this by introducing a new type parameter `B` that has `A` as a lower type bound.
 
 ```scala mdoc
-trait List[+B] {
-  def prepend[U >: B](elem: U): NonEmptyList[U] = NonEmptyList(elem, this)
+trait List[+A] {
+  def prepend[B >: A](elem: B): NonEmptyList[B] = NonEmptyList(elem, this)
 }
 
-case class NonEmptyList[+B](head: B, tail: List[B]) extends List[B]
+case class NonEmptyList[+A](head: A, tail: List[A]) extends List[A]
 
 object Nil extends List[Nothing]
 ```
-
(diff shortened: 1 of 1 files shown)


julienrf:
Thank you @som-snytt, however, I think it would be better to minimize the changes between the code snippet that does not contain a lower bound and the one that showcases a lower bound, to make it easier for the readers to understand the scope of type bounds. I suggest using `Nothing` and `head` and `tail` from the beginning, instead of doing this change here. What do you think?

som-snytt:
The suggestion on the linked PR was that the naming of "List" was potentially confusing.

If I'm allowed to muck with the entire example, is it OK to make it look like the familiar `List` and avoid the nerdy `Node` terminology?

I was trying avoid changing too much, in case there are pedagogical reasons.

_tour/lower-type-bounds.md
```diff
@@ -16,52 +16,56 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
...
-}
+case class NonEmptyList[+B](head: B, tail: List[B]) extends List[B]
 
-case class Nil[+B]() extends Node[B] {
-  def prepend[U >: B](elem: U): ListNode[U] = ListNode(elem, this)
-}
+object Nil extends List[Nothing]
 ```
 
+The type parameter for `List` is `B` to suggest we want to keep lists of birds.
```
julienrf, line 44: I am not sure about this part. Should we use `A` instead since it’s what people usually do?

_tour/lower-type-bounds.md
```diff
@@ -16,52 +16,56 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
...
-val birdList: Node[Bird] = africanSwallowList
-birdList.prepend(EuropeanSwallow())
+// assign swallows to birds
+val birds: List[Bird] = africanSwallows
+
+// add a swallow to birds
+val moreBirds = birds.prepend(EuropeanSwallow())
+
+// add disparate swallows together to get birds
+val allBirds = africanSwallows.prepend(EuropeanSwallow())
```
julienrf, line 62: That one might deserve more explanations because the type `U` is inferred to be the least upper bound of `EuropeanSwallow` and `AfricanSwallow`.

_tour/lower-type-bounds.md
```diff
@@ -16,52 +16,56 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
...
+
+// but this is a mistake! adding a list of birds widens the type arg too much. -Xlint will warn!
+val error = moreBirds.prepend(swallowsFromAntarctica)
```
julienrf, line 65: It would be useful here to mention what the inferred type is for `error`, I think.

_tour/lower-type-bounds.md
```diff
@@ -16,52 +16,56 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
...
+val swallowsFromAntarctica: List[Bird] = Nil
 
-val africanSwallowList = ListNode[AfricanSwallow](AfricanSwallow(), Nil())
-val birdList: Node[Bird] = africanSwallowList
-birdList.prepend(EuropeanSwallow())
+// assign swallows to birds
+val birds: List[Bird] = africanSwallows
+
+// add a swallow to birds
+val moreBirds = birds.prepend(EuropeanSwallow())
```
julienrf, line 59: Maybe before this we should show a simpler case (that does not involve a LUB computation):

~~~ scala
val someBird: Bird = EuropeanSwallow()
val moreBirds = africanSwallows.prepend(someBird)
~~~

That’s the simple case where the type of the resulting list is wider than the type of `africanSwallows` because `someBird` has be wider type than `AfricanSwallow`.

julienrf (commented):
Thank you! I have some suggestions.

_tour/lower-type-bounds.md
som-snytt: It's boring, but I'm OK with it. I needed to address why they used B.

_tour/lower-type-bounds.md
som-snytt: Unfortunately `Object`, which requires further explanation. If only it were just `Any`.

_tour/lower-type-bounds.md
som-snytt: It's only a tour and not a textbook, so I don't know, but OK if that seems simpler.

_tour/lower-type-bounds.md
```diff
@@ -16,52 +16,57 @@ While [upper type bounds](upper-type-bounds.html) limit a type to a subtype of a
...
+val someBirds = africanSwallows.prepend(someBird)
+
+// add a swallow to birds
+val moreBirds = birds.prepend(EuropeanSwallow())
+
+// add disparate swallows together, `B` is `Bird` because that is the supertype common to both swallows
+val allBirds = africanSwallows.prepend(EuropeanSwallow())
+
+// but this is a mistake! adding a list of birds widens the type arg too much. -Xlint will warn!
+val error = moreBirds.prepend(swallowsFromAntarctica)    // List[Object]
```
julienrf, line 66: I believe `AnyRef` is more scala-ish

```suggestion
val error = moreBirds.prepend(swallowsFromAntarctica)    // List[AnyRef]
```

julienrf: That’s a bug in the compiler then, I would expect the type `Object` to be shown as `AnyRef`. But this discussion is out of the scope of your PR :)

som-snytt:
@julienrf thanks for making it better

PR #2361, opened 2022-03-30
Update implicit-parameters.md

jpanku:
Recursive sum function missing last case iteration. Visible if we change unit implementation in StringMonoid for a string like for example "Empty String".


julienrf:
Is that case not redundant with the other cases? I mean, if the list contains exactly one element, the call to `sum(xs)` reduces to `m.add(xs.head, m.unit)`, which is equivalent to `xs.head` by the laws of monoids.

jpanku:
> 

Yes you are right. I totally missed that was a monoid, so the bug I see was assigning a random value to the neutral element of the monoid, and that totally breaks identity laws of monoids.

Pretty stup1d by me! xP.. thank you for your comment julienrf. :)

PR #2362, opened 2022-03-31
add instructions for viewing contents of .tasty files to Tasty Overview

olhotak:


julienrf (approved):

bishabosha:
Note that these instructions only work for the `scalac` directly downloaded from github, not by Coursier/`cs`

PR #2363, opened 2022-03-31
Add scastie, rename top links

bishabosha:
see https://github.com/scala/scala-lang/pull/1360

Screenshot 2022-03-31 at 11 32 02


julienrf (approved):

julienrf:
We should also apply the same change in `dotty`:

Screenshot from 2022-03-31 11-36-22

SethTisue (approved):

Issue #2369, opened 2022-04-07
Deprecated example for implicit type casting in "A First Look at Types"


gordcurrie:
When working through the examples I ran into some deprecated warnings specifically with the Section on [Type Casting](https://docs.scala-lang.org/scala3/book/first-look-at-types.html#type-casting)

```scala
val x: Long = 987654321
val y: Float = x  // 9.8765434E8 (note that some precision is lost in this case)
```

The warning is: 
```method long2float in object Long is deprecated since 2.13.1: Implicit conversion from Long to Float is dangerous because it loses precision. Write .toFloat instead.```

gordcurrie:
Closing, was resolved by https://github.com/scala/docs.scala-lang/pull/2370#issuecomment-1095095735

PR #2370, opened 2022-04-07
Update implicit casting example

gordcurrie:
Addresses the following issue https://github.com/scala/docs.scala-lang/issues/2369

Example of implicit casting uses depreciated example.

### DIFF
--- a/_overviews/scala3-book/first-look-at-types.md
+++ b/_overviews/scala3-book/first-look-at-types.md
@@ -249,7 +249,7 @@ You can only cast to a type if there is no loss of information. Otherwise, you n
 
 ```scala
 val x: Long = 987654321
-val y: Float = x.toFloat  // 9.8765434E8 (note that `.toFloat` is required because it results in percision loss)
+val y: Float = x.toFloat  // 9.8765434E8 (note that `.toFloat` is required because the cast results in percision loss)
 val z: Long = y  // Error
 ```
 
--- a/_overviews/scala3-book/first-look-at-types.md
+++ b/_overviews/scala3-book/first-look-at-types.md
@@ -239,7 +239,7 @@ For example:
 
 ```scala
 val x: Long = 987654321
-val y: Float = x  // 9.8765434E8 (note that some precision is lost in this case)
+val y: Float = x.toFloat  // 9.8765434E8 (note implicit conversion is depreciated since 2.13.1 when it would result in percision loss)
 
 val face: Char = '☺'
 val number: Int = face  // 9786
--- a/_overviews/scala3-book/first-look-at-types.md
+++ b/_overviews/scala3-book/first-look-at-types.md
@@ -246,7 +246,6 @@ val number: Int = face  // 9786
 ```
 
 You can only cast to a type if there is no loss of information. Otherwise, you need to be explicit about the cast:
-This will not compile:
 
 ```scala
 val x: Long = 987654321
--- a/_overviews/scala3-book/first-look-at-types.md
+++ b/_overviews/scala3-book/first-look-at-types.md
@@ -238,8 +238,8 @@ Value types can be cast in the following way:
 For example:
 
 ```scala
-val x: Long = 987654321
-val y: Float = x.toFloat  // 9.8765434E8 (note implicit conversion is depreciated since 2.13.1 when it would result in percision loss)
+val b: Byte = 127
+val i: Int = b  // 127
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/first-look-at-types.md
```diff
@@ -239,7 +239,7 @@ For example:
 
 ```scala
 val x: Long = 987654321
-val y: Float = x  // 9.8765434E8 (note that some precision is lost in this case)
+val y: Float = x.toFloat  // 9.8765434E8 (note implicit conversion is depreciated since 2.13.1 when it would result in percision loss)
```
julienrf, line 242: Thank you for the fix. There is typo in “deprec*i*ated” :)

However, I think this example should be moved below. It is better to first show the cast from `Char` to `Int`, which corresponds to a simple case of type widening.

We could maybe add another example (e.g. `Float` to `Double`?).

Then, there is this section below (from line 248 to 255) where we explain that casting to a narrower type does not work. I think we should move the `Long` to `Float` example here, and we should mention that it is also possible to cast from `Float` to `Long` by explicitly calling `toLong`.

gordcurrie: Great suggestions and feedback, thank you so much, and great catch on the typo.

I've updated, taking your feedback into consideration. I opted for a Byte -> Int example vs the suggested Float -> Double to avoid any confusion the increased precision might cause.
```scala
val f: Float = 3.14
val d: Double = f // 3.140000104904175
```

_overviews/scala3-book/first-look-at-types.md
```diff
@@ -248,9 +248,9 @@ val number: Int = face  // 9786
 Casting is unidirectional.
```
julienrf, line 248:
```suggestion
You can only cast to a type if there is no loss of information. Otherwise, you need to be explicit about the cast:
```

_overviews/scala3-book/first-look-at-types.md
```diff
@@ -248,9 +248,9 @@ val number: Int = face  // 9786
...
 This will not compile:
 
-```
+```scala
 val x: Long = 987654321
-val y: Float = x  // 9.8765434E8
+val y: Float = x.toFloat  // 9.8765434E8 (note implicit conversion is deprecated since 2.13.1, when it would result in percision loss)
```
julienrf, line 253:
```suggestion
val y: Float = x.toFloat  // 9.8765434E8 (note that `.toFloat` is required because it results in percision loss)
```

_overviews/scala3-book/first-look-at-types.md
```diff
@@ -248,9 +248,9 @@ val number: Int = face  // 9786
...
 val z: Long = y  // Does not conform
```
julienrf, line 254:
```suggestion
val z: Long = y  // Error
```

julienrf (commented):
Hey @gordcurrie, thank you for the changes, I have left additional suggestions, let me know what you think about them.

gordcurrie:
Thanks again for your suggestions, Julien. I have incorporated them in this latest commit.

_overviews/scala3-book/first-look-at-types.md
```diff
@@ -238,20 +238,20 @@ Value types can be cast in the following way:
...
+val b: Byte = 127
+val i: Int = b  // 127
 
 val face: Char = '☺'
 val number: Int = face  // 9786
 ```
 
-Casting is unidirectional.
+You can only cast to a type if there is no loss of information. Otherwise, you need to be explicit about the cast:
 This will not compile:
```
julienrf, line 249:
```suggestion
```

_overviews/scala3-book/first-look-at-types.md
```diff
@@ -238,20 +238,19 @@ Value types can be cast in the following way:
...
-Casting is unidirectional.
-This will not compile:
+You can only cast to a type if there is no loss of information. Otherwise, you need to be explicit about the cast:
 
-```
+```scala
 val x: Long = 987654321
-val y: Float = x  // 9.8765434E8
-val z: Long = y  // Does not conform
+val y: Float = x.toFloat  // 9.8765434E8 (note that `.toFloat` is required because it results in percision loss)
```
julienrf, line 252:
```suggestion
val y: Float = x.toFloat  // 9.8765434E8 (note that `.toFloat` is required because the cast results in percision loss)
```

PR #2371, opened 2022-04-09
Fix error in `given` definition

artemkorsakov:
We can see in the [reference](https://docs.scala-lang.org/scala3/reference/contextual/givens.html) that the right `given` definition is:
```
given intMonthConverter: MonthConverter[Int] with
```


julienrf (approved):
Good catch, thank you!

PR #2380, opened 2022-04-19
Display code blocks inside details/summary correctly

### LINKED ISSUE #2379: Code block within details disclosure element are not rendered correctly
On https://docs.scala-lang.org/scala3/guides/macros/inline.html there is the following details disclosure element which when expanded contains a code block that is not rendered correctly:

Screen Shot 2022-04-19 at 10 08 30 AM

Philippus:
Fixes #2379.

### DIFF
--- a/_sass/layout/details-summary.scss
+++ b/_sass/layout/details-summary.scss
@@ -1,0 +1,3 @@
+details > summary > p {
+  display: inline;
+}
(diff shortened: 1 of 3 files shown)


_sass/layout/details-summary.scss
```diff
@@ -0,0 +1,3 @@
+details > summary > p {
+  display: inline;
```
Philippus, line 2: This is to make sure the little arrow is on the same line as "See inlining steps".

bishabosha (approved):

PR #2383, opened 2022-04-21
Avoid dysphemisms

erikvanoosten:
I was reading the inclusive language guide at https://docs.scala-lang.org/contribute/inclusive-language-guide.html. I have 2 points of feedback. I initially posted this on Discord but then I noticed I could do a pull request. So here it is.

Re. blacklist/whitelist. Such lists are often used to regulate access. Therefore I propose we suggest the alternative 'allowlist/denylist' as it is probably appropriate in more circumstances. This is also the suggestion done from the linked page https://github.com/dialpad/inclusive-language.

### DIFF
--- a/_overviews/contribute/inclusive-language-guide.md
+++ b/_overviews/contribute/inclusive-language-guide.md
@@ -90,7 +90,7 @@ This list is neither comprehensive nor definitive, and it can evolve over time.
 * **blacklist/whitelist** \
   While the etymology of these words has no relation to racism, their use suggests an association between the color black and some form of badness or exclusion, and between the color white and some form of goodness or inclusion.
   Prefer alternatives when possible.
-  Several alternatives have been proposed but none sticks as "the one". We suggest using the pair *allowlist*/*denylist* or the pair *include*/*exclude*, as these are generic enough to replace most uses of *blacklist*/*whitelist*.
+  Several alternatives have been proposed but none sticks as "the one". We suggest using the pair *denylist*/*allowlist* or the pair *excludelist*/*includelist*, as these are generic enough to replace most uses of *blacklist*/*whitelist*.
 * **master/slave** \
   Never use *slave*.
   Never use *master* in conjunction with *slave*.
--- a/_overviews/contribute/inclusive-language-guide.md
+++ b/_overviews/contribute/inclusive-language-guide.md
@@ -90,7 +90,7 @@ This list is neither comprehensive nor definitive, and it can evolve over time.
 * **blacklist/whitelist** \
   While the etymology of these words has no relation to racism, their use suggests an association between the color black and some form of badness or exclusion, and between the color white and some form of goodness or inclusion.
   Prefer alternatives when possible.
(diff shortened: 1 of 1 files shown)


SethTisue:
ping @sjrd

SethTisue:
perhaps both "include/exclude" and "allow/deny" should be mentioned?

I do agree that "allow/deny" is what I would normally reach for first.

_overviews/contribute/inclusive-language-guide.md
```diff
@@ -116,6 +116,12 @@ We encourage you to research inclusive language guidelines applicable to your do
 
 You may want to use automated software like [In Solidarity](https://github.com/apps/in-solidarity) to steer contributors away from loaded words.
 
+## Dysphemism
+
+Dysphemisms, the opposite of euphemisms, should be avoided. For example, the English expressions "pull the trigger" (enforce a decision) and "bite the bullet" (endure hardship) can be disturbingly violent if you are not used to them.
+
+Using the direct meaning is preferred.
```
sjrd, line 123: To follow the writing style of the rest of the document, this should read as: (also one Markdown line per sentence, for better diffs later on)
```suggestion
Dysphemisms, the opposite of euphemisms, can be disturbingly violent if you are not used to them.
Examples include the English expressions "pull the trigger" (enforce a decision) and "bite the bullet" (endure hardship).
Prefer the direct meaning instead.
```
i.e.:

1. Problem
2. Examples
3. Propose a solution

with as much active voice as possible.

sjrd (commented):
I have reviewed the *form* of the contribution.

But I don't really have the liberty to make a review of the *content*. The content was discussed and approved by quite a number of stakeholders. I don't know how we should proceed with changes here.

/cc @darjutak @ValeriePe

erikvanoosten:
Thanks for the review Sébastien.

> But I don't really have the liberty to make a review of the content. The content was discussed and approved by quite a number of stakeholders. I don't know how we should proceed with changes here.

I thought I read somewhere feedback was requested. But that must have been my imagination because I cannot find such a request anywhere.

Anyways, the Scala Center could collect feedback for a couple of months and then use the same process as was used to create this document the first time.

If however you think the value is not worth the effort, I propose we just close this PR and pretend nothing happened.

SethTisue:
>I thought I read somewhere feedback was requested

Might have been me in a chat room :-)

Personally I think the adjustments in this PR are modest and uncontroversial enough to be mergeable, especially since the document as a whole is a collection of suggestions and recommendations rather than anything that's actually binding on anybody.

except that I do think that this should be adjusted:

> both "include/exclude" and "allow/deny" should be mentioned

partly because the include/exclude suggestion already went through this external-review process.

Regardless, whatever the Center decides will of course be fine with me.

erikvanoosten:
> except that I do think that this should be adjusted:
> 
> > both "include/exclude" and "allow/deny" should be mentioned

Done.

SethTisue (approved):
I'm hitting "approve", but as per discussion, someone from the Center needs to make the final call here

sjrd (approved):
We discussed this in a team-wide meeting at the Scala Center, and we collectively approve these changes.

Thank you for your contribution. :)

PR #2388, opened 2022-04-29
Put OS-specific get-started instructions in tabs

bishabosha:
Screenshot 2022-04-29 at 18 29 22

This is also integrated with all the translations of the page. As with the other PR, JS being disabled will show the "Other" tab by default, and expand the "test your setup" accordion.

sibling PR https://github.com/scala/scala-lang/pull/1373


bishabosha:
added some changes to reuse templates more

_install_tabs/4-other.md
```diff
@@ -0,0 +1,18 @@
...
+language: en
+---
+<div class="scala-in-action-content">
+  <div class="scala-in-action-code">
+    <div class="wrap">
+      <div class="scala-text scala-text-large">
+        <noscript>
+          <p><span style="font-style:italic;">JavaScript is disabled, showing the default options.</span></p>
+        </noscript>
+        <p>Follow the documentation from Coursier on <a href="https://get-coursier.io/docs/cli-installation" target="_blank">how to install and run <code class="bash">cs setup</code></a>.</p>
```
julienrf, line 14: I suggest making it clearer that users have to first install `cs`, and then run `cs setup`.

```suggestion
        <p>Follow the documentation from Coursier on <a href="https://get-coursier.io/docs/cli-installation" target="_blank">how to install <code>cs</code></a>, and then <a href="https://get-coursier.io/docs/cli-setup" target="_blank">run <code>cs setup</code></a>.</p>
```

What does `class="bash"` do? I see no changes when I remove it.

_sass/layout/alt-details.scss
```diff
@@ -0,0 +1,46 @@
+// ALT-DETAILS
+//------------------------------------------------
+//------------------------------------------------
+
+.alt-details {
```
julienrf, line 5: In general, I think it is better to follow a “utility-first CSS classes” approach, which guarantees that the look and feel is consistent throughout the whole website, and makes the CSS easier to maintain. Here, you introduce new “semantic classes” with specific colors and styles, which are not well integrated with the rest of the website.

You can find more details about “utility-first CSS classes” here: https://adamwathan.me/css-utility-classes-and-separation-of-concerns/

Could you try to make the style of the added content more consistent with the existing content?

resources/js/functions.js
```diff
@@ -392,15 +392,55 @@ function getOS() {
   return osname;
 }
 
+function toggleElement(evt, elemId) {
```
julienrf, line 395: I don’t have a strong opinion here, but isn’t it risky to introduce several top-level names like this one? The risk I see is name clashes, or cross-site code injections that would override the meaning of those names.

resources/js/functions.js
```diff
@@ -392,15 +392,55 @@ function getOS() {
...
+  if (elem.style.display === "none") {
+    elem.style.display = "block";
+  } else {
+    elem.style.display = "none";
+  }
+  $(btn).toggleClass("alt-details-closed");
+}
+
+$(document).ready(function() {
+  $('.alt-details-toggle').click();
```
julienrf, line 407: Would it be possible to remove this line? Change the CSS so that by default this element is correctly displayed (or hidden).

resources/js/functions.js
```diff
@@ -392,15 +392,55 @@ function getOS() {
...
+
+  // Show the current tab, and add an "active" class to the button that opened the tab
+  document.getElementById(category + '-' + tabName).style.display = "block";
+  queried.className += " active";
+}
+
 $(document).ready(function() {
+  var defaultTabs = document.getElementsByClassName('default-tab');
+  for (i = 0; i < defaultTabs.length; i++) {
+    defaultTabs[i].click();
```
julienrf, line 433: Same here

julienrf (changes requested):
Thank you Jamie, this is a great improvement. I have tested it locally and the OS detection did work correctly. I think the proposed changes work well, but the implementation details may lead to complex stylesheets, and the choice of colors and styles is not consistent with the styles used everywhere else in the website.

_install_tabs/4-other.md
bishabosha: hljs automatically injects its own classes into `<code>` inside of a `<pre>`, so I provide the language explicitly so it does not do that. Also, the instructions on the coursier website tell you to run `cs setup`, not only how to install `cs`

julienrf: OK for the highlighting, and ok for `cs setup`, you’re right!

resources/js/functions.js
bishabosha: is the suggestion here to make less generic names? - I was mostly following the convention for the whole js file

resources/js/functions.js
julienrf: I understand now that this is needed to handle visitors with disabled JS

resources/js/functions.js
bishabosha: so next step would be to instead of use the `onclick` attribute in html, add event listeners when the document loads

bishabosha: done

_sass/layout/alt-details.scss
bishabosha: done

bishabosha:
new look:
Screenshot 2022-05-10 at 15 53 28

julienrf (approved):
Very nice, thank you Jamie!

PR #2390, opened 2022-05-04
Make sure SIP numbers are used exactly once, and proposals use exactly one number

julienrf:
In #2389 I assigned the number 34 to a SIP although that number was already used by another one. I noticed that a couple of other proposals were using the number 36, and that another proposal was using two numbers.

I propose these changes so that numbers are used at most once, and proposals use at most one number.

### DIFF
--- a/_sips/sips/2017-11-20-byname-implicits.md
+++ b/_sips/sips/2017-11-20-byname-implicits.md
@@ -1,6 +1,6 @@
 ---
 layout: sip
-title: SIP-36 - Byname implicit arguments
+title: SIP-31 - Byname implicit arguments
 vote-status: complete
 vote-text: This SIP has been implemented in Scala 2.13.0 and Scala 3.0.0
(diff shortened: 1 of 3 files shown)


_sips/sips/2017-11-20-byname-implicits.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: sip
-title: SIP-36 - Byname implicit arguments
+title: SIP-31 - Byname implicit arguments
```
sjrd, line 3: Wasn't there any 31 before? :thinking:

sjrd (approved):

_sips/sips/2017-11-20-byname-implicits.md
julienrf: I couldn’t find it.

`$ grep -r "SIP-31" _sips/sips/` returned no match.

PR #2112, opened 2021-07-07
Copy content from dotty.epfl.ch into docs.scala-lang.org

julienrf:
This PR is huge because a lot of content has been copied. To ease the review, I suggest looking at commits individually.

This is the first step towards moving the content of https://dotty.epfl.ch into https://docs.scala-lang.org. The next steps will be to move the remaining content to https://scala-lang.org, and to deprecate the website https://dotty.epfl.ch.

The website https://dotty.epfl.ch contains several pages. Each commit of this PR handles one page, or one section of the Dotty website.

---

# Landing page

https://dotty.epfl.ch

The content is already covered by https://www.scala-lang.org/download/scala3.html

# Blog

https://dotty.epfl.ch/blog/index.html

I didn’t move the blog content. I just suggest the next blog posts be published on https://scala-lang.org/blog.

# Usage

## Getting Started

https://dotty.epfl.ch/docs/usage/getting-started.html

Copied some of the content in our Scala 3 book. See 9ffe4b9.

## Using Dotty wth sbt

https://dotty.epfl.ch/docs/usage/sbt-projects.html

Empty page, no need to merge

## IDE support for Scala 3

https://dotty.epfl.ch/docs/usage/ide-support.html

Added a mention of BSP in our getting started instructions. See 9b8c42b.

## Worksheet mode with Dotty IDE

https://dotty.epfl.ch/docs/usage/worksheet-mode.html

Introduce a page about worksheets in the Scala 3 book. See 4567802.

## Language versions

https://dotty.epfl.ch/docs/usage/language-versions.html

Created a page in the Scala 3 reference. See 8c1ff35.

## Using Dotty with cbt

https://dotty.epfl.ch/docs/usage/cbt-projects.html

Skipped.

# Reference

https://dotty.epfl.ch/docs/reference/overview.html

Copied to https://docs.scala-lang/scala3/reference/overview.html. See d947163.

# Contributing

Skipped. I think we should eventually move that content into the code repository itself.

# Internals

Skipped. I think we should eventually move that content into the code repository itself.

# Resources

https://dotty.epfl.ch/docs/resources/talks.html

Created a page https://docs.scala-lang.org/scala3/talks.html. See a600b4b.

# API

https://dotty.epfl.ch/api/index.html

Skipped. Will be merged into https://scala-lang.org/api instead.

---

Some screenshots:

Screenshot 2021-07-07 at 17-07-01 Overview

Screenshot from 2021-07-07 17-06-39

PR #2391, opened 2022-05-05
Remove source files in _scala3-reference/

julienrf:
Initially, the Dotty documentation used to be published on the website https://dotty.epfl.ch. When Dotty became Scala 3.0.0, it was natural to move its documentation here, on the website https://docs.scala-lang.org (see #2112). However, by doing so we realized that the compiler team lost the ability to implement something in the compiler and to document it in a single pull request. They now had to implement their features in the dotty repo, and to document these features here.

I believe the Scala 3 reference documentation should be hosted on the website docs.scala-lang.org, but I also understand that it is more convenient for the compiler team if they can implement things and document their work within a single repo. Consequently, we decided with VirtusLab to produce the HTML pages of the Scala 3 reference documentation from source files hosted on the dotty repo. These pages are published at the URL https://docs.scala-lang.org/scala3/reference, so they “shadow” any content produced by the docs.scala-lang repo. Therefore, we should remove the duplicated source files from this repo.


bishabosha:
some links need updating

bishabosha:
one more
```
Ran on 1090 files!
[30](https://github.com/scala/docs.scala-lang/runs/6307797627?check_suite_focus=true#step:7:30)
- ./_site/scala3/book/ca-summary.html
[31](https://github.com/scala/docs.scala-lang/runs/6307797627?check_suite_focus=true#step:7:31)
  *  External link https://docs.scala-lang.org/scala3/reference/contextual.html failed: 404 No error
```

Issue #2396, opened 2022-05-16
[CSS] Misaligned page content


lzm0:
![image](https://user-images.githubusercontent.com/7163127/168620750-56d42ec0-93f5-4fa6-a1f0-3ca283c8a144.png)
In Safari it looks fine.
Chrome version: 101.0.4951.64

lzm0:
@bishabosha FYI

PR #2397, opened 2022-05-16
Fix misaligned page content

lzm0:
Fix https://github.com/scala/docs.scala-lang/issues/2396
Before:
image

After:
image


bishabosha (approved):
Thanks a lot!
tested on Safari Firefox and Chrome

PR #2405, opened 2022-05-25
write tabs inline in markdown

bishabosha:
we can now write tabs inline in a markdown file, this makes it more simple to use


julienrf (approved):
Thank you Jamie. I’ve tested locally and the tabs still work on Firefox/Linux.

PR #2409, opened 2022-05-27
scroll tabs on smaller screens

### LINKED ISSUE #2408: tabs do not scale well for small devices
when we have a small device, the tabs can expand beyond the screen width - perhaps we should make the tabs scrollable when the width is too small, otherwise I could imagine turning the tabs into accordions when width is narrow
Screenshot 2022-05-27 at 09 50 38

bishabosha:
fixes #2408

### DIFF
--- a/_sass/components/tab.scss
+++ b/_sass/components/tab.scss
@@ -9,6 +9,8 @@
     @include display(flex);
     @include align-items(center);
     @include justify-content(flex-start);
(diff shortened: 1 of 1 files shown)


julienrf:
May I ask you to publish a screenshot?

bishabosha:
@julienrf here is what it looks like: (on my browser scroll bars only appear when you start scrolling)
Screenshot 2022-05-31 at 09 49 05

PR #2414, opened 2022-06-01
Improve getting-started experience for Java newcomers and OSS contributors

anatoliykmetyuk:


index.md
```diff
@@ -22,6 +22,10 @@ scala3-sections:
         description: "Learn Scala by reading a series of short lessons."
         icon: "fa fa-book-open"
         link: /scala3/book/introduction.html
+      - title: "Scala for Java Programmers"
```
bishabosha, line 25: we already have an article with this title at https://docs.scala-lang.org/tutorials/scala-for-java-programmers.html, this could be confusing

julienrf (changes requested):
Thank you Anatolii!

In general, I think we should try to minimize the number of similar content. Here, I see that your contributions have a unique value, but they also overlap a bit with some existing content. I suggest that we try to incorporate your contributions to the existing content instead. Would that work for you?

index.md
```diff
@@ -56,6 +60,10 @@ scala3-sections:
         description: "MOOCs to learn Scala, for beginners and experienced programmers."
         icon: "fa fa-cloud"
         link: /online-courses.html
+      - title: "Becoming a Scala OSS Contributor"
+        description: "How to start contributing to open-source Scala ecosystem"
+        icon: "fa fa-code-branch"
+        link: /scala3/newcomers/contributor.html
```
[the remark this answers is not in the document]
bishabosha: that guide is specifically about contributing to scala itself, rather than a community project, so I think it would need a lot of reworking to "contextualise" it for general contributing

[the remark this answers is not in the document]
julienrf: I think the index page is more general than that, see e.g. https://docs.scala-lang.org/contribute/#tools-and-libraries which is specifically about community libraries.

bishabosha:
@julienrf This is now ready for second review

_overviews/tutorials/scala-for-java-programmers.md
```diff
@@ -6,19 +6,79 @@ partof: scala-for-java-programmers
...
-This document gives a quick introduction to the Scala language and
-compiler. It is intended for people who already have some programming
-experience and want an overview of what they can do with Scala. A
-basic knowledge of object-oriented programming, especially in Java, is
-assumed.
+### Explore Further
 
-## A First Example
+When you finish these guides, we recommend to continue your Scala journey by reading the
+*Scala Book* or following a number of *online MOOCs*.
```
julienrf, line 75: These should probably be links to the corresponding pages (Scala book and online courses)

_plugins/modal-details-lib/modal-details.rb
```diff
@@ -0,0 +1,38 @@
+require 'erb'
+
+module Jekyll
+    module ModalDetails
```
julienrf, line 4: Is this one still used?

julienrf (changes requested):
Thank you very much Jamie, this is a great improvement.

_plugins/modal-details-lib/modal-details.rb
bishabosha: it has not been used but I thought why not add it still - its isolated to the commit https://github.com/scala/docs.scala-lang/pull/2414/commits/d0b1da04f7717e89995d82060be20a334f0fd839

bishabosha: I dropped it

index.md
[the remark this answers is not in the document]
bishabosha: done these suggestions

_overviews/tutorials/scala-for-java-programmers.md
bishabosha: these are immediately followed by the relevant links - so I do not think it makes sense:

Screenshot 2022-08-23 at 16 05 07

bishabosha: perhaps it would not be confusing to link twice

julienrf (approved):

PR #2420, opened 2022-06-08
add Scala 2.12.16

SethTisue:


SethTisue:
I should make sure the CI failure is resolved because that will mean the API docs have published

PR #2423, opened 2022-06-08
Unified the zh-cn translation of case class

### LINKED ISSUE #2422: Unified the zh-cn translation of case class
Translate case class to "样例类", which is the same as the official translation in the zh-cn version of Programing in Scala (both 4th / 5th Edition).

byyue:
Close #2422.


jxnu-liguobin:
LGTM

PR #2425, opened 2022-06-08
Indentation for scala 2 reference made consistent as others

Shorla:
The other examples made use of 2 indentation, only the reference to scala 2 example used four indentation.


bishabosha (approved):
Thanks!

PR #2426, opened 2022-06-08
Fixed error in detailed inlining steps

ThinkDumbIndustries:


_overviews/scala3-macros/tutorial/inline.md
```diff
@@ -276,8 +276,8 @@ val x = 2
 val x2 = x * x
 x2 * {
   if (0 == 0) 1.0
-  else if (0 % 2 == 1) x * power(x, 0 - 1) // dead branch
-  else power(x * x, 0 / 2) // dead branch
+  else if (0 % 2 == 1) x * power(x2, 0 - 1) // dead branch
```
julienrf, line 279:
```suggestion
  else if (0 % 2 == 1) x2 * power(x2, 0 - 1) // dead branch
```

julienrf (commented):
Good catch, thank you! There was still an occurrence of `x` that had be renamed to `x2`.

PR #2428, opened 2022-06-10
Document Vprofile Options

odersky:


julienrf (approved):

julienrf:
I would have expected this PR to have been merged after https://github.com/lampepfl/dotty/pull/15406 is available (so, after 3.2.0 is released). But it’s OK, I know it’s complicated to deal with two separate repos… As you noted in one of the comments of the PR, we may need to work on a mechanism to _generate_ documentation from the compiler options (so that we don’t have to maintain it manually).

PR #2431, opened 2022-06-12
Clarify case _, case x

### LINKED ISSUE #2313: catch-all-with-variable trick needs some clarification
I suggest adding a note about control structures-match-using default value

+ Using variable bindings, you have to keep in mind, that any form of catch-all effectively breaks the matching chain. Some systems provide lookup for the most specific matches, but Scala's match does not. This error is easier to catch in the case of underscore placeholder but can be easily missed for var bindings. Especially if you suppress or ignore (not dive in) compiler warnings. This notch is a regular source of bad mistakes:

```
val i = 5
val day = i match
  case 0 => "Sunday"
  case 1 => "Monday"
  case 2 => "Tuesday"
  case 3 => "Wednesday"
  case what => "Guesswho"
  case 4 => "Thursday"
  case 5 => "Friday"
  case 6 => "Saturday"

println(day)

Guesswho
```

som-snytt:
More introductory words explaining `case _ =>` must come last, and naming in `case x`.

Fixes #2313

### DIFF
--- a/_overviews/scala3-book/control-structures.md
+++ b/_overviews/scala3-book/control-structures.md
@@ -369,9 +369,11 @@ val day = i match
   case _ => "invalid day"   // the default, catch-all
 ```
 
-In this example the variable `i` is tested against the cases shown.
-If it’s between `0` and `6`, `day` is bound to a string that represents one of the days of the week.
-Otherwise, the catch-all case is represented by the `_` character, and `day` is bound to the string, `"invalid day"`.
+In this example, the variable `i` is tested against the cases shown.
+If it’s between `0` and `6`, `day` is bound to the string that represents that day of the week.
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/control-structures.md
```diff
@@ -369,28 +369,37 @@ val day = i match
...
-You can also use `_` as a name to ignore the value.
-
+```scala
+val N = 42
+i match
+  case 0 => println("1")
+  case 1 => println("2")
+  case N => println("42")
+  case n => println(s"You gave me: $n" )
+```
```
julienrf, line 402: I suggest adding a comment saying that here if the value of `i` is 42, it prints “42”, not “You gave me: 42”.

julienrf (approved):
Thank you @som-snytt, this is a nice improvement!

julienrf:
Thank you!

PR #2432, opened 2022-06-14
Typo fix

SewonYun:


SethTisue:
was hoping for review from @scala/docs-zh, since it didn't come, I'll just assume this is a good change :-)

PR #2437, opened 2022-06-18
Update settings.md

artemkorsakov:
`-social-links` parameter is displayed with errors:
https://docs.scala-lang.org/scala3/guides/scaladoc/settings.html

image

Maybe we can replace ' to `?


SethTisue:
Thank you!

PR #2440, opened 2022-06-21
Fix a typo in scala3 book

berberman:


Philippus (approved):

_overviews/scala3-book/packaging-imports.md
```diff
@@ -211,7 +211,7 @@ package foo
 
 class ClassA:
   import scala.util.Random   // inside ClassA
-  def printRandom {
+  def printRandom:
```
julienrf, line 214:
```suggestion
  def printRandom(): Unit =
```

julienrf (changes requested):
The change you proposed is incorrect. Please see my suggestion for an alternative.

_overviews/scala3-book/packaging-imports.md
griggt: Looks like line 202 needs this change as well.

julienrf (approved):
Thank you for the fix!

PR #2441, opened 2022-06-22
Update methods-main-methods.md

AhsanShuvo:
Added println(). without this console doesn't print anything. a newcomer like me might be confused.


julienrf (approved):
That’s a good improvement, thank you!

Issue #2266, opened 2021-12-06
add explanation in book that indentation is significant


A-Roso:
There is no explanation in the book that indentation is significant and needed for the compiler. When searching for 'Indentation' it is only referenced in the Style Guide, Migration Guide and Language Reference.
I think for new learners that start learning Scala with Scala3 and the Scala3 book, it would be very helpful to add a few lines of explanation e.g. in A Taste of Scala - Control Structures or even a chapter of its own, that indentation is significant and must be adhered to.

SethTisue:
@alvinj

alvinj:
Hi @SethTisue. Unfortunately I have to focus on paid work at the moment, but yes, if we didn’t mention that, it should be added. I’ll be glad to do this as part of some volunteer work again next year, but honestly, it won’t be until somewhere around March-June.

SethTisue:
maybe someone else would like to pick this up.

Shorla:
Hi @SethTisue, I am new here but would love to contribute to scala docs. can I be assigned to this issue?

SethTisue:
sure, go for it!

Shorla:
should i move the explanation here:https://docs.scala-lang.org/style/indentation.html to scala book?

bishabosha:
> should i move the explanation here:https://docs.scala-lang.org/style/indentation.html to scala book?

this explanation is for Scala 2, where this indentation is only a style preference - also it uses curly braces, which in Scala 3. does not turn on the new indentation rules.

Here are the official rules for Scala 3 (but they are not very beginner friendly) https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html

alvinj:
This is actually a pretty fundamental thing that I missed in the book. I agree with the initial comment that it could go in the [Control Structures](https://docs.scala-lang.org/scala3/book/taste-control-structures.html) section, since that’s where it’s first shown, or just prior to that section.

FWIW, I also see on that Control Structures page that I left the Scala 2 example indented with four spaces, which is not consistent with the rest of the page.

PR #2445, opened 2022-06-23
indentaion explaination for scala 3 book

Shorla:
This solves the issue at #2266


_overviews/scala3-book/taste-indentation.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Indentation
+type: section
+description: This section demonstrates Scala 3 Indentation.
```
julienrf, line 4: I don’t think it’s correct to say that we “demonstrate” indentation. What you could say, instead is that this section shows that some lexical scopes can be delimited by indentation instead of braces.

_overviews/scala3-book/taste-indentation.md
```diff
@@ -0,0 +1,98 @@
...
+num: 8
+previous-page: taste-vars-data-types
+next-page: taste-control-structures
+---
+
+Indentation refers to the increase or decrease of space between the left and right margin of a paragraph.
+In some programming languages, tab key is used to make indentations. This is not the case in Scala. Tab key automatically creates four spaces which Scala doesn't support.
```
julienrf, line 11: This does not seem to be true, according to the language reference: https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html#spaces-vs-tabs

_overviews/scala3-book/taste-indentation.md
```diff
@@ -0,0 +1,98 @@
...
+In Scala each level of indentation is 2 spaces.  Thus, instead of
```
julienrf, line 12: Actually, this is not a requirement, but you could say that this is common practice.

julienrf (changes requested):
Thank you for opening a PR @Shorla.

I am a bit confused, however, because I thought you would introduce the fact that the indentation is significant in Scala 3, but you mostly talk about code style best practices. I would expect to see a discussion about braces being optional, as explained here: https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html

bishabosha:
Hi @Shorla, thank you for opening this PR. I noticed that you marked @julienrf comments as resolved, do you have some extra commits to push? The points about updating the metadata in other markdown files is important otherwise the documentation will not render correctly.

Shorla:
@bishabosha I made the adjustments for the metadata, but I'm yet to push it because I'm trying to conclude on the article. But I should push the full changes before the end of the week.

bishabosha:
> I made the adjustments for the metadata, but I'm yet to push it because I'm trying to conclude on the article. But I should push the full changes before the end of the week.

Ok great :), let us know if you want any more reviews

Shorla:
hi @bishabosha @julienrf, I couldn't add all of the information on optional braces. I added a part then a link to the Scala 3 reference article.

_overviews/scala3-book/taste-indentation-significant-syntax.md
```diff
@@ -0,0 +1,189 @@
...
+      3.1415)
+
+This way, all parameters line up, and you don't need to re-align them if
+you change the name of the method later on.
+
+Great care should be taken to avoid these sorts of invocations well into
+the length of the line. More specifically, such an invocation should be
+avoided when each parameter would have to be indented more than 50
+spaces to achieve alignment. In such cases, the invocation itself should
+be moved to the next line and indented two spaces:
```
som-snytt, line 84: This happens often. It's very annoying. There is no good solution except scalafmt, which I don't use, but I'd like to see the config that will format to some standard, and also see an example. I'm not likely to memorize rules.

_overviews/scala3-book/taste-indentation-significant-syntax.md
```diff
@@ -0,0 +1,189 @@
...
+   If an `<outdent>` is inserted, the top element is popped from `IW`.
+   If the indentation width of the token on the next line is still less than the new current indentation width, step (2) repeats. Therefore, several `<outdent>` tokens
+   may be inserted in a row.
+
+   The following two additional rules support parsing of legacy code with ad-hoc layout. They might be withdrawn in future language versions:
+
+    - An `<outdent>` is also inserted if the next token following a statement sequence starting with an `<indent>` closes an indentation region, i.e. is one of `then`, `else`, `do`, `catch`, `finally`, `yield`, `}`, `)`, `]` or `case`.
+
+    - An `<outdent>` is finally inserted in front of a comma that follows a statement sequence starting with an `<indent>` if the indented region is itself enclosed in parentheses.
+
```
som-snytt, line 158: The preceding is more like spec. I can't review that ATM, but I think some guidance about "when do I need to think about indentation" would be worthwhile. That is, normally, it should just work. When does it not just work?

som-snytt (changes requested):
Some structural notes.

som-snytt:
There is "common style" now.

I wonder if @SethTisue has an idea about addressing stale contributions.

Issue #2417, opened 2022-06-03
scalafiddle.io down?


eugenevankampen:
In the Basics tour there is a reference to https://scalafiddle.io. This website - today - returns a Error 522. Not sure if I should log it here as an issue. If not, please ignore.
![image](https://user-images.githubusercontent.com/106658852/171847453-72fd9308-dfa2-48d7-80bb-8ce69218f74a.png)

slee500:
Same for me, I raised a PR to remove references to it. See https://github.com/scala/docs.scala-lang/pull/2446

![image](https://user-images.githubusercontent.com/14228817/175500945-c9b53151-748f-4923-ad32-5f9b498684b8.png)

PR #2446, opened 2022-06-24
remove scalafiddle.io reference

slee500:
https://downinspector.com/check/scalafiddle.io 

Looks like scalafiddle.io has been down for a long time. So removing references to it.


sjrd (approved):
Unfortunately, that's probably what we have to do, indeed.

_tour/basics.md
```diff
@@ -16,7 +16,7 @@ In this page, we will cover the basics of Scala.
 
 You can run Scala in your browser with _ScalaFiddle_. This is an easy, zero-setup way to experiment with pieces of Scala code:
```
julienrf, line 17:
```suggestion
You can run Scala in your browser with _Scastie_. This is an easy, zero-setup way to experiment with pieces of Scala code:
```

julienrf: We should also update the text in the other languages.

julienrf (changes requested):
Thank you! Could you please also update the references to “ScalaFiddle” in the text?

_tour/basics.md
slee500: Thanks, updated. I guess once merged we can close this issue https://github.com/scala/docs.scala-lang/issues/2417 too

slee500:
I guess once merged we can close this issue https://github.com/scala/docs.scala-lang/issues/2417 too

julienrf (approved):
Thank you!

PR #2450, opened 2022-06-30
Add Scala version picker to Scala Book Hello World Page, remember preference

bishabosha:
changes the Hello World page to feature examples for Scala 2 and Scala 3, switched via code tabs.

We then activate all other matching code tabs on the page when one is selected.

We further use the localStorage API to remember which Scala version a user has selected, for when the page is revisited at a later time (or when other pages adopt the same code tabs)

### DIFF
--- a/_overviews/contribute/add-guides.md
+++ b/_overviews/contribute/add-guides.md
@@ -61,24 +61,6 @@ on [docs.scala-lang.org][home].
 
     And a paragraph, with a [link](https://www.scala-lang.org).
 
-You can contribute code in a markdown document by either
-- in-line by putting backticks around it,
-- surrounding by triple backticks,
-- or indenting it by 4 spaces, e.g.:
-
-~~~
-inline example: `val x = 23`
-
-block example:
-```scala
-println("hello")
-```
-
-indented example:
-
-    case class Foo(x: Int)
-~~~
-
 Tables of contents will be automatically generated in a sidebar for your document, and syntax highlighting
 is provided.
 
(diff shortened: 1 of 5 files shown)


resources/js/functions.js
```diff
@@ -382,6 +382,78 @@ $(document).ready(function() {
...
+}
+
+// Store preference for Scala 2 vs 3
+$(document).ready(function() {
+
+  const Storage = (namespace) => {
+    return ({
+      getPreference(key, defaultValue) {
+        const res = localStorage.getItem(`${namespace}.${key}`);
+        return !res ? defaultValue : res;
```
julienrf, line 406:
```suggestion
        return res === null ? defaultValue : res;
```

Would be clearer IMHO but YMMV.

resources/js/functions.js
```diff
@@ -382,6 +382,78 @@ $(document).ready(function() {
...
+        const old = this.getPreference(key, null);
+        if (old !== value) { // activate effect only if value changed.
+          localStorage.setItem(`${namespace}.${key}`, value);
+          onChange(old);
+        }
+      }
+    });
+  };
+
+  function setupScalaVersionTabs(scalaVersionTabs) {
```
julienrf, line 418: A comment describing the purpose of this function would be welcome.

julienrf (commented):
It works nicely!

Screenshot from 2022-07-01 08-47-11

However, the nested boxes don’t look very good, IMHO. What do you think of having just a `border-left` for tabbed sections, instead of a whole box?

bishabosha:
@julienrf here is the new block quote look (only in tabs)

Screenshot 2022-07-01 at 11 27 51

resources/js/functions.js
bishabosha: done

resources/js/functions.js
bishabosha: I added a comment

bishabosha:
@julienrf a quieter tab design that is applied across the site (so get-started guide as well)

Screenshot 2022-07-01 at 12 18 51

julienrf (approved):
Great, thank you Jamie!

Do you mind also adding some documentation about this feature? E.g., the fact that documentation authors can (and are encouraged to) write code examples for both Scala 2 and Scala 3 syntaxes, by using the special `tabs` Jekyll directive with `class=tabs-scala-version`. Probably the best place to put this information would be in this page? https://docs.scala-lang.org/contribute/add-guides.html

bishabosha:
the other question to answer is should this preference be global across the site, currently the key is specific to the scala-3 book, but the css class is very generic - so I think if this should be the one way to do it across the site I should make the storage key a bit less specific

julienrf:
Ah, this is a good point, yes I believe that should apply to the whole site.

bishabosha:
@julienrf I have added a doc explainer, and a more generic storage key

_overviews/contribute/add-guides.md
```diff
@@ -79,24 +101,93 @@ indented example:
...
+```scala
+@main def hello() = println("Hello, World!")
+```
+{% endtab %}
+
+{% endtabs %}
+~~~
+<!-- {% endraw  %} -->
+
+It is crucial that you use the `tabs-scala-version` class to benefit from some cool user interactions:
```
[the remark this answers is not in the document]
bishabosha: I would like to try - the only issue is that the inner tabs seem to be evaluated before the outer tab section, (and evaluation of inner tabs is done by the framework, rather than directly from the outer tabs) also the data for the inner tabs is added to a huge global hash table so we do need the id to be passed to the inner tabs somehow to avoid collisions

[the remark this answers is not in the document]
julienrf: OK. That sounds complicated then, let’s keep things as they are…

PR #2455, opened 2022-07-05
Tabs for Scala 2/3 in some pages

sjrd:


_overviews/core/string-interpolation.md
```diff
@@ -67,14 +67,36 @@ interpolator, all variable references should be followed by a `printf`-style for
 The `f` interpolator is typesafe.  If you try to pass a format string that only works for integers but pass a double, the compiler will issue an
 error.  For example:
 
-    val height: Double = 1.9d
+{% tabs cacheF class=tabs-scala-version %}
```
bishabosha, line 70: reusing the `cacheF` (not important but maybe uninformative to a reader of source code)

_tour/traits.md
```diff
@@ -78,6 +128,30 @@ animals.append(dog)
...
+```scala
+import scala.collection.mutable.ArrayBuffer
+
+trait Pet:
+  val name: String
+
+class Cat(val name: String) extends Pet
+class Dog(val name: String) extends Pet
+
+val dog = new Dog("Harry")
```
bishabosha, line 143: drop new

_overviews/collections-2.13/maps.md
```diff
@@ -24,68 +24,87 @@ The fundamental operations on maps are similar to those on sets. They are summar
...
-|  `ms get k`  	            |The value associated with key `k` in map `ms` as an option, `None` if not found.|
-|  `ms(k)`  	            |(or, written out, `ms apply k`) The value associated with key `k` in map `ms`, or exception if not found.|
-|  `ms getOrElse (k, d)`    |The value associated with key `k` in map `ms`, or the default value `d` if not found.|
-|  `ms contains k`  	    |Tests whether `ms` contains a mapping for key `k`.|
-|  `ms isDefinedAt k`  	    |Same as `contains`.                             |
-|   **Subcollections:**     |						     |
+| WHAT IT IS                | WHAT IT DOES                                   |
+| ------                    | ------                                         |
+|  **Lookups:**             |                                                |
+|  `ms.get(k)`              |The value associated with key `k` in map `ms` as an option, `None` if not found.|
```
julienrf, line 30: Thank you for fixing that!

_overviews/core/string-interpolation.md
```diff
@@ -105,42 +127,106 @@ In Scala, all processed string literals are simple code transformations.   Anyti
...
-        parseJson(buf)
-      }
+```scala
+new JsonHelper(new StringContext("{ name: ", ", id: ", " }")).json(name, id)
+```
+{% endtab %}
+
+{% tab 'Scala 3' for=extension-desugaring %}
+```scala
+new StringContext("{ name: ", ", id: ", " }").json(name, id)
```
julienrf, line 183: You could drop `new` here as well

_overviews/core/string-interpolation.md
```diff
@@ -105,42 +127,106 @@ In Scala, all processed string literals are simple code transformations.   Anyti
...
+```
+
+The extension method is then used to rewrite it to the following:
+
+```scala
+json(new StringContext("{ name: ", ", id: ", " }"))(name, id)
+```
```
julienrf, line 190: Why do we want to add that?

sjrd: Because it corresponds to the explanation in Scala 2. But maybe it's more confusing than useful?

julienrf: Yes, I would leave this part because users don’t need to know that in this article. They only need to understand that their `json` method will be called when they write `json"something"`.

_overviews/core/string-interpolation.md
```diff
@@ -105,42 +127,106 @@ In Scala, all processed string literals are simple code transformations.   Anyti
...
+{% endtab %}
+
+{% tab 'Scala 3' for=json-fake-implementation %}
+```scala
+extension (sc: StringContext)
+  def json(args: Any*): JSONObject =
+    val strings = sc.parts.iterator
+    val expressions = args.iterator
+    var buf = new StringBuilder(strings.next())
+    while (strings.hasNext) do
```
bishabosha, line 223:
```suggestion
    while strings.hasNext do
```

_overviews/core/string-interpolation.md
```diff
@@ -105,42 +127,106 @@ In Scala, all processed string literals are simple code transformations.   Anyti
     id"string content"
 
 it transforms it into a method call (`id`) on an instance of [StringContext](https://www.scala-lang.org/api/current/scala/StringContext.html).
-This method can also be available on implicit scope.   To define our own string interpolation, we simply need to create an implicit class that adds a new method
-to `StringContext`.  Here's an example:
+This method can also be available on implicit scope.
+To define our own string interpolation, we need to create an implicit class (Scala 2) or an `extension` method (Scala 3) that adds a new method to `StringContext`.
```
bishabosha, line 131: instead of mentioning both Scala 2 and Scala 3 in this sentence, you could add a specialised line of text to either tab

sjrd: It doesn't really "look well" if I do that, IMO.

julienrf (approved):

PR #2458, opened 2022-07-05
Compare Scala 2 vs Scala 3 with tabs in Tour of Scala / Basics

bishabosha:
use scala 2 vs scala 3 tabs for tour/basics

### DIFF
--- a/_tour/basics.md
+++ b/_tour/basics.md
@@ -157,6 +157,9 @@ There are some other differences, but for now, you can think of methods as somet
 
 Methods can have multi-line expressions as well:
 
+{% tabs get-square-string class=tabs-scala-version %}
+
+{% tab 'Scala 2' for=get-square-string %}
 ```scala mdoc
 def getSquareString(input: Double): String = {
   val square = input * input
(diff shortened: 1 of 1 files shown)


_tour/basics.md
```diff
@@ -167,36 +157,78 @@ There are some other differences, but for now, you can think of methods as somet
...
+```scala
+def getSquareString(input: Double): String =
+  val square = input * input
+  square.toString
+  
+println(getSquareString(2.5)) // 6.25
+```
+{% endtab %}
+
+{% endtabs %}
```
julienrf, line 182: We had a discussion about whether it is worth showing two tabs for a difference of a couple of braces. I wonder if the other website maintainers have an opinion about that? Should we keep showing just one code example (with braces, so that it compiles with both Scala 2 and Scala 3), or should we highlight this difference between Scala 2 and Scala 3 by showing those separate tabs?

julienrf: Maybe @SethTisue you have an opinion?

SethTisue: Well, I believe we need _some_ solution here. I don't think it's acceptable for it to be unclear to the reader whether some code is Scala 2 code, or Scala 3 code, or some code that has been carefully made to work on either.

The two basic possible solutions are: 
* have separate Scala 2 and 3 versions of some pages
* have a single version, but use code tabs

It sounds like in this case it's already decided that we don't want to have separate 2 and 3 versions of the entire page. In that case, I think we should do the tabs.

There is _some_ value in the freedom this gives us to, for example, omit braces from the Scala 3 version. But as I see it, the _primary_ value, the main reason we ought to do this, is to be clear with the reader whether the code is Scala 2 or 3 or both. That's important to do even if the Scala 2 and 3 versions of the code are identical!

julienrf: Thank you for your feedback! We decided to not create tabs when the code snippets are identical, to remove some work. I hope this is fine too?

SethTisue: Could it not have tabs, but say something like "Scala 2 or 3"?

SethTisue: That's what I'm suggesting, yeah.

julienrf (approved):

PR #2461, opened 2022-07-05
Add code tabs for _tour/pattern-matching and scala3-book/control-structures

adpi2:


_overviews/scala3-book/control-structures.md
```diff
@@ -28,30 +28,76 @@ These are all demonstrated in the following sections.
...
 The `if`/`else` syntax looks like this:
 
+{% tabs control-structures-3 class=tabs-scala-version %}
+{% tab 'Scala 2' for=control-structures-3 %}
+```scala
+if (x == 1) {
+  println("x is 1, as you can see:")
+  println(x)
+} else
+  println("x was not 1")
```
julienrf, line 73: It might be a matter of taste, but I usually also add the braces to the `else` branch even if it’s a one-liner, if there are braces to the “then” branch.

_overviews/scala3-book/control-structures.md
```diff
@@ -184,13 +328,28 @@ i = 2, j = b, k = 6
...
+for {
+  i <- 1 to 5
+  if i % 2 == 0
+} {
+  println(i)
+}
+```
+{% endtab %}
+{% tab 'Scala 3' for=control-structures-14 %}
+```scala mdoc
```
bishabosha, line 343: I think there are a few other examples of this
```suggestion
```scala
```

_overviews/scala3-book/control-structures.md
```diff
@@ -60,28 +106,57 @@ else if x == 0 then
...
   println("positive")
 ```
+{% endtab %}
+{% endtabs %}
 
-You can optionally include an `end if` statement at the end of each expression, if you prefer:
+### `end if` statement
+
+<blockquote class="help-info">
+<i class="fa fa-info"></i>&nbsp;&nbsp;This is only new in Scala 3.
```
julienrf, line 115: I would say maybe “This is only supported in Scala 3”, or “This is new in Scala 3, and not supported in Scala 2”, but I am not sure about “This is only new in Scala 3”.

_overviews/scala3-book/control-structures.md
```diff
@@ -354,6 +591,25 @@ Pattern matching is a major feature of functional programming languages, and Sca
 In the most simple case you can use a `match` expression like a Java `switch` statement, matching cases based on an integer value.
 Notice that this really is an expression, as it evaluates to a result:
 
+{% tabs control-structures-22 class=tabs-scala-version %}
+{% tab 'Scala 2' for=control-structures-22 %}
+```scala
+import scala.annotation.switch
```
julienrf, line 597: It is unused. (also in the original code example…)

_tour/pattern-matching.md
```diff
@@ -140,6 +246,20 @@ def findPlaceToSit(piece: Furniture): String = piece match {
   case b: Chair => "Sit on the chair"
 }
 ```
+{% endtab %}
+{% tab 'Scala 3' for=pattern-matching-7 %}
+```scala
+sealed abstract class Furniture
+case class Couch() extends Furniture
+case class Chair() extends Furniture
```
julienrf, line 254: I suggest taking the opportunity to make this example more idiomatic by using `case objects` instead of `case classes`.

julienrf (changes requested):

_tour/pattern-matching.md
bishabosha: shouldn't this also be an enum?

julienrf: The goal is to explain `sealed` here, so we can’t use an enum. Maybe we want to use an example where the leaves are parameterized classes?

_tour/pattern-matching.md
```diff
@@ -106,40 +163,103 @@ println(showImportantNotification(importantEmail, importantPeopleInfo)) // print
...
 ```scala mdoc
-sealed abstract class Furniture
-case class Couch() extends Furniture
-case class Chair() extends Furniture
+sealed trait Furniture
+case object Couch extends Furniture
+case object Chair extends Furniture
 
 def findPlaceToSit(piece: Furniture): String = piece match {
   case a: Couch => "Lie on the couch"
```
bishabosha, line 245:
```suggestion
  case a: Couch => "Lie on the couch"
```
```suggestion
  case Couch => "Lie on the couch"
```

_tour/pattern-matching.md
```diff
@@ -106,40 +163,103 @@ println(showImportantNotification(importantEmail, importantPeopleInfo)) // print
...
   case b: Chair => "Sit on the chair"
 }
 ```
+{% endtab %}
+{% tab 'Scala 3' for=pattern-matching-7 %}
+```scala
+sealed trait Furniture
```
bishabosha, line 252: enum?

adpi2: Yes probably we should use an enum here. The problem is that they have not yet been introduced in the Tour of Scala.

adpi2: Do you mean type parameter or value parameter?
Do you think I should do that in this PR?

bishabosha: Here is my suggestion for both:
```scala
sealed abstract class Furniture:
  def descriptor: String

case class Couch(descriptor: String) extends Furniture
case class Chair(descriptor: String) extends Furniture

def findPlaceToSit(piece: Furniture): String = piece match
  case Couch(descriptor) => "Lie on the " + descriptor + " couch"
  case Chair(descriptor) => "Sit on the " + descriptor + " chair"
```

julienrf: Sounds great to me!

bishabosha (approved):

julienrf (approved):
I think more work is needed about `sealed` (e.g. showing more explicitly that it’s impossible to extend a sealed type from a different file, and that the compiler performs exhaustivity checking).

I understand that this is going too far beyond the scope of the original PR, so I am fine with merging this PR as it is and improve it later. Thanks a lot Adrien for your work!

adpi2:
Thanks for your reviews.

It seems I don't have the right to merge so please go ahead.

PR #2462, opened 2022-07-06
Elaborate Regular Expression Patterns

### LINKED ISSUE #2452: To elaborate Regular Expression Patterns
When reading [Regular Expression Patterns](https://docs.scala-lang.org/tour/regular-expression-patterns.html), one may like to know how to use regular expressions to differentiate input. I suggest adding a showcase to elaborate on the combination of regular expression patterns and pattern matching.

byyue:
Close #2452.

image

### DIFF
--- a/_tour/regular-expression-patterns.md
+++ b/_tour/regular-expression-patterns.md
@@ -58,3 +58,35 @@ key: margin value: 0
 key: height value: 108px
 key: width value: 100
 ```
+
+Moreover, you can combine Pattern Matching and Regular Expression Patterns:
+
+```scala mdoc
+def saveContactInfomation(contact: String): Unit = {
+  import scala.util.matching.Regex
+
+  val emailPattern: Regex = """^(\w+)@(\w+(.\w+)+)$""".r
+  val phonePattern: Regex = """^(\d{3}-\d{3}-\d{4})$""".r
+
+  contact match {
+    case emailPattern(localPart, domainName, _) => 
+      println(s"Hi $localPart, we have saved your email address.")
+    case phonePattern(phoneNumber) => 
+      println(s"Hi, we have saved your phone number $phoneNumber.")
+    case _ => 
+      println("Invalid contact information, neither an email address nor phone number.")
+  }
+}
+
+saveContactInfomation("123-456-7890")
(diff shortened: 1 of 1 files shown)


_tour/regular-expression-patterns.md
```diff
@@ -58,3 +58,35 @@ key: margin value: 0
 key: height value: 108px
 key: width value: 100
 ```
+
+Moreover, you can combine Pattern Matching and Regular Expression Patterns:
```
julienrf, line 62:
```suggestion
Moreover, regular expressions can be used as patterns (in `match` expressions) to conveniently extract the matched groups:
```

_tour/regular-expression-patterns.md
```diff
@@ -58,3 +58,35 @@ key: margin value: 0
...
+
+```scala mdoc
+def saveContactInfomation(contact: String): Unit = {
+  import scala.util.matching.Regex
+
+  val emailPattern: Regex = """^(\w+)@(\w+(.\w+)+)$""".r
+  val phonePattern: Regex = """^(\d{3}-\d{3}-\d{4})$""".r
```
julienrf, line 69: What do you think of reusing the example introduced above, `keyValPattern`, instead of introducing a new one?

_tour/regular-expression-patterns.md
```diff
@@ -58,3 +58,35 @@ key: margin value: 0
...
+saveContactInfomation("2 Franklin St, Mars, Milky Way")
+```
+
+The output would be:
+
+```
+Hi, we have saved your phone number 123-456-7890.
+Hi JohnSmith, we have saved your email address.
+Invalid contact information, neither an email address nor phone number.
+```
```
julienrf, line 92: Maybe consider adding a warning about the fact that the number of extracted groups is not checked: users may, by mistake, write something like `case keyValuePattern(_, _, _)`, which would never match because `keyValuePattern` can only capture two groups.

julienrf (commented):
Thank you for your help @byyue! I have a few comments.

julienrf (approved):

Philippus (approved):

SethTisue:
thank you!

PR #2464, opened 2022-07-06
nit: redundant call to `sb.toString` before print

geminicaprograms:
`Command line arguments` chapter ends with the following code snippet:
```
  ...
  sb.toString
  println(sb)
```

and those two lines IMHO should be replaced with one as:
* `sb.toString` performs actual string materialisation from buffer and returns it,
  but nothing is actually using it
* `println(sb)` relies on StringBuilder having properly implemented `toString`
  method (which is obviously the case) and calls it implicitly again and prints
  it to the std out this time

I have proposed to reduce it to `println(sb.toString)` to give clear intent of `toString`
being actually used to produce the content that is outputted but `println(sb)` would
do to, WDYT?


bishabosha (approved):
great, thanks!

PR #2468, opened 2022-07-12
Subsection for Immutability and Variance

KarahanS:
Proposed the addition of a subsection to [Variance](https://docs.scala-lang.org/tour/variances.html) explaining the relationship between immutability and variance very briefly with an example. Referenced the related part of the document for immutable/mutable classes. Compared two Collections, List and Array to show why one is covariant while the other is invariant. 
Based on the discussion [Why are Arrays invariant, but Lists covariant?](https://stackoverflow.com/questions/6684493/why-are-arrays-invariant-but-lists-covariant).

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -169,7 +169,7 @@ abstract class Serializer[-A] {
 }
 
 val animalSerializer: Serializer[Animal] = new Serializer[Animal] {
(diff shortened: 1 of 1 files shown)


bishabosha (approved):
Thanks @KarahanS, I'm sorry it took so long to review. Array is a bit of a special case in Scala so I changed to ListBuffer as it is closer to something the user can write themselves.

PR #2477, opened 2022-07-26
Feature: Add Scala 3 to the compat table

tgodzik:
I think it would be a bit clearer to include it instead of just having it at the bottom of the page.

### DIFF
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -10,14 +10,14 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
 
 ## Version compatibility table
 
-| JDK version | Minimum Scala versions           | Recommended Scala versions                                 |
-|:-----------:|:---------------------------------|:-----------------------------------------------------------|
-| 19          | 2.13.9 (forthcoming), 2.12.16    | 2.13.9 (forthcoming), 2.12.16                              |
-| 18          | 2.13.7, 2.12.15                  | 2.13.8, 2.12.16                                            |
-| 17          | 2.13.6, 2.12.15                  | 2.13.8, 2.12.16                                            |
-| 11          | 2.13.0, 2.12.4, 2.11.12          | 2.13.8, 2.12.16, 2.11.12                                   |
-| 8           | 2.13.0, 2.12.0, 2.11.0, 2.10.2   | 2.13.8, 2.12.16, 2.11.12, 2.10.7                           |
-| 6, 7        | 2.11.0, 2.10.0                   | 2.11.12, 2.10.7                                            |
+| JDK version | Minimum Scala versions                  | Recommended Scala versions                                 |
+|:-----------:|:----------------------------------------|:-----------------------------------------------------------|
+| 19          | 2.13.9 (forthcoming), 2.12.16           | 2.13.9 (forthcoming), 2.12.16                                     |
+| 18          | 3.1.3, 2.13.7, 2.12.15                  | 2.13.8, 2.12.16                                            |
+| 17          | 3.0.0, 2.13.6, 2.12.15                  | 2.13.8, 2.12.16                                            |
+| 11          | 3.0.0, 2.13.0, 2.12.4, 2.11.12          | 2.13.8, 2.12.16, 2.11.12                                   |
+| 8           | 3.0.0, 2.13.0, 2.12.0, 2.11.0, 2.10.2   | 2.13.8, 2.12.16, 2.11.12, 2.10.7                           |
+| 6, 7        | 2.11.0, 2.10.0                          | 2.11.12, 2.10.7                                            |
 
 Even when a version combination isn't listed as supported, most features may still work.  (But Scala 2.12+ definitely doesn't work at all on JDK 6 or 7.)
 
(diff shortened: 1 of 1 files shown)


_overviews/jdk-compatibility/overview.md
```diff
@@ -10,13 +10,13 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
...
-| 11          | 2.13.0, 2.12.4, 2.11.12          | 2.13.8, 2.12.16, 2.11.12                                   |
-| 8           | 2.13.0, 2.12.0, 2.11.0, 2.10.2   | 2.13.8, 2.12.16, 2.11.12, 2.10.7                           |
-| 6, 7        | 2.11.0, 2.10.0                   | 2.11.12, 2.10.7                                            |
+| JDK version | Minimum Scala versions                  | Recommended Scala versions                                 |
+|:-----------:|:----------------------------------------|:-----------------------------------------------------------|
+| 18          | 2.13.7, 2.12.15, 3.0.0                  | 2.13.8, 2.12.16                                            |
+| 17          | 2.13.6, 2.12.15, 3.0.0                  | 2.13.8, 2.12.16                                            |
+| 11          | 2.13.0, 2.12.4, 2.11.12, 3.0.0          | 2.13.8, 2.12.16, 2.11.12                                   |
+| 8           | 2.13.0, 2.12.0, 2.11.0, 2.10.2, 3.0.0   | 2.13.8, 2.12.16, 2.11.12, 2.10.7                           |
+| 6, 7        | 2.11.0, 2.10.0, 3.0.0                   | 2.11.12, 2.10.7                                            |
```
sjrd, line 19: This last line is misleading. No Scala 3.x version supports JDK 6/7.

tgodzik: Ah, right! I will also make sure about when 3.0.0 started supporting JDK 18

tgodzik: Updated the docs based on :
https://github.com/lampepfl/dotty/blame/20d54d454bdc92edfeb44687b647a5491e35dc11/compiler/src/dotty/tools/backend/jvm/BCodeIdiomatic.scala

som-snytt:
Apparently, 3.x is not a recommended Scala version.

I haven't looked at how 3.x handles `-release`, but user questions suggest that "support" means `-release` and `-target` work.

SethTisue:
Maybe we should just get rid of the "Recommended" column and add a sentence like "Using the latest patch version is always recommended."

SethTisue:
needs rebase

tgodzik:
> Maybe we should just get rid of the "Recommended" column and add a sentence like "Using the latest patch version is always recommended."

Good idea, otherwise it will always be outdated.

_overviews/jdk-compatibility/overview.md
```diff
@@ -10,16 +10,18 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
...
-| 18          | 2.13.7, 2.12.15                  | 2.13.8, 2.12.16                                            |
-| 17          | 2.13.6, 2.12.15                  | 2.13.8, 2.12.16                                            |
-| 11          | 2.13.0, 2.12.4, 2.11.12          | 2.13.8, 2.12.16, 2.11.12                                   |
-| 8           | 2.13.0, 2.12.0, 2.11.0, 2.10.2   | 2.13.8, 2.12.16, 2.11.12, 2.10.7                           |
-| 6, 7        | 2.11.0, 2.10.0                   | 2.11.12, 2.10.7                                            |
-
-Even when a version combination isn't listed as supported, most features may still work.  (But Scala 2.12+ definitely doesn't work at all on JDK 6 or 7.)
+| JDK version | Minimum Scala versions                                       | 
+|:-----------:|:-------------------------------------------------------------|
+| 19          | 3.2.0 (forthcoming), 2.13.9 (forthcoming), 2.12.16           |
```
tgodzik, line 15: I also added `3.2.0` here, which will support JDK 19

tgodzik:
It's rebased and adjusted! :tada:

_overviews/jdk-compatibility/overview.md
```diff
@@ -10,16 +10,18 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
...
+| 18          | 3.1.3, 2.13.7, 2.12.15                                       |
+| 17          | 3.0.0, 2.13.6, 2.12.15                                       |
+| 11          | 3.0.0, 2.13.0, 2.12.4, 2.11.12                               |
+| 8           | 3.0.0, 2.13.0, 2.12.0, 2.11.0, 2.10.2                        |
+| 6, 7        | 2.11.0, 2.10.0                                               |
+
+**Using latest patch version is always recomended**
+
+Even when a version combination isn't lis ted as supported, most features may still work.  (But Scala 2.12+ definitely doesn't work at all on JDK 6 or 7.)
```
SethTisue, line 24: s/lis ted/listed/

SethTisue (changes requested):

_overviews/jdk-compatibility/overview.md
tgodzik: :fearful:  Fixed!

SethTisue (approved):

PR #2483, opened 2022-07-31
Fix difficult to read value name in example.

elmira-mazan:
In my font (default for Firefox I think) the old name `l2` (i.e. _<lowercase L>_2) looks a lot like `12` (i.e. twelve). This changes the name to `log2` (and same for `l3` → `log3`) since that's a lot easier to parse. I was very confused thinking I was reading
 
```scala
val d: Double = 12 // ERROR: Found Logarithm required Double
```
and wondering "how did the compiler decide that twelve was a `Logarithm`???"

Great book overall, thanks for putting it together!


SethTisue (approved):

julienrf:
Thank you!

PR #2486, opened 2022-08-05
Update Gemfile.lock

byyue:
Probably related to macOS.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -54,6 +54,8 @@ GEM
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
     mercenary (0.4.0)
+    nokogiri (1.13.6-x86_64-darwin)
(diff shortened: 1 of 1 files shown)


julienrf:
Why is this change necessary?

byyue:
It is imported automatically after installing dependencies on macOS. Otherwise, you may want to add this file into .gitignore.

julienrf:
Does someone know what is the best practice regarding how to manage the Gemfile.lock?

bishabosha:
I see this discussion suggesting the lock file should be committed as a "gem" for CI purposes, and then the one for your local machine you do not commit:
https://stackoverflow.com/a/4151540/2538602

SethTisue (approved):
none of us are really Gemfile jockeys afaik.

but, seeing as how it already has a similar entry for Linux and this seems to just do the same thing for MacOS, I can't see a reason not to merge this.

PR #2488, opened 2022-08-08
Variance rewrite

martijnhoekstra:
All but a full rewrite. I have always found this section of the tour confusing at least, particularly the part where it does the function.

In my experience, most people start with the false intuition that things should automatically be invariant. That's why I start with invariance, show why things can't be covariant by default. Then I show covariance, and then I show contravariance as the opposite.

I completely omit the profunctor thing with co and contravariance at the same time: I don't think it helps understanding of variance.

It's written a bit more tutorial style than most of the tour, because this tend to be a point where many questions arise, and readers can use a gentle push on understanong.

### DIFF
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -56,7 +56,7 @@ Why could this be a problem? We can get the cat from the box, and it's still an 
   mayAnimalBox.content = Dog("fido")
 ```
 
-There now is a Dog in the Animal box. That's all fine, you can put Dogs in Animal boxes, because Dogs are Animals. But our Animal Box is a Cat Box! You can't put a Dog in a Cat box. Doing so would not only be unpleasant for the dog, if we try to get the cat from our catbox, it would turn out to be a dog, breaking type soundness
+There now is a Dog in the Animal box. That's all fine, you can put Dogs in Animal boxes, because Dogs are Animals. But our Animal Box is a Cat Box! You can't put a Dog in a Cat box. If we could, and then try to get the cat from our catbox, it would turn out to be a dog, breaking type soundness.
 
 ```scala
   val myCat: Cat = myCatBox.content //myCat would be fido the dog!
--- a/_tour/variances.md
+++ b/_tour/variances.md
@@ -10,7 +10,7 @@ previous-page: generic-classes
 redirect_from: "/tutorials/tour/variances.html"
 ---
 
-Variance is a property of parameterized types that relates the subtyping relationship of parameterized types to the subtyping releationship of its type parameter. Scala supports variance annotations of type parameters of [generic classes](generic-classes.html), to allow them to be covariant, contravariant, or invariant if no annotations are used. The use of variance in the type system allows us to make intuitive connections between complex types.
+Variance lets you control how type parameters behave with regards to subtyping. Scala supports variance annotations of type parameters of [generic classes](generic-classes.html), to allow them to be covariant, contravariant, or invariant if no annotations are used. The use of variance in the type system allows us to make intuitive connections between complex types.
 
 ```scala mdoc
 class Foo[+A] // A covariant class
@@ -20,7 +20,7 @@ class Baz[A]  // An invariant class
 
 ### Invariance
 
-By default, type paramters in scala are invariant: subtyping relationships between the type paramter aren't reflected in the parameterized type. To explore why this works the way it does, we look at a simple parameterized type, the mutable box.
+By default, type parameters in Scala are invariant: subtyping relationships between the type parameters aren't reflected in the parameterized type. To explore why this works the way it does, we look at a simple parameterized type, the mutable box.
 
 ```scala mdoc
 class Box[A](var content: A)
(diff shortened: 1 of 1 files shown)


_tour/variances.md
```diff
@@ -10,19 +10,23 @@ previous-page: generic-classes
 redirect_from: "/tutorials/tour/variances.html"
 ---
 
-Variance is the correlation of subtyping relationships of complex types and the subtyping relationships of their component types. Scala supports variance annotations of type parameters of [generic classes](generic-classes.html), to allow them to be covariant, contravariant, or invariant if no annotations are used. The use of variance in the type system allows us to make intuitive connections between complex types, whereas the lack of variance can restrict the reuse of a class abstraction.
+Variance is a property of parameterized types that relates the subtyping relationship of parameterized types to the subtyping releationship of its type parameter. Scala supports variance annotations of type parameters of [generic classes](generic-classes.html), to allow them to be covariant, contravariant, or invariant if no annotations are used. The use of variance in the type system allows us to make intuitive connections between complex types.
```
SethTisue, line 13: s/releationship/relationship/

SethTisue: Maybe begin with "Variance lets you control how type parameters behave with regards to subtyping."

_tour/variances.md
```diff
@@ -10,19 +10,23 @@ previous-page: generic-classes
...
 class Foo[+A] // A covariant class
 class Bar[-A] // A contravariant class
 class Baz[A]  // An invariant class
 ```
 
-### Covariance
+### Invariance
 
-A type parameter `T` of a generic class can be made covariant by using the annotation `+T`. For some `class List[+T]`, making  `T` covariant implies that for two types `A` and `B` where `B` is a subtype of `A`, then `List[B]` is a subtype of `List[A]`. This allows us to make very useful and intuitive subtyping relationships using generics.
+By default, type paramters in scala are invariant: subtyping relationships between the type paramter aren't reflected in the parameterized type. To explore why this works the way it does, we look at a simple parameterized type, the mutable box.
```
SethTisue, line 23: s/paramters/parameters/ (in more than one place)

SethTisue: s/scala/Scala/

_tour/variances.md
```diff
@@ -32,110 +36,90 @@ case class Cat(name: String) extends Animal
...
-def printAnimalNames(animals: List[Animal]): Unit =
-  animals.foreach {
-    animal => println(animal.name)
-  }
+  val myAnimal: Animal = Cat("felix")
+```
 
-val cats: List[Cat] = List(Cat("Whiskers"), Cat("Tom"))
-val dogs: List[Dog] = List(Dog("Fido"), Dog("Rex"))
+What about boxes? If I have a `Box[Cat]`, is it a subtype of `Box[Animal]`, like `Cat` is a subtype of `Animal`? At first sight, it looks like that may be plausible, but if we try to do that, the compiler will tell us we have an error:
```
SethTisue, line 45: "is it a subtype" is faulty parallelism — an instance can't be a subtype, only another type can be a subtype. Maybe pedantic, but I'd suggest rethinking that sentence to avoid the problem.

_tour/variances.md
```diff
@@ -32,110 +36,90 @@ case class Cat(name: String) extends Animal
...
-### Contravariance
+Why could this be a problem? We can get the cat from the box, and it's still an Animal, isn't it? Well, yes. But that's not all we can do. We can also replace the cat in the box with a different animal
 
-A type parameter `A` of a generic class can be made contravariant by using the annotation `-A`. This creates a subtyping relationship between the class and its type parameter that is similar, but opposite to what we get with covariance. That is, for some `class Printer[-A]`, making `A` contravariant implies that for two types `A` and `B` where `A` is a subtype of `B`, `Printer[B]` is a subtype of `Printer[A]`.
+```scala
+  mayAnimalBox.content = Dog("fido")
+```
 
-Consider the `Cat`, `Dog`, and `Animal` classes defined above for the following example:
+There now is a Dog in the Animal box. That's all fine, you can put Dogs in Animal boxes, because Dogs are Animals. But our Animal Box is a Cat Box! You can't put a Dog in a Cat box. If we could, and then try to get the cat from our catbox, it would turn out to be a dog, breaking type soundness.
```
SethTisue, line 59: s/catbox/cat box/ — worth fixing because "catbox" is a litter box 💩

_tour/variances.md
```diff
@@ -32,110 +36,90 @@ case class Cat(name: String) extends Animal
...
 
-```scala mdoc
-abstract class Printer[-A] {
-  def print(value: A): Unit
-}
+```scala
+  val myCat: Cat = myCatBox.content //myCat would be fido the dog!
```
SethTisue, line 62: s/fido/Fido/

_tour/variances.md
SethTisue: even "cat box" has the potential of being read that way, of course, but oh well :-)

_tour/variances.md
```diff
@@ -32,110 +36,90 @@ case class Cat(name: String) extends Animal
...
-### Other Examples
+// prints: Fido, Rex
+printAnimalNames(dogs)
+```
 
-Another example that can help one understand variance is `trait Function1[-T, +R]` from the Scala standard library. `Function1` represents a function with one parameter, where the first type parameter `T` represents the parameter type, and the second type parameter `R` represents the return type. A `Function1` is contravariant over its parameter type, and covariant over its return type. For this example we'll use the literal notation `A => B` to represent a `Function1[A, B]`.
+### Contravariance
 
-Assume the similar `Cat`, `Dog`, `Animal` inheritance tree used earlier, plus the following:
+We've seen we can accomplish Covariance by making sure that we can't put something in the Covariant type, but only get something out. What if we had the opposite, something you can put something in, but can't take out? This situation arises if we have something like a serializer, that takes values of type A, and converts them to a serialized format.
```
SethTisue, line 103: Covariance and Covariant shouldn't be capitalized here

_tour/variances.md
```diff
@@ -32,110 +36,90 @@ case class Cat(name: String) extends Animal
...
+We say that `Serializer` is *contravariant* in `A`, and this is indicated by the `-` before the `A`. A more general serializer is a subtype of a more specific serializer.
+
+More formally, that gives us the reverse relationship: given some `class Contra[-T]`, then if `A` is a subtype of `B`, `Cov[B]` is a subtype of `Cov[A]`. 
 
 ### Comparison With Other Languages
 
 Variance is supported in different ways by some languages that are similar to Scala. For example, variance annotations in Scala closely resemble those in C#, where the annotations are added when a class abstraction is defined (declaration-site variance). In Java, however, variance annotations are given by clients when a class abstraction is used (use-site variance).
+
+Scalas tendency towards immutable types makes it that covariant and contravariant types are more common than in other languages, since a mutabable generic type can't be covariant or contravariant.
```
SethTisue, line 125: s/Scalas/Scala's/
s/mutabable/mutable/
s/can't be covariant or contravariant/must be invariant/

SethTisue (changes requested):
big improvement overall, but I have nitpicks

_tour/variances.md
martijnhoekstra: 😹

SethTisue (approved):

SethTisue:
thank you Martijn!

PR #2490, opened 2022-08-15
language tour: Multiple Parameter Lists: clarify relationship to 'currying'

SethTisue:


_tour/multiple-parameter-lists.md
```diff
@@ -96,3 +96,23 @@ println(squares) // List(1, 4, 9, 16, 25, 36, 49, 64, 81, 100)
...
+
+We discourage the use of the word "curry" in reference to Scala's multiple parameter lists, for two reasons:
+
+1) In Scala, multiple parameters and multiple parameter lists are
+specified and implemented directly, as part of the language, rather
+being derived from single-parameter functions.
+
+2) There is danger of confusion with the Scala standard library's
+[`curried`](https://www.scala-lang.org/api/current/scala/Function2.html#curried:T1=%3E(T2=%3ER))
+and [`uncurried`](https://www.scala-lang.org/api/current/scala/Function$.html#uncurried[T1,T2,R](f:T1=%3E(T2=%3ER)):(T1,T2)=%3ER) methods, which don't involve multiple parameter lists at all.
```
julienrf, line 118: I think there is a relation between multiple parameter lists and currying because calling a method defined with multiple parameter lists _looks like_ calling multiple functions. It is also often used to partially apply a method, which is a common pattern when working with curried functions.

Example:

~~~ scala
def add(x: Int)(y: Int): Int = x + y

add(1)(2) // same syntax as if `add` was `val add: Int => Int => Int = x => y => x + y`

List(1, 2, 3).map(add(1)) // nice syntax for partial application (otherwise, you’d write `.map(add(1, _))`)
~~~

So, I still think there is a relation that would be worth highlighting in the documentation.

SethTisue: Okay, I'll see if I can find a way to convey that.

SethTisue: I've added another commit that addresses this.

I was reluctant to go on and discuss partial application further, because then it's more complicated. In an example such as `List(1, 2, 3).map(add(1))` the syntax is the same either way, but that's because `map` provides an expected type, removing the need to do explicit eta expansion. But sometimes with multiple parameter lists you have to explicitly eta expand in order to partially apply, and the same isn't true of curried functions. Consider:

```scala
scala 2.13.8> def add(m: Int)(n: Int) = m + n
def add(m: Int)(n: Int): Int

scala 2.13.8> val add3 = add(3) _
val add3: Int => Int = $Lambda$1094/0x00000008005f1000@41bdaa81

scala 2.13.8> val curried = (m: Int) => (n: Int) => m + n
val curried: Int => (Int => Int) = $Lambda$1113/0x00000008005fa000@3b8b5b40

scala 2.13.8> val add3 = curried(3)
val add3: Int => Int = $Lambda$1123/0x00000008005fae20@c7cf8c4
```

I guess it would be possible to unpack all this, but I feel like in the language tour we need to a strike a balance between concision and completeness, and I think the commit I've added strikes that balance in a way I'm happy with.

julienrf (approved):
Thank you Seth!

PR #2499, opened 2022-08-17
Add scala 3 code demos for Scala for Java Programmers

bishabosha:
Update text a bit, however the whole thing reads still like it is for 2006 Java programmers.

### DIFF
--- a/_overviews/tutorials/scala-for-java-programmers.md
+++ b/_overviews/tutorials/scala-for-java-programmers.md
@@ -274,7 +274,7 @@ second.
 ```scala
 object Timer {
   def oncePerSecond(callback: () => Unit): Unit = {
-    while (true) { callback(); Thread sleep 1000 }
+    while (true) { callback(); Thread.sleep(1000) }
   }
   def timeFlies(): Unit = {
     println("time flies like an arrow...")
@@ -319,7 +319,7 @@ program, passing an anonymous function to `oncePerSecond` instead of `timeFlies`
 ```scala
 object TimerAnonymous {
   def oncePerSecond(callback: () => Unit): Unit = {
-    while (true) { callback(); Thread sleep 1000 }
+    while (true) { callback(); Thread.sleep(1000) }
   }
   def main(args: Array[String]): Unit = {
     oncePerSecond(() =>
@@ -802,9 +802,9 @@ several operations on the expression `(x+x)+(7+y)`: it first computes
 its value in the environment `{ x -> 5, y -> 7 }`, then
 computes its derivative relative to `x` and then `y`.
 
-{% tabs hello-world-demo class=tabs-scala-version %}
+{% tabs calc-main class=tabs-scala-version %}
 
-{% tab 'Scala 2' for=hello-world-demo %}
+{% tab 'Scala 2' for=calc-main %}
 ```scala
 import Tree._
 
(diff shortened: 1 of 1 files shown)


_overviews/tutorials/scala-for-java-programmers.md
```diff
@@ -159,85 +220,133 @@ types.
...
-        oncePerSecond(timeFlies)
-      }
-    }
+{% tabs callback-demo class=tabs-scala-version %}
+
+{% tab 'Scala 2' for=callback-demo %}
+```scala
+object Timer {
+  def oncePerSecond(callback: () => Unit): Unit = {
+    while (true) { callback(); Thread sleep 1000 }
```
julienrf, line 277: I think there is an opportunity to change the code style to this:

```suggestion
    while (true) { callback(); Thread.sleep(1000) }
```

_overviews/tutorials/scala-for-java-programmers.md
```diff
@@ -159,85 +220,133 @@ types.
...
+Here is a revised version of the timer
+program, passing an anonymous function to `oncePerSecond` instead of `timeFlies`:
+
+{% tabs callback-demo-refined class=tabs-scala-version %}
+
+{% tab 'Scala 2' for=callback-demo-refined %}
+```scala
+object TimerAnonymous {
+  def oncePerSecond(callback: () => Unit): Unit = {
+    while (true) { callback(); Thread sleep 1000 }
```
julienrf, line 322:
```suggestion
    while (true) { callback(); Thread.sleep(1000) }
```

_overviews/tutorials/scala-for-java-programmers.md
```diff
@@ -607,11 +969,41 @@ individual fields. The default implementation of `equals` is not
...
+
+{% tab 'Scala 2' for=equals-definition %}
+```scala
+class Date(y: Int, m: Int, d: Int) extends Ord {
+  // previous decls here
+
+  override def equals(that: Any): Boolean =
+    that.isInstanceOf[Date] && {
+      val o = that.asInstanceOf[Date]
+      o.day == day && o.month == month && o.year == year
```
julienrf, line 982: Should we use pattern matching instead?

julienrf: Eg

```suggestion
    that match {
      case date: Date => date.day == day && date.month == month && date.year == year
      case _ => false
    }
```

julienrf (commented):
Awesome, Jamie! It not only adds Scala 3 code examples, but it also greatly improves the content of the document!

_overviews/tutorials/scala-for-java-programmers.md
bishabosha: im not a fan of keeping this section, but ill make the change

bishabosha: I kept the comparison of `instanceof` and `isInstanceOf`, but instead say it is not idiomatic to use instance of in Scala.

_overviews/tutorials/scala-for-java-programmers.md
bishabosha: good catch, done

_overviews/tutorials/scala-for-java-programmers.md
bishabosha: done

julienrf (approved):
Thank you again Jamie!

PR #2500, opened 2022-08-18
Update comment in example code

mattzhou-db:
Help clarify that x is set to 1, and y is set to default value. This comment is more explicit in what x and y are set to. 

The reason why I proposed this change was it's easy to misread, especially as x and y look visually similar on smaller screens.


julienrf (commented):
Thank you, I like your idea, but could you please also update the Scala 3 code snippet? (line 114)

mattzhou-db:
> Scala 3 code snippet?

Done, sorry for the delay on this.

bishabosha (approved):

PR #2501, opened 2022-08-20
Improve code

AlinGabriel:
1. Improved some `case class` outdated code. 
2. Refactored an outdated example to include invalid `var` member in value class.
3. Other minor spelling improvements.


bishabosha (approved):
thanks a lot!

PR #2504, opened 2022-08-25
Change parameter name

AlinGabriel:
Changed name of the collection to match the collection name `coll` given in the paragraph below the code.


julienrf (approved):
Good catch, thank you!

PR #2507, opened 2022-09-01
Fix import that changed with the update to scalatest 3.2.11

tockrock:
Similar to #2377, this PR fixes where the Scala 2 book is still based on scalatest 3.0.8


julienrf (approved):
Thank you!

Issue #2481, opened 2022-07-27
Update code examples to show both Scala 2 and Scala 3 syntax


julienrf:
We would like to have a single website documenting both Scala 2 and Scala 3 since, in most of the cases, their syntax is very similar, and their features are almost all the same. However, there can still be some syntactic differences between idiomatic Scala 2 code and idiomatic Scala 3 code. Therefore, we want to display both styles in our code examples by using tabs:

![Screenshot 2022-07-27 at 14-37-55 Futures and Promises](https://user-images.githubusercontent.com/332812/181248486-43448acf-f5ba-48a9-bdca-8842728cade8.png)

For code examples that are idiomatic in both Scala 2 and Scala 3, we show a single tab labelled “Scala 2 and 3”:

![Screenshot from 2022-07-27 14-39-09](https://user-images.githubusercontent.com/332812/181248699-a9748caf-1951-43e9-82ec-a106c4365bc6.png)

---

Unfortunately, there are lots of code examples in the documentation, and updating all of them is a huge (and repetitive) task for one person. But if a few dozen members of the community were each updating a couple of pages, the job would be done very quickly. So, everyone is welcome to help!

The policy for updating code examples is the following: show the most idiomatic style for all the code examples, for Scala 2 and Scala 3. For Scala 3 code examples, this means using the [new control syntax](https://docs.scala-lang.org/scala3/reference/other-new-features/control-syntax.html), the [indentation-significant syntax](https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html), `using` and `given`, `@main`, `.*` wildcard import, `?` wildcard type argument, tuple indexed access, etc.

You can find the instructions to write tabbed code examples in the [documentation](https://docs.scala-lang.org/contribute/add-guides.html#code-blocks).

Here is the list of pages that need to be updated. If you are interested in contributing, please add a comment to announce that you start working on a page, and I will mark it “in progress” to prevent overlapping contributions:

- [x] Tour of Scala (source files are in the `_tour/` directory)
  - [x] ~Basics~ #2458
  - [x] ~Unified Types~ #2492
  - [x] ~Classes~ #2474
  - [x] ~Default Parameter Values~ #2493
  - [x] ~Named Arguments~ #2521
  - [x] ~Traits~ #2455
  - [x] ~Tuples~ #2476
  - [x] ~Class Composition with Mixins~ #2524
  - [x] ~Higher-order Functions~ #2525
  - [x] ~Nested Methods~ #2528
[...]

PR #2513, opened 2022-09-05
Compare Scala 2 vs Scala 3 with tabs in Tour of Scala / Named Arguments

flomebul:
See https://github.com/scala/docs.scala-lang/issues/2481:
This is a simple page, so a good first try for document update :-) Waiting for feedback.


bishabosha:
I've opened https://github.com/scala/docs.scala-lang/pull/2521 instead, with your commit

PR #2525, opened 2022-09-15
Add code tabs for _tour/higher-order-functions

flomebul:


_tour/higher-order-functions.md
```diff
@@ -20,27 +20,72 @@ In a pure Object Oriented world a good practice is to avoid exposing methods par
 
 One of the most common examples is the higher-order
 function `map` which is available for collections in Scala.
+
+{% tabs map_example_1 class=tabs-scala-version %}
+
+{% tab 'Scala 2' for=map_example_1 %}
```
bishabosha, line 26: surely this tab could be Scala 2 and 3? (scala 2 supports underscores in numeric literals)

_tour/higher-order-functions.md
```diff
@@ -20,27 +20,72 @@ In a pure Object Oriented world a good practice is to avoid exposing methods par
...
+
 `doubleSalary` is a function which takes a single Int, `x`, and returns `x * 2`. In general, the tuple on the left of the arrow `=>` is a parameter list and the value of the expression on the right is what gets returned. On line 3, the function `doubleSalary` gets applied to each element in the
 list of salaries.
 
 To shrink the code, we could make the function anonymous and pass it directly as
 an argument to map:
+
+{% tabs map_example_2 class=tabs-scala-version %}
+
+{% tab 'Scala 2' for=map_example_2 %}
```
bishabosha, line 52: same here, one tab please

_tour/higher-order-functions.md
flomebul: I have try, but it was refused by `mdoc`step (see [8e87cbb](https://github.com/scala/docs.scala-lang/pull/2525/commits/8e87cbb6dcb3d74077b0703948a1e85ddbe4cd75)) if I understand well...

bishabosha: Right I didn't check the other commits, that's unfortunate, perhaps the mdoc configuration is outdated. However I'd still prefer to have 1 tab without the underscores - I don't think they are a fundamental improvement because at least for now our syntax highlighting doesn't support it

flomebul: Ok. If fact, mdoc look like to check with scala 2.12.4, that does not support such constant...
We just have to not use "mdoc". I will perform the change.

bishabosha (approved):
Thanks!

bishabosha:
updated https://github.com/scala/docs.scala-lang/issues/2481

PR #2526, opened 2022-09-16
Add code tabs for _tour/nested-functions

flomebul:


_tour/nested-functions.md
```diff
@@ -24,10 +27,32 @@ In Scala it is possible to nest method definitions. The following object provide
...
+{% tab 'Scala 3' for=Nested_functions_definition %}
+```scala
+ def factorial(x: Int): Int =
+    def fact(x: Int, accumulator: Int): Int =
+      if x <= 1 then accumulator
+      else fact(x - 1, x * accumulator)
+    fact(x, 1)
+
+ println("Factorial of 2: " + factorial(2))
+ println("Factorial of 3: " + factorial(3))
```
bishabosha, line 41: The indentation is inconsistent here - `fact` is indented 3 spaces from `factorial`, but the `if` is only indented 2 spaces. Also both 2 and 3 code examples seem to have a 1 space margin

bishabosha: I realise it was already indented this way, but it would be nice to clean it up now

bishabosha (approved):

bishabosha:
@flomebul not sure what happened here

flomebul:
I wrongly delete some file locally...

PR #2529, opened 2022-09-16
Check mdoc code fences with scala 2.13 syntax

### LINKED ISSUE #2527: mdoc checking is for 2.12 syntax
@SethTisue @julienrf can we update to use the 2.13 version? or do we need to support 2.12 for a reason?

bishabosha:
try upgrade mdoc for scala 2.13

fixes #2527


bishabosha:
@SethTisue passes CI now

SethTisue (approved):

PR #2535, opened 2022-09-17
 Add code tabs for _tour/extractor-objects

flomebul:


_tour/extractor-objects.md
```diff
@@ -31,28 +34,68 @@ customer1ID match {
...
+
+{% tab 'Scala 2 and 3' for=extractor-objects_use-case-3 %}
+```scala mdoc:crash
 val CustomerID(name3) = "-asdfasdfasdf"
 ```
+{% endtab %}
+
+{% endtabs %}
 
 The return type of an `unapply` should be chosen as follows:
```
bishabosha, line 100: A follow up PR could customise this text for the new name based pattern matching in scala 3

bishabosha (approved):
thanks!

PR #2539, opened 2022-09-18
correct valueOrError to valueOrAbort

lmlynik:
In some places the docs mention `valueOrError` which is deprecated in favor of `valueOrAbort`.

The missing `import quotes.reflect.*` made it hard to figure out where the `report` comes from, I think it's worth including it if someone wants to follow along with the code in their IDE.

### DIFF
--- a/_overviews/scala3-macros/tutorial/macros.md
+++ b/_overviews/scala3-macros/tutorial/macros.md
@@ -113,7 +113,7 @@ def powerCode(
   x: Expr[Double],
   n: Expr[Int]
 )(using Quotes): Expr[Double] =
-  val value: Double = pow(x.valueOrError, n.valueOrError)
+  val value: Double = pow(x.valueOrAbort, n.valueOrAbort)
   Expr(value)
 ```
 Here, the `pow` operation is a simple Scala function that computes the value of `xⁿ`.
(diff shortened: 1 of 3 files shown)


_overviews/scala3-macros/tutorial/macros.md
```diff
@@ -131,28 +131,30 @@ Other types can also work if a `ToExpr` is implemented for it, we will [see this
...
 It attempts to extract a value of type `T` from an expression of type `Expr[T]`.
 This can only succeed, if the expression directly contains the code of a value, otherwise, it will throw an exception that stops the macro expansion and reports that the expression did not correspond to a value.
 
-Instead of `valueOrError`, we could also use the `value` operation, which will return an `Option`.
+Instead of `valueOrAbort`, we could also use the `value` operation, which will return an `Option`.
 This way we can report the error with a custom error message.
 
 ```scala
   ...
+  import quotes.reflect.*
```
julienrf, line 143: Where does `quotes` come from?

julienrf (commented):
Thank you for your help! Please see my question below.

_overviews/scala3-macros/tutorial/macros.md
lmlynik: This is a little bit of magic that I don't understand myself, but it's coming from `using Quotes`
How does that work here exactly I don't know. If the example would like this

```scala
  x: Expr[Double],
  n: Expr[Int]
)(using quotes: Quotes): Expr[Double] =
    import quotes.reflect.*
```
it would be more explicit, 
but it also works as 

```scala
  x: Expr[Double],
  n: Expr[Int]
)(using Quotes): Expr[Double] =
    import quotes.reflect.*
```
But definitely `report` is coming from `quotes: Quotes`

julienrf: Is the fully qualified import `scala.quoted.quotes.*` then? As I see it seems to be this definition: https://scala-lang.org/api/3.x/scala/quoted.html#quotes-1e0

bishabosha: I added in https://github.com/scala/docs.scala-lang/pull/2539/commits/0423507de3df93dedfd00fc1320d13102454c2fe some explanation

_overviews/scala3-macros/tutorial/macros.md
```diff
@@ -190,23 +208,24 @@ Varargs in Scala are represented with `Seq`, hence when we write a macro with a
 It is possible to recover each individual argument (of type `Expr[T]`) using the `scala.quoted.Varargs` extractor.
 
 ```scala
-import scala.quoted.Varargs
+import scala.quoted.{Varargs, *}
```
julienrf, line 211: Is this different from that?

```suggestion
import scala.quoted.*
```

bishabosha: they are equivalent, I only put it to make it clear Varargs was imported, but I can go back to just the `*`

julienrf: No, I like clarity. Would it be possible to avoid the wildcard imports in this page? Or is it impractical to work with the API without a wildcard import?

julienrf (approved):

PR #2546, opened 2022-09-21
Fix localized contents

artemkorsakov:
As you can see all localized pages that extend the template `layout: multipage-overview` have empty content.
For example:
https://docs.scala-lang.org/ja/overviews/parallel-collections/concrete-parallel-collections.html
https://docs.scala-lang.org/es/overviews/parallel-collections/concrete-parallel-collections.html
https://docs.scala-lang.org/ru/overviews/collections-2.13/introduction.html
...

It's because `pg.id` and `pg.url` already have localized values. And if we add `page.language`, we'll get the wrong value - `/ru/ru/scala3/book/introduction`

Fixed!


bishabosha (approved):
Fantastic! thank you!

PR #2552, opened 2022-09-22
Add code tabs for _tour/compound-types

flomebul:


_tour/compound-types.md
```diff
@@ -24,28 +26,66 @@ trait Resetable {
...
+```scala mdoc:fail
 def cloneAndReset(obj: Cloneable with Resetable): Cloneable = {
   //...
 }
 ```
-
 Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
 The general form is: `A with B with C ... { refinement }`
+{% endtab %}
+{% tab 'Scala 3' for=compound-types_2 %}
```
bishabosha, line 77:
```suggestion
{% tab 'Scala 3' for=compound-types_3 %}
```

_tour/compound-types.md
```diff
@@ -24,28 +26,66 @@ trait Resetable {
...
-
-```
+```scala mdoc:fail
 def cloneAndReset(obj: Cloneable with Resetable): Cloneable = {
   //...
 }
 ```
-
 Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
 The general form is: `A with B with C ... { refinement }`
```
Sporarum, line 75: I suggest we update like this:
```suggestion
Note that you can have more than two types: `A with B with C with ...`.
This means the same as thing as `(...(A with B) with C`) with ... )`
```

_tour/compound-types.md
```diff
@@ -24,28 +26,66 @@ trait Resetable {
...
+This compound type is written like this in Scala: `Cloneable & Resetable`.
+
+Here's the updated function:
+```scala
+def cloneAndReset(obj: Cloneable & Resetable): Cloneable = {
+  //...
+}
+```
+Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
+The general form is: `A & B & C ... { refinement }`
```
Sporarum, line 87: I suggest we update like this:
```suggestion
Note that you can have more than two types: `A & B & C & ...`. 
And `&` is associative, so parentheses can be added around any part without changing the meaning.
```

Sporarum (changes requested):
Hello, thank you so much for your work on updating the pages !

I think this page is a bit particular in that aspect, as "compound type" specifically refers to `A with B`, the name for `A & B` is an [intersection type](https://docs.scala-lang.org/scala3/reference/new-types/intersection-types.html)

Since the title of page explicitly refers to the outdated term, I suggest we change it, for example to "Combining types"

The name of the file should be updated to reflect the change of name. To ensure the old address still works, please use `redirect-from` ([examples](https://github.com/jekyll/jekyll-redirect-from#usage))

_tour/intersection-types.md
```diff
@@ -0,0 +1,98 @@
...
+{% tab 'Scala 2' for=compound-types_3 %}
+This compound type is written in Scala as `Cloneable with Resetable`.
+
+Here's the updated function:
+```scala mdoc:fail
+def cloneAndReset(obj: Cloneable with Resetable): Cloneable = {
+  //...
+}
+```
+Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
```
Sporarum, line 79: This either needs to be dropped, or an example needs to be given

In all cases, we need a user-facing article on what a refinement is (there is none currently)
Especially non-structural refinement
(This is broader than this PR)

_tour/intersection-types.md
```diff
@@ -0,0 +1,98 @@
...
+{% tab 'Scala 3' for=compound-types_3 %}
+This intersection type is written in Scala as `Cloneable & Resetable`.
+
+Here's the updated function:
+```scala
+def cloneAndReset(obj: Cloneable & Resetable): Cloneable = {
+  //...
+}
+```
+Compound types can consist of several object types and they may have a single refinement which can be used to narrow the signature of existing object members.
```
Sporarum, line 92: Same as for Scala 2

bishabosha:
@flomebul I hope you don't mind me adding some extra commits, but I think changing the content of the article was out of the scope of "add code tabs" so I didn't want to burden that on you

Sporarum (approved):

bishabosha:
@flomebul thanks again!

bishabosha (approved):

bishabosha:
added to #2481

PR #2555, opened 2022-09-22
Update scala-features.md

artemkorsakov:
We can see 'for Apache Spark' twice on the page
https://docs.scala-lang.org/scala3/book/scala-features.html#ai-machine-learning

May be it's a typo?


bishabosha (approved):
I interpret it as the expansion of the acronym, followed by what the library is used for, but I agree that together the second is redundant

bishabosha:
thanks!

PR #2558, opened 2022-09-22
modified num3 and num4 according to latest English version.

benluo:


benluo:
I will correct them all and pull request seperately.

PR #2563, opened 2022-09-24
Add first five pages of Scala3 book in russian

artemkorsakov:
Add first five pages of Scala3 book in russian. If it's OK, I'll continue translate Scala3 book.


artemkorsakov:
Up :)

PR #2565, opened 2022-09-27
Add code tabs for _tour/implicit-conversions

flomebul:
As this update is more complex due to evolution between 2 and 3, thanks to re-read carefully.

### DIFF
--- a/_tour/implicit-conversions.md
+++ b/_tour/implicit-conversions.md
@@ -10,7 +10,18 @@ previous-page: implicit-parameters
 redirect_from: "/tutorials/tour/implicit-conversions.html"
 ---
 
+{% tabs implicit-conversions_1 class=tabs-scala-version %}
+{% tab 'Scala 2' for=implicit-conversions_1 %}
+
 An implicit conversion from type `S` to type `T` is defined by an implicit value which has function type `S => T`, or by an implicit method convertible to a value of that type.
+
+{% endtab %}
+{% tab 'Scala 3' for=implicit-conversions_1 %}
+
+An implicit conversion from type `S` to type `T` is defined by an `implicit def` of type `S => T`, or by a `given` instance of type `conversion[S, T]`.
+
+{% endtab %}
+{% endtabs %}
 
 Implicit conversions are applied in two situations:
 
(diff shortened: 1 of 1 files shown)


_tour/implicit-conversions.md
```diff
@@ -38,17 +69,36 @@ implicit def list2ordered[A](x: List[A])
...
 
+{% tabs implicit-conversions_5 %}
+{% tab 'Scala 2 and 3' for=implicit-conversions_5 %}
 ```scala mdoc
 import scala.language.implicitConversions
 
-implicit def int2Integer(x: Int) =
-  java.lang.Integer.valueOf(x)
+implicit def int2Integer(x: Int): java.lang.Integer =
+    x.asInstanceOf[java.lang.Integer]
```
bishabosha, line 98: not sure why this change to cast was made

_tour/implicit-conversions.md
```diff
@@ -10,8 +10,19 @@ previous-page: implicit-parameters
...
 
+{% tabs implicit-conversions_1 class=tabs-scala-version %}
+{% tab 'Scala 2' for=implicit-conversions_1 %}
+
 An implicit conversion from type `S` to type `T` is defined by an implicit value which has function type `S => T`, or by an implicit method convertible to a value of that type.
 
+{% endtab %}
+{% tab 'Scala 3' for=implicit-conversions_1 %}
+
+An implicit conversion from type `S` to type `T` is defined by an `implicit def` of type `S => T`, or by a `given` instance of type `conversion[S, T]`.
```
bishabosha, line 21:
```suggestion
An implicit conversion from type `S` to type `T` is defined by an `implicit def` of type `S => T`, or by a `given` instance of type `Conversion[S, T]`.
```

_tour/implicit-conversions.md
```diff
@@ -20,13 +31,33 @@ Implicit conversions are applied in two situations:
...
+{% endtab %}
+{% tab 'Scala 3' for=implicit-conversions_2 %}
+
+If an implicit method `List[A] => Ordered[List[A]]` is in scope, as well as a `given`instance of `Conversion[Int, Ordered[Int]]`, the following operation on the two lists of type `List[Int]` is legal:
+
+{% endtab %}
+{% endtabs %}
+
+{% tabs implicit-conversions_3 %}
+{% tab 'Scala 2 and 3' for=implicit-conversions_3 %}
```
bishabosha, line 48: I think it looks better if this code example is duplicated in both of the above tabs, rather than making a separate tab group

bishabosha (changes requested):
I think the majority of this article is pretty confusing and outdated, so trying to update it for scala 3 is adding more confusion

_tour/implicit-conversions.md
bishabosha, line 21: I think its better to go all-in on the Scala 3 way to do things
```suggestion
An implicit conversion from type `S` to type `T` is defined by a `given` instance of type `scala.Conversion[S, T]`.
```

bishabosha:
I'm sorry to close this, but I think this chapter of the tour needs a restructuring to make sense - so I will not accept this PR, feel free to reopen if you want to take an attempt.

julienrf:
Thank you @flomebul for your great help so far! To add to Jamie’s comment, we want to leverage the “Scala 2/3 code tabs” effort to also clean up some pages of the documentation that need to be refreshed, and this page is one of them. We will either come up with more specific instructions for external contributors like you to help, or we will open the PR ourselves.

PR #2573, opened 2022-10-02
removed wrong variable(languages) from result

kameshkotwani:
The result of code section `val(num,str,person) = t` will give out only num,str and person. There is no `languages` val in Person class as well as the tuple, so removed it to prevent confusion.


Philippus (approved):

Philippus:
Thank you @kameshkotwani  !

PR #2574, opened 2022-10-03
Add code tabs for collections-2.13/arrays, strings, equality and views

flomebul:


bishabosha (approved):
thanks a lot, I also added a commit to add more spacing to the repl examples and avoid infix method calls, I also had to add a scala 3 code tab for the varargs spread operator

PR #2579, opened 2022-10-05
Talk about top-level-definitions in the tour of Scala

bishabosha:
Adds code tabs and new explanations for top-level-definitions vs package objects.

Talk about exporting at top level as an alternative to package objects extending traits

### DIFF
--- a/_tour/package-objects.md
+++ b/_tour/package-objects.md
@@ -1,29 +1,50 @@
 ---
 layout: tour
-title: Package Objects
+title: Top Level Definitions in Packages
 partof: scala-tour
 
 num: 36
 previous-page: packages-and-imports
 ---
 
-# Package objects
+Often, it is convenient to have definitions accessible accross an entire package, and not need to invent a
+name for a wrapper `object` to contain them.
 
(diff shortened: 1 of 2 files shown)


_tour/package-objects.md
```diff
@@ -62,11 +104,53 @@ object PrintPlanted {
   }
 }
 ```
+{% endtab %}
+{% tab 'Scala 3' for=pkg-obj-vs-top-lvl_4 %}
+```
+// in file PrintPlanted.scala
+import gardening.fruits.*
+
+@main def PrintPlanted: Unit =
```
julienrf, line 113: This is a bit unusual. Main methods are usually written in lower case:

```suggestion
@main def printPlanted(): Unit =
```

On line 90, maybe you can say “the following program imports…” (without giving the name of the program)?

julienrf (approved):
Thank you, this is a great change! I have left one minor comment.

_tour/package-objects.md
bishabosha: that's a good suggestion! thanks

PR #2580, opened 2022-10-05
add code tabs in num8

benluo:


_overviews/scala3-book/taste-control-structures.md
```diff
@@ -19,12 +19,24 @@ Scala has the control structures you find in other programming languages, and al
 
 These structures are demonstrated in the following examples.
 
-
-
 ## `if`/`else`
 
-Scala’s `if`/`else` control structure looks similar to other languages:
+Scala’s `if`/`else` control structure looks similar to other languages. In Scala 2, it was constructed differently, with parentheses required and curly brackets instead of the keyword `then`:
```
bishabosha, line 24:
```suggestion
Scala’s `if`/`else` control structure looks similar to other languages.
```

bishabosha (dismissed):
Firstly thank you for taking the time to add the code tabs, however I have a point with the style of text:

I realise that the present book currently compares to Scala 2, but going forward we would like to present the page more neutrally, so it is not biased towards scala 3. I would accept if the text was presented more neutrally

bishabosha (approved):
Thanks!

PR #2588, opened 2022-10-06
add code tabs in num15.

benluo:


_overviews/scala3-book/taste-toplevel-definitions.md
```diff
@@ -12,6 +12,8 @@ next-page: taste-summary
 In Scala 3, all kinds of definitions can be written at the “top level” of your source code files.
 For instance, you can create a file named _MyCoolApp.scala_ and put these contents into it:
 
+{% tabs toplevel_1 class=tabs-scala-version %}
```
bishabosha, line 15: `class=tabs-scala-version` should not be used for single tabs. Soon I am adding assertions to make CI fail for this
```suggestion
{% tabs toplevel_1 %}
```

bishabosha (approved):
Thanks! - I think going forward we should consider how to cover package objects in the book

PR #2593, opened 2022-10-07
Add assertions for formatting of code tabs

bishabosha:
update the plugin to validate:
- no duplicate tab labels
- when using `class=tabs-scala-version`, validate that only `Scala 2` and `Scala 3` are used as tab labels
- validate that tab with `for=my-tab-group` is in a tab group called `my-tab-group`

I've also made it optional to put `for=my-tab-group` on each tab, now we track the enclosing tab-group.

Adding these assertions exposed some formatting problems with already merged updates to docs - so it would be good to add this asap


julienrf (approved):
Thank you Jamie! I must say that I don’t feel super comfortable with the Ruby code, but that looks fine to me, let’s merge it!

PR #2594, opened 2022-10-07
add code tabs in num24.

benluo:


_overviews/scala3-book/methods-most.md
```diff
@@ -311,20 +541,54 @@ cat.speak()
...
+   */
+  def containsWhitespace(s: String): Boolean =
+    s.matches(".*\\s.*")
+
+}
+```
+
+{% endtab %}
+
+{% tab 'Scala 3' for=method_1 %}
```
bishabosha, line 590:
```suggestion
{% tab 'Scala 3' for=method_22 %}
```

_overviews/scala3-book/methods-most.md
```diff
@@ -359,30 +624,46 @@ Their main purpose is to let you add new functionality to closed classes.
...
   def area: Double = math.Pi * c.radius * c.radius
 ```
 
+{% endtab %}
+{% endtabs %}
+
 Now when you have a `Circle` instance named `aCircle`, you can call those methods like this:
 
+{% tabs method_25 %}
+{% tab 'Scala 2 and 3' for=method_25 %}
```
bishabosha, line 655: this does not make any sense to have a scala 2 and 3 tab when these methods were never defined in scala 2

_overviews/scala3-book/methods-most.md
```diff
@@ -359,30 +624,46 @@ Their main purpose is to let you add new functionality to closed classes.
...
 case class Circle(x: Double, y: Double, radius: Double)
 ```
 
+{% endtab %}
+{% endtabs %}
+
 When you want to add methods to this class, you can define them as extension methods, like this:
 
+{% tabs method_24 %}
+{% tab 'Scala 3 only' for=method_24 %}
```
bishabosha, line 640: there should also be a tab with scala 2 equivalent

_overviews/scala3-book/methods-most.md
benluo: changed to Scala 3 Only.
I cannot write scala 2 code.

_overviews/scala3-book/methods-most.md
benluo: changed to Scala 3 Only.
I cannot write scala 2 code.

_overviews/scala3-book/methods-most.md
```diff
@@ -123,39 +184,79 @@ The body of a method can use all the different features of the language:
 
 As an example of a real-world multiline method, this `getStackTraceAsString` method converts its `Throwable` input parameter into a well-formatted `String`:
 
+{% tabs method_8 class=tabs-scala-version %}
+{% tab 'Scala 2' for=method_8 %}
+
+```scala
+def getStackTraceAsString(t: Throwable): String = {
+  val sw = StringWriter()
```
bishabosha, line 192:
```suggestion
  val sw = new StringWriter()
```

_overviews/scala3-book/methods-most.md
```diff
@@ -123,39 +184,79 @@ The body of a method can use all the different features of the language:
...
+  t.printStackTrace(PrintWriter(sw))
```
bishabosha, line 193:
```suggestion
  t.printStackTrace(new PrintWriter(sw))
```

_overviews/scala3-book/methods-most.md
```diff
@@ -241,37 +382,76 @@ def isTruthy(a: Any) =
...
 ## Using `match` as a method body
 
 A `match` expression can also be used as the entire method body, and often is.
 Here’s another version of `isTruthy`, written with a `match` expression :
 
+{% tabs method_18 class=tabs-scala-version %}
+{% tab 'Scala 2' for=method_18 %}
+
+```scala
+def isTruthy(a: Matchable) = a match {
```
bishabosha, line 412: There is no `Matchable` in Scala 2 - please test any examples you change
```suggestion
def isTruthy(a: Any) = a match {
```

_overviews/scala3-book/methods-most.md
```diff
@@ -241,37 +382,76 @@ def isTruthy(a: Any) =
...
 ```scala
 def isTruthy(a: Matchable) = a match
   case 0 | "" | false => false
   case _ => true
 ```
 
+{% endtab %}
+{% endtabs %}
+
 This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type to accept any value that supports pattern matching.
```
bishabosha, line 431: however now this needs to be moved to the scala 3 tab because it mentions `Matchable`

_overviews/scala3-book/methods-most.md
```diff
@@ -241,37 +382,76 @@ def isTruthy(a: Any) =
...
 
 For more details on the `Matchable` trait, see the [Reference documentation][reference_matchable].
```
bishabosha, line 433: this sentence also needs to go to the scala 3 tab

_overviews/scala3-book/methods-most.md
benluo: changed.

_overviews/scala3-book/methods-most.md
benluo: changed.

_overviews/scala3-book/methods-most.md
```diff
@@ -241,37 +382,76 @@ def isTruthy(a: Any) =
...
 def isTruthy(a: Matchable) = a match
   case 0 | "" | false => false
   case _ => true
 ```
 
-This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type to accept any value that supports pattern matching.
+> This method works just like the previous method that used an `if`/`else` expression. We use `Matchable` instead of `Any` as the parameter's type to accept any value that supports pattern matching.
 
-For more details on the `Matchable` trait, see the [Reference documentation][reference_matchable].
+> For more details on the `Matchable` trait, see the [Reference documentation][reference_matchable].
```
bishabosha, line 430: I'm not sure why yet but it seems the code tabs plugin does not resolve these links - probably because it sees each tab as an individual markdown document, so I will make the link explicit
```suggestion
> For more details on the `Matchable` trait, see the [Reference documentation]({{ site.scala3ref }}/other-new-features/matchable.html).
```

bishabosha (approved):
Thank you for your work

Issue #1627, opened 2020-01-22
iterable.stringPrefix still exist???


ShapelessCat:
The `stringPrefix` method is mentioned in this page: https://docs.scala-lang.org/overviews/collections-2.13/trait-iterable.html
However, when I open a specific API page like `Iterable`, I **can't** find this method. Also in the REPL, I **can't** invoke this method.

A related issue, when I search "stringPrefix" I can see it in the searching result page of API Doc, but when I open a specific trait/class API page, I can't find it.

SethTisue:
I think a pull request simply removing the reference to `stringPrefix` from that page would be accepted. the method exists, as does its sibling `className`, but they are declared `protected[this]`, which is why you can't call them from the REPL.

afaics, these methods are only of interest to collections implementers, not collections users.

> However, when I open a specific API page like Iterable, I can't find this method

that seems like a bug in the Scaladoc tool — if the method isn't shown on the page, because it is inaccessible except to implementers, it shouldn't be included in search results, either. a bug report on that over at https://github.com/scala/bug would be welcome (if one doesn't already exist, that is!)

som-snytt:
Here is the previous link: https://scala-lang.org/files/archive/api/2.13.0/scala/collection/Iterable.html#stringPrefix:String

API consumed by implementers ought to be in the doc.

I know there is an issue to restore an "access" option to scaladoc, so you can pick whether to show only public or also private elements.

```
scala> new Iterable[Int] { def iterator = Iterator.empty[Int]}
res1: Iterable[Int] = Iterable()

scala> new Iterable[Int] { def iterator = Iterator.empty[Int] ; override def stringPrefix = "Empty" }
                                                                             ^
       warning: overriding method stringPrefix in trait Iterable is deprecated (since 2.13.0): Override className instead
res2: Iterable[Int] = Empty()
```

som-snytt:
I guess they [fiddled] with it.

SethTisue:
> API consumed by implementers ought to be in the doc

but the place for it is https://docs.scala-lang.org/overviews/core/custom-collections.html IMO

som-snytt:
You're right, the overview should be tuned. I'll fix my comment to say scaladoc [sic]. In Java, the API is the javadoc and the wordy stuff is a tutorial.

Jasper-M:
You have to hover over "Filter all members". Then click the triangular arrow that appears. Then you can make protected members visible.

som-snytt:
[...]

PR #2600, opened 2022-10-10
stringPrefix is replaced by className which is private.

flomebul:
I propose to remove from this page. `className` is describe in customisation.


flomebul:
see also #1627

PR #2603, opened 2022-10-11
Update macros.md

boggye:
I wanted to make it clear that in the `inspect` example presented, the println call is executed at compile time.

### DIFF
--- a/_overviews/scala3-macros/tutorial/macros.md
+++ b/_overviews/scala3-macros/tutorial/macros.md
@@ -20,7 +20,7 @@ A Scala expression with type `T` is represented by an instance of the type `scal
 We will dig into the details of the type `Expr[T]`, as well as the different ways of analyzing and constructing instances, when talking about [Quoted Code][quotes] and [Reflection][tasty].
 For now, it suffices to know that macros are metaprograms that manipulate expressions of type `Expr[T]`.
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-macros/tutorial/macros.md
```diff
@@ -20,7 +20,7 @@ A Scala expression with type `T` is represented by an instance of the type `scal
 We will dig into the details of the type `Expr[T]`, as well as the different ways of analyzing and constructing instances, when talking about [Quoted Code][quotes] and [Reflection][tasty].
 For now, it suffices to know that macros are metaprograms that manipulate expressions of type `Expr[T]`.
 
-The following macro implementation prints the expression of the provided argument:
+The following macro implementation prints the expression of the provided argument at compile time:
```
julienrf, line 23:
```suggestion
The following macro implementation prints the expression of the provided argument at compile-time:
```

julienrf (changes requested):
Thank you for clarifying that! I left a minor suggestion.

Other than that, I think it might be even clearer to say where this is going to be printed (in the standard output of the compiler process, I guess?)

julienrf:
It would be great if you could update the PR with the change!

boggye:
> It would be great if you could update the PR with the change!

Done.

julienrf (approved):
Thank you!

PR #2608, opened 2022-10-18
Empty table of contents on the glossary page

### LINKED ISSUE #2607: Empty table of contents on the glossary page
The sidebar with the table of contents (TOC) on the [glossary page](https://docs.scala-lang.org/glossary/index.html) is currently empty. This is due to a change made by commit b88401fc (PR #1861). There, `h4` elements were excluded from the TOC. I think the easiest way to fix this would be to render the glossary terms using the `h3` element instead of `h4`.

lorenzpahl:
Use `h3` headings instead of `h4` on the glossary page. With this change, the glossary terms will appear in the sidebar again.

Resolves #2607


bishabosha (approved):
Really great! thank you

Issue #1286, opened 2019-02-11
improve Docker instructions


SethTisue:
right now the README just says "run `docker-compose up`", but I found that on my MacOS 10.14 laptop, I needed to do the following:

* `brew cask install VirtualBox`
* `brew install docker-machine docker-compose`
* `docker-machine start`
* `eval $(docker-machine env)`
* `docker compose up`

this is probably routine stuff for those familiar with Docker, but some of us are nearly as ignorant of Docker as we are of Ruby and Bundler :-)

the above commands seemed to work, and what I saw in my terminal window ended with:

```
jekyll_1  | Configuration file: /site/_config.yml
jekyll_1  |             Source: /site
jekyll_1  |        Destination: /site/_site
jekyll_1  |  Incremental build: disabled. Enable with --incremental
jekyll_1  |       Generating... 
jekyll_1  |                     done in 29.876 seconds.
jekyll_1  |  Auto-regeneration: enabled for '/site'
jekyll_1  |     Server address: http://0.0.0.0:4000
```

the readme says the site will be served at http://localhost:8080, but it isn't. I'm not sure why, perhaps it's because VirtualBox is installed? how do I view the site? (visiting http://0.0.0.0:4000 doesn't work either)

/cc @kotobotov

kotobotov:
http://0.0.0.0:4000 - that's inside container
witch is bind to out world with 8080 port, 
_VirtualBox_ may have own restriction, 
maybe better just link to https://docs.docker.com/install/ , and note about if you want to use site with docker you need install docker first

SethTisue:
hmm, perhaps I should be using Docker Desktop for Mac https://docs.docker.com/docker-for-mac/docker-toolbox/ in order to leave VirtualBox out of it

PR #2622, opened 2022-11-03
Fix the Docker setup

julienrf:
I found that our previous Docker setup was not good enough and was failing on some machines.

I propose the following changes, based on what has been done on the scala.epfl.ch website. The main “fix” is that we indicate a specific version of Ruby, rouge, and bundler. We also propose a different way to invoke docker-compose up that writes the generated files as a regular user (instead of root).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,13 @@
-FROM ruby:2.6
+FROM ruby:2.6.9
 
-RUN gem install bundler jekyll
+RUN gem install rouge -v3.30.0
+RUN gem install bundler:2.3.10 jekyll
 
 WORKDIR /srv/jekyll
 
 COPY Gemfile .
 COPY Gemfile.lock .
 
+RUN echo -n "bundle version: " && bundle --version
+RUN chmod u+s /bin/chown
 RUN bundle install
-
--- a/README.md
+++ b/README.md
@@ -19,9 +19,13 @@ For more details, read on.
 
 To build and view site with Docker:
 
-    docker-compose up
+    env UID="$(id -u)" GID="$(id -g)" docker-compose up
 
 It will incrementally build and serve site at `http://localhost:4000`.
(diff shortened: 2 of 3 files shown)


SethTisue (approved):
do you see "Docker guy" on my LinkedIn profile?!?!?!?! :-P

before this PR, on my Mac I got:

```
#0 103.4 ERROR:  Error installing jekyll:
#0 103.4 	The last version of rouge (>= 3.0, < 5.0) to support your Ruby & RubyGems was 3.30.0. Try installing it with `gem install rouge -v 3.30.0` and then running the current command again
#0 103.4 	rouge requires Ruby version >= 2.7. The current ruby version is 2.6.10.210.
```

and after this PR it worked

so I'd say that's an "Approve", even though I have no idea what any of this is about

as I also said at https://github.com/scala/docs.scala-lang/issues/1286,

> how do I view the site? (visiting http://0.0.0.0:4000/ doesn't work

I guess because the site is running inside the container and my web browser is outside of the container? So this is "working" in the sense that site seems to build, but without understanding how to actually view the site, all the Docker stuff still isn't actually helping me.

julienrf:
I guess you don't see the website because of VirtualBox.

SethTisue:
I did switch from using VirtualBox to using the Docker desktop app.

It appears from https://www.mend.io/free-developer-tools/blog/docker-expose-port/ that we should be able to add something to the Dockerfile that exposes port 4000. I might mess with that if I get a chance. (I joke about not wanting to mess with Docker, but it’s also been frustrating messing with rbenv and bundler and such over the years, so I actually am interested in improving the Docker option.)

fsalvi:
Port 4000 is already exposed, it works fine under linux. I guess it's a bit different on mac...

fsalvi (approved):
Works fine under linux

Issue #2542, opened 2022-09-20
Remove Scala Actors Migration Guide/Deprecate Page


Khalian:
Akka has recently changed its licensing terms in https://github.com/akka/akka/pull/31561 and as a result https://docs.scala-lang.org/overviews/core/actors-migration-guide.html should not be encouraged since licensing is now an issue and the resulting code wont be Open Source by default. 

Two possible suggestions

1. Put a deprecation warning with the Akka License change documentation https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka as reason

2. Removing that document entirely, since its version 2.11 and above have no dependencies on Akka and scala actors have been a non factor too. 

Thoughts? I can submit a PR either way.

julienrf:
I prefer keeping the page in case it is still useful to someone. It does not hurt, it’s already at the very bottom of the guides page.

I am happy with the solution of adding a note regarding the current license of Akka (even though that does not really apply to the version of Akka used in the migration guide).

Khalian:
@julienrf Understood. I can put a note in there. Thanks.

SethTisue:
I'd have been fine with just removing it, but if we're keeping it, then yes, adding notes or other adjustments that would make the current situation clear to current readers SGTM

strelec:
This docs page looks so outdated it serves nobody, only to confuse.

A simple removal is the best option.

julienrf:
I am fine with removing, in that case.

Khalian:
Hey so I made the change in https://github.com/scala/docs.scala-lang/pull/2604 and deleted the actor migration guide. I presume at some point it would be picked up and reflected in the main scala docs website and we can close this?

julienrf:
I agree to close this issue. Your changes have been published, as you can see here: https://docs.scala-lang.org/overviews/core/actors.html, and here: https://docs.scala-lang.org/overviews/index.html

However, for some reason the deleted page https://docs.scala-lang.org/overviews/core/actors-migration-guide.html is still here. It seems that the script that pulls the new website content does not remove deleted pages. I will open a separate issue to track that.

julienrf:
Thank you for you help @Khalian!

strelec:
Thanks!

Moreover, is there a reason that we are keeping the main Actors page?

https://docs.scala-lang.org/overviews/core/actors.html

I think there is a value of removing stuff that could potentially confuse new users. There are possibly more pages like this one that shall be removed.

[...]

PR #2627, opened 2022-11-06
Delete actors.md

Khalian:
Based on follow up conversations in https://github.com/scala/docs.scala-lang/issues/2542

Specifically this comment

```
Thanks!

Moreover, is there a reason that we are keeping the main Actors page?

https://docs.scala-lang.org/overviews/core/actors.html

I think there is a value of removing stuff that could potentially confuse new users. There are possibly more pages like this one that shall be removed.
```


julienrf (approved):
Thank you @Khalian!

PR #2630, opened 2022-11-08
update tour/implicit-conversions for Scala 3

bishabosha:
adapt Tour of Scala implicit conversions page for scala 3

### DIFF
--- a/_tour/implicit-conversions.md
+++ b/_tour/implicit-conversions.md
@@ -10,46 +10,33 @@ previous-page: implicit-parameters
 redirect_from: "/tutorials/tour/implicit-conversions.html"
 ---
 
-An implicit conversion from type `S` to type `T` is defined by an implicit value which has function type `S => T`, or by an implicit method convertible to a value of that type.
+{% tabs implicit-conversion-defn class=tabs-scala-version %}
+{% tab 'Scala 2' %}
+In Scala 2, an implicit conversion from type `S` to type `T` is defined by an [implicit value]({% link _tour/implicit-parameters.md %}) which has function type `S => T`, or by an implicit method convertible to a value of that type.
+{% endtab %}
+{% tab 'Scala 3' %}
+In Scala 3, an implicit conversion from type `S` to type `T` is defined by a [given instance]({% link _tour/implicit-parameters.md %}) which has type `scala.Conversion[S, T]`, or by an implicit method which can be eta-expanded to the function type `S => T`.
+{% endtab %}
+{% endtabs %}
 
 Implicit conversions are applied in two situations:
 
-* If an expression `e` is of type `S`, and `S` does not conform to the expression's expected type `T`.
-* In a selection `e.m` with `e` of type `S`, if the selector `m` does not denote a member of `S`.
+1. If an expression `e` is of type `S`, and `S` does not conform to the expression's expected type `T`.
+2. In a selection `e.m` with `e` of type `S`, if the selector `m` does not denote a member of `S`.
 
 In the first case, a conversion `c` is searched for which is applicable to `e` and whose result type conforms to `T`.
+
+An example is to pass a `scala.Int`, e.g. `x`, to a method that expects `scala.Long`. In this case, the implicit conversion `Int.int2long(x)` is inserted.
+
+
 In the second case, a conversion `c` is searched for which is applicable to `e` and whose result contains a member named `m`.
 
-If an implicit method `List[A] => Ordered[List[A]]` is in scope, as well as an implicit method `Int => Ordered[Int]`, the following operation on the two lists of type `List[Int]` is legal:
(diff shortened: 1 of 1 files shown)


_tour/implicit-conversions.md
```diff
@@ -10,46 +10,33 @@ previous-page: implicit-parameters
 redirect_from: "/tutorials/tour/implicit-conversions.html"
 ---
 
-An implicit conversion from type `S` to type `T` is defined by an implicit value which has function type `S => T`, or by an implicit method convertible to a value of that type.
+{% tabs implicit-conversion-defn class=tabs-scala-version %}
+{% tab 'Scala 2' %}
+In Scala 2, an implicit conversion from type `S` to type `T` is defined by an [implicit value]({% link _tour/implicit-parameters.md %}) which has function type `S => T`, or by an implicit method convertible to a value of that type.
+{% endtab %}
+{% tab 'Scala 3' %}
+In Scala 3, an implicit conversion from type `S` to type `T` is defined by a [given instance]({% link _tour/implicit-parameters.md %}) which has type `scala.Conversion[S, T]`, or by an implicit method which can be eta-expanded to the function type `S => T`.
```
julienrf, line 18: I think in Scala 3 the support for implicit methods (and implicit values of type functions) is there only for compatibility with Scala 2, no? I would rather say that in Scala 3 an implicit conversion is defined by a `given` instance of type `Conversion`. And then, say that for compatibility with Scala 2, they can also be defined as `implicit` methods (and tell the readers that they will find more information on the Scala 2 tab).

_tour/implicit-conversions.md
```diff
@@ -10,46 +10,33 @@ previous-page: implicit-parameters
...
+An example is to pass a `scala.Int`, e.g. `x`, to a method that expects `scala.Long`. In this case, the implicit conversion `Int.int2long(x)` is inserted.
 
-The implicitly imported object `scala.Predef` declares several aliases to frequently used types (e.g. `scala.collection.immutable.Map` is aliased to `Map`) and methods (e.g. `assert`) but also several implicit conversions.
 
-For example, when calling a Java method that expects a `java.lang.Integer`, you are free to pass it a `scala.Int` instead. That's because Predef includes the following implicit conversions:
+In the second case, a conversion `c` is searched for which is applicable to `e` and whose result contains a member named `m`.
 
-```scala mdoc
-import scala.language.implicitConversions
+An example is to compare two strings `"foo" < "bar"`. In this case, `String` has no member `<`, so the implicit conversion `Predef.augmentString("foo") < "bar"` is inserted.
```
julienrf, line 34: Maybe remind that `Predef.*` is always imported by Scala programs (by default).

_tour/implicit-conversions.md
```diff
@@ -58,3 +45,17 @@ To turn off the warnings take either of these actions:
 * Invoke the compiler with `-language:implicitConversions`
 
 No warning is emitted when the conversion is applied by the compiler.
+{% endtab %}
+{% tab 'Scala 3' %}
+Because implicit conversions can have pitfalls if used indiscriminately the compiler warns in two situations:
+- when compiling the implicit conversion definition (for Scala 2 style conversions).
```
julienrf, line 51: If the Scala 3 compiler does not warn on `given Conversion` instances, then I would make that more explicit:

```suggestion
- when compiling a Scala 2 style implicit conversion definition.
```

julienrf (changes requested):
Thank you Jamie. I would expect also an explanation of where does the compiler search conversions. It can be a link to the implicit-parameters page maybe?

_tour/implicit-conversions.md
bishabosha: that's a good suggestion, I will follow up with that

_tour/implicit-conversions.md
```diff
@@ -15,7 +15,7 @@ redirect_from: "/tutorials/tour/implicit-conversions.html"
 In Scala 2, an implicit conversion from type `S` to type `T` is defined by an [implicit value]({% link _tour/implicit-parameters.md %}) which has function type `S => T`, or by an implicit method convertible to a value of that type.
 {% endtab %}
 {% tab 'Scala 3' %}
-In Scala 3, an implicit conversion from type `S` to type `T` is defined by a [given instance]({% link _tour/implicit-parameters.md %}) which has type `scala.Conversion[S, T]`, or by an implicit method which can be eta-expanded to the function type `S => T`.
+In Scala 3, an implicit conversion from type `S` to type `T` is defined by a [given instance]({% link _tour/implicit-parameters.md %}) which has type `scala.Conversion[S, T]`. For compatibility with Scala 2, they can also be defined by an implicit method (read more in the Scala2 tab).
```
julienrf:
```suggestion
In Scala 3, an implicit conversion from type `S` to type `T` is defined by a [given instance]({% link _tour/implicit-parameters.md %}) which has type `scala.Conversion[S, T]`. For compatibility with Scala 2, they can also be defined by an implicit method (read more in the Scala 2 tab).
```

julienrf: You already changed it

PR #2633, opened 2022-11-15
Update default jekyll tab

bishabosha:
- sets the initial tab to be the default (unless specified, or class=tabs-scala-version)
- add notice to all overviews that code tabs still work when javascript is disabled
- validate that `class=tabs-scala-version` implies exactly `'Scala 2'`, `'Scala 3'` tabs in that order

fixes https://github.com/scalacenter/docs.scala-lang/issues/12

### DIFF
--- a/_layouts/multipage-overview.html
+++ b/_layouts/multipage-overview.html
@@ -17,6 +17,12 @@ includeCollectionTOC: true
 						{% endif %}
 						{% include version-specific-notice.html language=versionSpecificLang %}
 					{% endif %}
+					<noscript>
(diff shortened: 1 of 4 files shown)


julienrf (approved):
Should we add a special rule to select the Scala 3 tab by default in Scala 2/3 tabs?

bishabosha:
> Should we add a special rule to select the Scala 3 tab by default in Scala 2/3 tabs?

this makes sense - we already special case a lot

julienrf (approved):

PR #2635, opened 2022-11-16
Minor change improving readability in pattern matching

vreyespue:


bishabosha (changes requested):
Hello, thank you for opening, I am not willing to accept this however unless "concrete" is still mentioned, as it is the standard vocabulary

vreyespue:
ok, now I see that the word "concrete" is used many times across the documentation 🤔  it just sounded weird to me in english (actually for instance in spanish or german would be fine) 👍 anyhow, I proceed to close this small PR 😄 thanks for considering

PR #2637, opened 2022-11-17
Japanese homepage uses regular layout

bishabosha:
I have updated `_ja/index.md` to use the same layout as the other versions, however I would still like to notify the community because even though I tried to reuse as many links and descriptions as possible, many are now in English.

@hiramekun @ser1zw @rysh I hope you don't mind me mentioning you here, but I see that you have contributed before to the Japanese translations


julienrf (commented):
Thank you for simplifying this. I am fine with merging this PR if the Japanese maintainers approve it.

eed3si9n:
FYI there's also @scala/docs-ja team for review purpose. (I no longer have auth to add new members there, but maybe it would be good to add them in there as well)

eed3si9n (approved):
I am fine with merging this as is, and someone else could follow up with translating links where appropriate?

PR #2634, opened 2022-11-16
Make outdated banner noisier, cleanup layouts

bishabosha:
The outdated notice now is presented as sticky and in the main content window (floating above other sticky content such as the "need help" banner in getting started guide).

I also removed unused layouts, fixed misused layouts, and renamed layouts for clarity.

Also tour of scala layout includes the noscript notice for code tabs

Screenshot 2022-11-16 at 12 08 37

PR #2638, opened 2022-11-17
Fix orphan translation

artemkorsakov:
List of languages are missed on all translated pages.
For example: https://docs.scala-lang.org/ru/scala3/book/first-look-at-types.html

Maybe it's because of this commit:
https://github.com/scala/docs.scala-lang/pull/2634/commits/b6f198bf8d33bb3369efd17327bed0d4106e2b06

Fixed! Please, check it!


bishabosha (approved):
Thank you for noticing and fixing this so quickly!

PR #2639, opened 2022-11-18
fix language dropdown for getting-started pages

bishabosha:
`singlepage-overview` layout has to dynamically lookup the English page from translations to find the list of other languages, but it is hardcoded to look inside the `_overviews` collection, so is broken when used as a layout on any translated page where the English page is not in the `_overviews` collection.

I did not find a way to systematically relate e.g. `_es/getting-started/index.md` to `_getting-started/index.md`, when we also have to relate `_zh-cn/tutorials/scala-for-java-programmers.md` to `_overviews/tutorials/scala-for-java-programmers.md` so the simplest thing is to move the source files for `_getting-started` to `overviews/getting-started`.

before:
Screenshot 2022-11-18 at 10 55 56

after:
Screenshot 2022-11-18 at 10 56 07

alternatives:
- making a new layout for getting-started - but why when we are not changing the styling, just more fragile things people must remember
- adding more metadata to front matter of getting-started translations to say which collection should be searched - not scalable and prone to people forgetting


julienrf (approved):
Great idea!

PR #2644, opened 2022-11-19
add code tabs to num30

benluo:


_overviews/scala3-book/fun-eta-expansion.md
```diff
@@ -67,22 +90,29 @@ Unapplied methods are only converted to functions when a function type is expect
...
+{% endtab %}
+
+{% tab 'Scala 3' for=fun_5 %}
 
 ```scala
 val functions = List(isEvenFunction)   // works
 val methods = List(isEvenMethod)       // works
 ```
 
+As shown in that error message, there is a manual way to convert a method into a function in Scala 2, but the important part for Scala 3 is that the Eta Expansion technology is improved, so now when you attempt to use a method as a variable, it just works---you don’t have to handle the manual conversion yourself:
```
bishabosha, line 104: This reference to an error does not make sense because the error message is not visible unless they click on the scala 2 tab

benluo: resolved

bishabosha (approved):
Thank you for contributing

PR #2645, opened 2022-11-19
Add newest version for scala-parser-combinators

plopez26DPU:
The previous version suggested was causing compilation issues.


bishabosha (approved):
thanks!

PR #2646, opened 2022-11-20
Add subsection "Default Parameters for Overloaded Methods"

KarahanS:
Add a subsection to [Default Parameter Values](https://docs.scala-lang.org/tour/default-parameter-values.html) as to why we can't use default parameters for more than one overloaded method. Sources:
* https://groups.google.com/g/scala-user/c/FyQK3-cqfaY/m/fXLHr8QsW_0J
* https://stackoverflow.com/a/10927256/16530078


KarahanS:
Provided an example of overloaded functions with default parameters. It failed the `mdoc` check since having overloaded functions with default parameters for both of them is not allowed in Scala (that was the whole point though) - So I converted my example from `scala mdoc` to simply `scala`.

_tour/default-parameter-values.md
```diff
@@ -62,3 +62,24 @@ public class Main {
 ```
 {% endtab %}
 {% endtabs %}
+
+### Default Parameters for Overloaded Methods
+
+It's important to notice that Scala doesn't allow having two methods with default parameters and with the same name (overloaded). The reason is to keep a deterministic naming-scheme for the generated methods which return default arguments.
```
bishabosha, line 68: > The reason is to keep a deterministic naming-scheme for the generated methods which return default arguments.

This is an implementation detail which could change in the future

```suggestion
It's important to notice that Scala doesn't allow having two methods with default parameters and with the same name (overloaded).
```

_tour/default-parameter-values.md
```diff
@@ -62,3 +62,24 @@ public class Main {
...
+
+It's important to notice that Scala doesn't allow having two methods with default parameters and with the same name (overloaded).
+
+For example, if you write `def f(a: Int = 1)`, the compiler generates `def f$default$1 = 1`. If you have two overloads with defaults on the same parameter position, then we would need a different naming scheme. In this situation, it is preferred to keep the generated byte-code stable over multiple compiler runs.
+
+The other reason is to avoid the ambiguity that can be caused due to the existence of default parameters. To illustrate the problem, let's consider the method declarations provided below:
+
+{% tabs default-parameter-values-5 %}
+{% tab 'Scala 2 and 3' for=default-parameter-values-5 %}
+```scala
```
bishabosha, line 76: This will check in the CI that the snippet fails to compile
```suggestion
```scala mdoc:fail
```

_tour/default-parameter-values.md
```diff
@@ -62,3 +62,24 @@ public class Main {
...
 {% endtab %}
 {% endtabs %}
+
+### Default Parameters for Overloaded Methods
+
+It's important to notice that Scala doesn't allow having two methods with default parameters and with the same name (overloaded).
+
+For example, if you write `def f(a: Int = 1)`, the compiler generates `def f$default$1 = 1`. If you have two overloads with defaults on the same parameter position, then we would need a different naming scheme. In this situation, it is preferred to keep the generated byte-code stable over multiple compiler runs.
+
+The other reason is to avoid the ambiguity that can be caused due to the existence of default parameters. To illustrate the problem, let's consider the method declarations provided below:
```
bishabosha, line 72:
```suggestion
An important reason why is to avoid the ambiguity that can be caused due to the existence of default parameters. To illustrate the problem, let's consider the method declarations provided below:
```

_tour/default-parameter-values.md
```diff
@@ -62,3 +62,24 @@ public class Main {
 ```
 {% endtab %}
 {% endtabs %}
+
+### Default Parameters for Overloaded Methods
+
+It's important to notice that Scala doesn't allow having two methods with default parameters and with the same name (overloaded).
+
+For example, if you write `def f(a: Int = 1)`, the compiler generates `def f$default$1 = 1`. If you have two overloads with defaults on the same parameter position, then we would need a different naming scheme. In this situation, it is preferred to keep the generated byte-code stable over multiple compiler runs.
```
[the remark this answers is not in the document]
KarahanS: Sorry, which line would you remove? Is the example too related with a low-level design choice that may change in the future? If so I can remove the second paragraph.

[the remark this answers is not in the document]
bishabosha: > For example, if you write `def f(a: Int = 1)`, the compiler generates `def f$default$1 = 1`. If you have two overloads with defaults on the same parameter position, then we would need a different naming scheme. In this situation, it is preferred to keep the generated byte-code stable over multiple compiler runs.

Sorry yes to be clear, line 70 doesnt say much when you remove the low level implementation detail, so I think the code example line 77 is clear enough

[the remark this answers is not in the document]
KarahanS: Got it! Removed that part.

_tour/default-parameter-values.md
```diff
@@ -62,3 +62,24 @@ public class Main {
...
+{% tabs default-parameter-values-5 %}
+{% tab 'Scala 2 and 3' for=default-parameter-values-5 %}
+```scala mdoc:fail
+object A {
+  def func(x: Int = 34): Unit
+  def func(y: String = "abc"): Unit
+}
+```
+{% endtab %}
+{% endtabs %}
```
bishabosha, line 80: This final change is just for consistency
~~~suggestion
{% tabs default-parameter-values-5 class=tabs-scala-version %}
{% tab 'Scala 2' %}
```scala mdoc:fail
object A {
  def func(x: Int = 34): Unit
  def func(y: String = "abc"): Unit
}
```
{% endtab %}
{% tab 'Scala 3' %}
```scala
object A:
  def func(x: Int = 34): Unit
  def func(y: String = "abc"): Unit
```
{% endtab %}
{% endtabs %}
~~~

bishabosha (approved):
Thanks a lot!

PR #2647, opened 2022-11-21
translated according to the newest version.

benluo:


bishabosha (approved):
rendered output looks ok, thanks

PR #2649, opened 2022-11-24
Fix actual label for the translated pages

artemkorsakov:
[This commit](https://github.com/artemkorsakov/docs.scala-lang/commit/5a01937a15e806c3e2b832994d0bd13302f3ff69) fixed the following:

```
Fix multiple highlighted headlines in TOC
Previously, multiple equally named headlines were highlighted as the
current page, if the name was the same as on the actually current page.
This issued showed up in the book, where for instance "Summary" was used
more than once. Instead, here I am using the page ordinal instead of the
title.

It should be safe to assume that the page has a number, since otherwise
it would not show up in the TOC, IIUC.
```

But unfortunately only the English versions of the pages were corrected.

The bug remained in the other versions:
Screenshot_1

Fixed for the translated pages.


bishabosha (approved):
Thanks! I tested on the zh-cn translation of scala 3 book, which has multiple summary pages.

PR #2657, opened 2022-12-09
Tweak FAQ for placeholder syntax

### LINKED ISSUE #2381: Placeholder FAQ claim overreaches
## Reproduction steps

```scala
scala 2.13.8> List(42).flatMap(List(_ + 1))
                                    ^
              error: missing parameter type for expanded function ((<x$1: error>) => x$1.$plus(1))

scala 2.13.8> List(42).flatMap(List apply _ + 1)
val res1: List[Int] = List(43)
```

## Problem

The [FAQ](https://docs.scala-lang.org/tutorials/FAQ/index.html#why-doesnt-my-function-literal-with-_-in-it-work) claims the function can't be written with placeholder syntax.

som-snytt:
Fixes #2381 

Addresses `xs.map(_)`.

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -177,18 +177,46 @@ according to the situation, it usually means _"anything"_.
 
 ### Why doesn't my function literal with `_` in it work?
 
-Not all function literals (aka lambdas) can be expressed with the `_`
-syntax.
+Not all function literals (aka lambdas) can be expressed with underscores or `_`,
+which is called placeholder syntax.
 
-Every occurrence of `_` introduces a new variable.  So `_ + _` means
-`(x, y) => x + y`, not `x => x + x`.  The latter function cannot be
-written using the `_` syntax.
+The underscore or `_` takes the place of a single parameter, which is used exactly once.
 
-Also, the scope of `_` is always the smallest enclosing expression.
-The scope is determined purely syntactically, during parsing, without
-regard to types. So for example, `foo(_ + 1)` always means `foo(x =>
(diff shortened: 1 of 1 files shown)


SethTisue:
What you've written is good, but this isn't what the FAQ is for. The FAQ exists to host _very short_ answers and/or links to content hosted elsewhere.

If you feel the existing answer can be improved without making it longer, I'd merge that. And if you host a longer writeup elsewhere and feel we should link to it, I could do that too.

But as it stands, this isn't a direction I want the FAQ to take.

SethTisue:
(Part of the problem is that a lot of the material you've added doesn't address questions that are actually frequently asked. The fact that topic X is frequently asked about doesn't mean that the best answer is something that goes into every nuance about X.)

SethTisue:
One further thought: I acknowledge that it's perhaps sad or not ideal that there is such a gulf between the brief style of documentation in the language tour and the very complete and technical style in the language specification. But even if that gap could or should be filled somehow, it isn't the FAQ's place to do it.

SethTisue:
(cc @bishabosha, since this is good context to have for anyone who might be reviewing FAQ changes)

som-snytt:
There might be a SO entry, but "why doesn't map(_) mean map(x => x)" is a FAQ. Maybe it should be a FAQ with an answer but no explanation. Not sure if there is currently an explanation short of the spec.

som-snytt:
Or instead of adding to the FAQ, we could start a section, "QOTOA" questions one tires of answering.

som-snytt:
Worth adding that the original "One-Question FAQ" was a thorough explanation.

Oh, the current FAQ found a clever solution to accommodate it:

image

PR #2658, opened 2022-12-12
Revert "Scope of name"

SethTisue:
This reverts commit f19728cccca2d94782c2cd044aca4546815e2aab.

For context, see my remarks about the scope of the FAQ at https://github.com/scala/docs.scala-lang/pull/2657 .

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -321,27 +321,7 @@ In many cases one should instead write:
 
     ThisBuild / scalaVersion := "2.13.10"
 
-Conversely, you should not write:
-
-    ThisBuild / name := "sample"
-
-which will result in a warning:
-
-    [warn] there's a key that's not used by any other settings/tasks:
-    [warn]
-    [warn] * ThisBuild / name
-    [warn]   +- sample-project/build.sbt:11
-
-For your quick single-project build with bare settings, the minimal settings are:
-
-    scalaVersion := "2.13.10"
-    organization := "sampler"
-    version := "0.1"
-    name := "sample"          // must not be scoped to ThisBuild
-
-    scalacOptions ++= Seq("-Werror", "-Xlint")  // append to options
-
-Other possible solutions to provide a setting everywhere include:
+Other possibilities include:
 
 * the common settings pattern, where you put shared settings
   in a `val`, typically named `commonSettings`, and then


julienrf:
Since 'name' is a common setting, I think it would be helpful to still have a sentence saying that it is always scoped to a project and not to a wider scope such as the build or the global scope.

SethTisue:
@julienrf in all these years, I have never seen or heard anyone ask about the scoping of `name`? whereas the original FAQ entry is very much something that comes up over and over and over again with real users

I think we should resist the urge to include tangentially related, possibly-relevant, possibly-irrelevant information in the FAQ and stick only to concisely present only the highest-value information — even if that information is a link to relevant documentation with a fuller discussion of nuances and alternate scenarios

I think that's an especially good guideline to apply in this case  since when you get the scoping of `name` wrong, sbt warns you; the user won't be at sea. in theory, they might not know what the warning means, and in theory, they might come to our chat rooms and forums to ask what it means, but in practice I have not observed any such users to exist

cc @som-snytt

som-snytt:
FTR, I added info that I ask frequently. Finding it in the docs is non-trivial. I think this one was, what's the simplest build.sbt without remembering how to use the giter8 thing or whatever it's cleverly called. When I followed the FAQ, I got it wrong.

I agree the FAQ should be relatively short and sweet. Maybe it can have "more details" links or folds.

By the time we settle on an ideal compromise of pithy vs complete, we'll be able to ask ChatGPT.

SethTisue:
> what's the simplest build.sbt 

I think that's a remarkably subtle and even [contentious](https://github.com/scala/scala-seed.g8/issues/2) question. I often use none at all, or only a bare `scalaVersion` setting. https://www.scala-sbt.org/1.x/docs/sbt-by-example.html shows a sample progression.

Setting `organization`, `name`, and `version` is only necessary or desirable if you're going to publish, and publishing is a big subject that's already well covered by https://www.scala-sbt.org/1.x/docs/Publishing.html , I think.

som-snytt:
I think I had to test something locally, so I had to publish an example jar locally.  I have read the linked quick-start, but sbt docs are long-form reading. After years of "how do I?" with sbt, I still dread trying to find the one line of doc to tell me something.

> get the scoping of name wrong, sbt warns you

The other goal is to help people _before_ they throw the laptop across the room.

PR #2659, opened 2022-12-15
Merged the Type Class lessons and moved the Extension Methods lesson

alvinj:
I merged the two Type Class lessons, primarily by copying content from one into the other. I also moved the Extension Methods lesson to be the first Contextual Abstractions lesson. Then I updated all the relevant file numbers (`num:` field) and next/previous fields, and updated the Contextual Abstractions introduction to reflect the Extension Methods lesson being the first lesson in this section.

It’s possible that these lessons could all be reordered, but I didn’t know what the right answer is yet, so I stopped there.


bishabosha (approved):
Thank you, I think this makes a lot of sense to merge the pages :)

PR #2661, opened 2022-12-20
fix link to correct address.

doehyunbaek:
Two links in scala-for-java-programmers.md were misconfigured. I tested the change locally.


julienrf (approved):
Good catch, thank you for the fix! :pray:

PR #2662, opened 2022-12-21
Document the pattern for evolving case classes in a compatible manner

sideeffffect:
Documenting the patter coming from _SIP-50 - Struct Classes_ https://github.com/scala/improvement-proposals/pull/50#issuecomment-1361474980
English is not my first language, so I'd be very grateful for suggestions on how to improve the wording, etc.

Rendered:
https://github.com/sideeffffect/docs.scala-lang/blob/patch-1/_overviews/tutorials/binary-compatibility-for-library-authors.md#evolving-code-without-breaking-binary-compatibility

### DIFF
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -1111,7 +1111,7 @@ As mentioned, case classes support functional programming (FP):
 
 Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
 
-To achieve that, use this pattern:
+To achieve that, follow this pattern:
  * make the constructor `private`
  * define `private` `unapply` function in the companion object
  * define `withXXX` methods on the case class that create a new instance with the respective field changed
@@ -1153,7 +1153,7 @@ object Person:
 {% endtab %}
 {% endtabs %}
 
-The original users can use the case class `Person` as before, all the methods that existed before are present after this change, thus the compatibility with them is maintained.
+The original users can use the case class `Person` as before, all the methods that existed before are present unmodified after this change, thus the compatibility with the users is maintained.
 
 A regular case class not following this pattern would break its users, because by adding a new field some methods (which could be used by somebody else) change, for example `copy` or the constructor itself.
 
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -1107,56 +1107,6 @@ As mentioned, case classes support functional programming (FP):
   This process can be referred to as “update as you copy.”
 - Having an `unapply` method auto-generated for you also lets case classes be used in advanced ways with pattern matching.
 
-### Changing case class definition in a compatible manner
-
-Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
-
-To achieve that, follow this pattern:
- * make the constructor `private`
- * define `private` `unapply` function in the companion object
- * define `withXXX` methods on the case class that create a new instance with the respective field changed
- * define custom `apply` factory method(s) in the companion object (these can use the private constructor)
-
-Example:
-
-{% tabs case_class_compat_1 %}
-{% tab 'Scala 3 Only' %}
-
-```scala
-case class Person private (name: String, age: Int):
-  def withName(name: String) = copy(name = name)
-  def withAge(age: Int) = copy(age = age)
-
-object Person:
-  def apply(name: String, age: Int) = new Person(name, age)
-  private def unapply(p: Person) = p
-```
-{% endtab %}
-{% endtabs %}
-
-Later in time, you can ammend the original case class definition. You
- * add a new field `address`,
- * add a custom `withAddress` method and
- * add an `apply` factory method to the companion.
-
-{% tabs case_class_compat_2 %}
-{% tab 'Scala 3 Only' %}
-```scala
-case class Person private (name: String, age: Int, address: String = ""):
-  ...
-  def withAddress(address: String) = copy(address = address)
-
-object Person:
-  ...
-  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
-```
-{% endtab %}
-{% endtabs %}
-
-The original users can use the case class `Person` as before, all the methods that existed before are present unmodified after this change, thus the compatibility with the users is maintained.
-
-A regular case class not following this pattern would break its users, because by adding a new field some methods (which could be used by somebody else) change, for example `copy` or the constructor itself.
-
 {% comment %}
 NOTE: We can use this following text, if desired. If it’s used, it needs to be updated a little bit.
 
--- a/_overviews/scala3-book/domain-modeling-tools.md
+++ b/_overviews/scala3-book/domain-modeling-tools.md
@@ -1107,6 +1107,56 @@ As mentioned, case classes support functional programming (FP):
   This process can be referred to as “update as you copy.”
 - Having an `unapply` method auto-generated for you also lets case classes be used in advanced ways with pattern matching.
 
+### Changing case class definition in a compatible manner
+
+Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
+
+To achieve that, use this pattern:
+ * make the constructor `private`
+ * define `private` `unapply` function in the companion object
+ * define `withXXX` methods on the case class that create a new instance with the respective field changed
+ * define custom `apply` factory method(s) in the companion object (these can use the private constructor)
+
+Example:
+
+{% tabs case_class_compat_1 %}
+{% tab 'Scala 3 Only' %}
+
+```scala
+case class Person private (name: String, age: Int):
+  def withName(name: String) = copy(name = name)
+  def withAge(age: Int) = copy(age = age)
+
+object Person:
+  def apply(name: String, age: Int) = new Person(name, age)
+  private def unapply(p: Person) = p
+```
+{% endtab %}
+{% endtabs %}
+
+Later in time, you can ammend the original case class definition. You
+ * add a new field `address`,
+ * add a custom `withAddress` method and
+ * add an `apply` factory method to the companion.
+
+{% tabs case_class_compat_2 %}
+{% tab 'Scala 3 Only' %}
+```scala
+case class Person private (name: String, age: Int, address: String = ""):
+  ...
+  def withAddress(address: String) = copy(address = address)
+
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
+```
+{% endtab %}
+{% endtabs %}
+
+The original users can use the case class `Person` as before, all the methods that existed before are present after this change, thus the compatibility with them is maintained.
+
+A regular case class not following this pattern would break its users, because by adding a new field some methods (which could be used by somebody else) change, for example `copy` or the constructor itself.
+
 {% comment %}
 NOTE: We can use this following text, if desired. If it’s used, it needs to be updated a little bit.
 
--- a/_overviews/tutorials/binary-compatibility-for-library-authors.md
+++ b/_overviews/tutorials/binary-compatibility-for-library-authors.md
@@ -180,7 +180,7 @@ Sometimes, it is desirable to change the definition of a case class (adding and/
 To achieve that, follow this pattern:
  * make the constructor private (this also makes private the `copy` method of the class)
  * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used in an extractor pattern match)
- * define `withXXX` methods on the case class that create a new instance with the respective field changed
+ * for all the fields, define `withXXX` methods on the case class that create a new instance with the respective field changed
  * define custom `apply` factory method(s) in the companion object (these can use the private constructor)
 
 Example:
--- a/_overviews/tutorials/binary-compatibility-for-library-authors.md
+++ b/_overviews/tutorials/binary-compatibility-for-library-authors.md
@@ -219,7 +219,7 @@ alice match
 ~~~
 Later in time, you can amend the original case class definition to, say, add an optional `address` field. You
  * add a new field `address` and a custom `withAddress` method,
- * add the former constructor signature as a secondary constructor, private to the companion object. This step is necessary because the public `apply` method in the companion object calls the former constructor, which was effectively public in the bytecode produced by the compiler.
+ * add the former constructor signature as a secondary constructor, private to the companion object. This step is necessary because the compilers currently emit the private constructors as public constructors in the bytecode (see [#12711](https://github.com/scala/bug/issues/12711) and [#16651](https://github.com/lampepfl/dotty/issues/16651)).
 
 {% tabs case_class_compat_2 %}
 {% tab 'Scala 3 Only' %}
--- a/_overviews/tutorials/binary-compatibility-for-library-authors.md
+++ b/_overviews/tutorials/binary-compatibility-for-library-authors.md
@@ -178,7 +178,7 @@ Again, we recommend using MiMa to double-check that you have not broken binary c
 Sometimes, it is desirable to change the definition of a case class (adding and/or removing fields) while still staying backwards-compatible with the existing usage of the case class, i.e. not breaking the so-called _binary compatibility_. The first question you should ask yourself is “do you need a _case_ class?” (as opposed to a regular class, which can be easier to evolve in a binary compatible way). A good reason for using a case class is when you need a structural implementation of `equals` and `hashCode`.
 
 To achieve that, follow this pattern:
- * make the primary constructor private (this makes private the `copy` method of the class)
+ * make the primary constructor private (this makes the `copy` method of the class private as well)
  * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used as an extractor in match expressions)
  * for all the fields, define `withXXX` methods on the case class that create a new instance with the respective field changed (you can use the private `copy` method to implement them)
  * create a public constructor by defining an `apply` method in the companion object (it can use the private constructor)
@@ -197,7 +197,7 @@ case class Person private (name: String, age: Int):
 
 object Person:
   // Create a public constructor (which uses the primary constructor)
-  def apply(name: String, age: Int) = new Person(name, age)
+  def apply(name: String, age: Int): Person = new Person(name, age)
   // Make the extractor private
   private def unapply(p: Person) = p
 ```
@@ -236,6 +236,7 @@ alice match
 
 Later in time, you can amend the original case class definition to, say, add an optional `address` field. You
  * add a new field `address` and a custom `withAddress` method,
+ * update the public `apply` method in the companion object to initialize all the fields,
  * tell MiMa to [ignore](https://github.com/lightbend/mima#filtering-binary-incompatibilities) changes to the class constructor. This step is necessary because MiMa does not yet ignore changes in private class constructor signatures (see [#738](https://github.com/lightbend/mima/issues/738)).
 
 {% tabs case_class_compat_4 %}
@@ -244,6 +245,10 @@ Later in time, you can amend the original case class definition to, say, add an 
 case class Person private (name: String, age: Int, address: Option[String]):
   ...
   def withAddress(address: Option[String]) = copy(address = address)
+
+object Person:
+  // Update the public constructor to also initialize the address field
+  def apply(name: String, age: Int): Person = new Person(name, age, None)
 ```
 {% endtab %}
 {% endtabs %}
@@ -286,6 +291,22 @@ println(bob.address)
 {% endtabs %}
 
 A regular case class not following this pattern would break its usage, because by adding a new field changes some methods (which could be used by somebody else), for example `copy` or the constructor itself.
+
+Optionally, you can also add overloads of the `apply` method in the companion object to initialize more fields
+in one call. In our example, we can add an overload that also initializes the `address` field:
+
+{% tabs case_class_compat_7 %}
+{% tab 'Scala 3 Only' %}
+~~~ scala
+object Person:
+  // Original public constructor
+  def apply(name: String, age: Int): Person = new Person(name, age, None)
+  // Additional constructor that also sets the address
+  def apply(name: String, age: Int, address: String): Person =
+    new Person(name, age, Some(address))
+~~~
+{% endtab %}
+{% endtabs %}
 
 ## Versioning Scheme - Communicating compatibility breakages
 
--- a/_overviews/tutorials/binary-compatibility-for-library-authors.md
+++ b/_overviews/tutorials/binary-compatibility-for-library-authors.md
@@ -227,7 +227,7 @@ Later in time, you can amend the original case class definition to, say, add an 
 case class Person private (name: String, age: Int, address: Option[String]):
   ...
   // Add back the former primary constructor signature
-  private[Person] def this(name: String, age: Int): Person = this(name, age, None)
+  private[Person] def this(name: String, age: Int) = this(name, age, None)
   def withAddress(address: Option[String]) = copy(address = address)
 ```
 {% endtab %}
--- a/_overviews/tutorials/binary-compatibility-for-library-authors.md
+++ b/_overviews/tutorials/binary-compatibility-for-library-authors.md
@@ -199,7 +199,7 @@ object Person:
 {% endtab %}
 {% endtabs %}
 
-Later in time, you can ammend the original case class definition. You
+Later in time, you can amend the original case class definition. You
  * add a new field `address`,
  * add a custom `withAddress` method and
  * add an `apply` factory method to the companion.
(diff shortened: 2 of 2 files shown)


sideeffffect:
Also, do we have some part(s) of the documentation where we discuss the concept of "binary compatibility"? It would be desirable from them to this newly created section.

sjrd:
https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html

bjornregnell:
I think it should be explained that the ability use pattern matching is lost if no unapply?

sjrd:
IMO this whole section belongs the guide on compatibility for library authors. Application authors never need to care about this.

kubukoz:
nitpick: PR title has a typo (patter -> pattern)

_overviews/scala3-book/domain-modeling-tools.md
```diff
@@ -1107,6 +1107,56 @@ As mentioned, case classes support functional programming (FP):
...
+{% tabs case_class_compat_2 %}
+{% tab 'Scala 3 Only' %}
+```scala
+case class Person private (name: String, age: Int, address: String = ""):
+  ...
+  def withAddress(address: String) = copy(address = address)
+
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
```
kubukoz, line 1151: this isn't a binary compatible change, right? You're removing the method `apply` that takes 2 parameters.

kubukoz (changes requested):

_overviews/scala3-book/domain-modeling-tools.md
sideeffffect: Nope, you're not removing anything. You're only adding new stuff. That's the point.

Unless I missed something :sweat_smile:

sideeffffect:
Thanks for the feedback, I've reworked it.

_overviews/scala3-book/domain-modeling-tools.md
kubukoz: ahh sorry, I assumed this was replacing the `apply` present before. Of course you're right

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -167,12 +167,60 @@ For example, the use of these language features are a common source of binary co
...
 
 Again, we recommend using MiMa to double-check that you have not broken binary compatibility after making changes.
 
+### Changing case class definition in a compatible manner
+
+Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
+
+To achieve that, follow this pattern:
+ * make the constructor `private`
+ * define `private` `unapply` function in the companion object (note that by doing that the case class looses the ability to be used in a pattern match)
```
kubukoz, line 182:
```suggestion
 * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used in a pattern match)
```

typo (loses) + "private" can be considered part of "private ... function" so I wouldn't use backticks here (it could also give a false impression that `private unapply` is valid syntax)

kubukoz: (I'd untick the line above as well)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
+```
+{% endtab %}
+{% endtabs %}
+
+The original users can use the case class `Person` as before, all the methods that existed before are present unmodified after this change, thus the compatibility with the users is maintained.
+
+A regular case class not following this pattern would break its users, because by adding a new field some methods (which could be used by somebody else) change, for example `copy` or the constructor itself.
```
Philippus, line 222: "would break its users" sounds a bit strange, don't have an alternative yet though.

Philippus (commented):
Maybe add a note on the "deprecated" annotation?

_overviews/tutorials/binary-compatibility-for-library-authors.md
bjornregnell: "would brake usage" or "would break compatible usage" or similar perhaps?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
 
 Again, we recommend using MiMa to double-check that you have not broken binary compatibility after making changes.
 
+### Changing case class definition in a compatible manner
+
+Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
+
+To achieve that, follow this pattern:
+ * make the constructor private
+ * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used in a pattern match)
```
bjornregnell, line 182: Well usage in patterns is possible e.g. type patterns `case p: Person =>` so it is "only" constructor patterns that are not possible if I'm correct? So `case Person(n,a) => ` would not work as `unapply` is private, but other patterns work. So I think this should be clarified as the reader might not know about all the unapply mechanics under the hood...

bjornregnell: Edit: I guess extractor patterns could be made to work if the library author provides custom extractors?

bjornregnell: Anyway, I think it would be good to explain the implications on pattern matching in more detail...

sideeffffect:
I've reworded the mentioned parts.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
 
 Again, we recommend using MiMa to double-check that you have not broken binary compatibility after making changes.
 
+### Changing case class definition in a compatible manner
```
julienrf, line 176:
```suggestion
### Changing a case class definition in a backwards-compatible manner
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+
+Sometimes it is desirable to be able to change the definition of a case class (adding and/or removing fields) while still staying compatible with the users of the case class, i.e. not breaking the so called _binary compatibility_.
```
julienrf, line 178:
```suggestion
Sometimes, it is desirable to change the definition of a case class (adding and/or removing fields) while still staying backwards-compatible with the existing usage of the case class, i.e. not breaking the so-called _binary compatibility_.
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+
+To achieve that, follow this pattern:
+ * make the constructor private
```
julienrf, line 181:
```suggestion
 * make the constructor private (this also makes private the `copy` method of the class)
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+ * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used in an extractor pattern match)
+ * define `withXXX` methods on the case class that create a new instance with the respective field changed
```
julienrf, line 183:
```suggestion
 * for all the fields, define `withXXX` methods on the case class that create a new instance with the respective field changed
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+ * define custom `apply` factory method(s) in the companion object (these can use the private constructor)
+
+Example:
+
+{% tabs case_class_compat_1 %}
+{% tab 'Scala 3 Only' %}
+
+```scala
+case class Person private (name: String, age: Int):
+  def withName(name: String) = copy(name = name)
```
julienrf, line 193:
```suggestion
  def withName(name: String): Person = copy(name = name)
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+  def withAge(age: Int) = copy(age = age)
```
julienrf, line 194:
```suggestion
  def withAge(age: Int): Person = copy(age = age)
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+object Person:
+  def apply(name: String, age: Int) = new Person(name, age)
+  private def unapply(p: Person) = p
+```
+{% endtab %}
+{% endtabs %}
+
+Later in time, you can ammend the original case class definition. You
```
julienrf, line 202:
```suggestion
Later in time, you can amend the original case class definition. You
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+ * add a new field `address`,
+ * add a custom `withAddress` method and
+ * add an `apply` factory method to the companion.
+
+{% tabs case_class_compat_2 %}
+{% tab 'Scala 3 Only' %}
+```scala
+case class Person private (name: String, age: Int, address: String = ""):
```
julienrf, line 210: I would not recommend using an empty `String` to model “the absence of a value”. Instead, I recommend using `Option[String]` and `None`:

```suggestion
case class Person private (name: String, age: Int, address: Option[String]):
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+  ...
+  def withAddress(address: String) = copy(address = address)
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
+```
```
julienrf, line 216: IIRC, that change will still trigger a MiMa error because the primary constructor, although private from the point of view of Scala, was not private at the bytecode level (this is an “implementation detail”). Therefore, you have several options here:
- add back the previous constructor signature as a secondary constructor:
  ~~~ scala
  case class Person private (name: String, age: Int, address: Option[String]):
    def this(name: String, age: Int): Person = this(name, age, None)
  ~~~
- add an exclusion filter to the MiMa settings (in such a case, the filter is not 100% safe because there could be existing _Java_ code calling the previous constructor that will break when run with the new class definition)

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,54 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+
+{% tabs case_class_compat_2 %}
+{% tab 'Scala 3 Only' %}
+```scala
+case class Person private (name: String, age: Int, address: String = ""):
+  ...
+  def withAddress(address: String) = copy(address = address)
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, address)
```
julienrf, line 215: I think you should also say that the previous `apply` method needs to be adapted:

~~~ diff
- def apply(name: String, age: Int) = new Person(name, age)
+ def apply(name: String, age: Int) = new Person(name, age, None)
~~~

julienrf (commented):
Thank you very much @sideeffffect for creating the PR. Here are my suggestions.

Also, before we merge that it would be better to test that our recommendation does work as expected. We should create a small project with the former class definition, and the new class definition in a separate module, and check the output of MiMa.

_overviews/tutorials/binary-compatibility-for-library-authors.md
sideeffffect: So it turns out we have to do 3 things now:
 * add a new constructor, private to the companion object, with the original fields
 * add a custom `withAddress` method and
 * add an `apply` factory method to the companion.

This pattern doesn't seem as attractive as it did in the beginning... :confused:

_overviews/tutorials/binary-compatibility-for-library-authors.md
sideeffffect: It doesn't, because we manually re-introduce the constructor with the original fields.

_overviews/tutorials/binary-compatibility-for-library-authors.md
odersky: Or: Fix MiMa to not flag this pattern. That would be the correct thing to do, which might also help other code.

sjrd: The problem in this case is that Java code would see this as a perfectly normal constructor. No funny name, no synthetic or artifact flag. Just your regular, public constructor.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,56 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+  ...
+  private[Person] def this(name: String, age: Int): Person = this(name, age, None)
+  def withAddress(address: String) = copy(address = address)
+object Person:
+  ...
+  def apply(name: String, age: Int, address: String) = new Person(name, age, Some(address))
+```
+{% endtab %}
+{% endtabs %}
+
```
julienrf, line 221:
```suggestion

> Note that an alternative solution, instead of adding back the previous constructor signatures as secondary constructors, consists of adding a [MiMa filter](https://github.com/lightbend/mima#filtering-binary-incompatibilities) to simply ignore the problem. Even though the constructors are effectively public in the bytecode, they can’t be called from Scala programs (but they could be called by Java programs). In an sbt build definition you would add the following setting:
> ~~~ scala
> import com.typesafe.tools.mima.core._
> mimaBinaryIssueFilters += ProblemFilters.exclude[DirectMissingMethodProblem]("Person.this")
> ~~~
> Otherwise, MiMa would fail with an error like “method this(java.lang.String,Int)Unit in class Person does not have a correspondent in current version”.
```

julienrf (commented):
I’ve done a couple of tests to double-check that what we say is correct, and I propose some suggestions to elaborate on a few points. Please let me know what you think about them, I wonder if some of my suggestions may add more confusion instead of clarification…

_overviews/tutorials/binary-compatibility-for-library-authors.md
sideeffffect: "Private" primary constructors vs MiMa seems like a topic that needs further discussion.
I've kicked off https://contributors.scala-lang.org/t/private-primary-constructor-vs-mima/6050

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,85 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+> ~~~
+> Otherwise, MiMa would fail with an error like “method this(java.lang.String,Int)Unit in class Person does not have a correspondent in current version”.
+The original users can use the case class `Person` as before, all the methods that existed before are present unmodified after this change, thus the compatibility with the existing usage is maintained.
+
+The new field `address` can be used as follows:
+
+~~~ scala
+// The public constructor sets the address to None by default.
+// To set the address, we call withAddress:
+val bob = Person("Bob", 21).withAddress(Some("Atlantic ocean"))
```
sideeffffect, line 249: @julienrf Thanks for you suggestions. I've applied them all. But I think this part is worth a discussion.

Shouldn't we recommend readers to also include the `apply` methods in the companion object? Because
```scala
Person("Bob", 21).withAddress(Some("Atlantic ocean"))
```
looks a bit clumsy. This looks much better
```scala
Person("Bob", 21, "Atlantic ocean")
```
I know, we save on some boilerplate in the class definition. But we're just pushing the boilerplate onto the users of the class. Is that the right choice to make?

Ideally we'd have our cake and eat it too, but I don't know if it's possible without a new language level feature...

sideeffffect:
There are parts of this pattern which are pretty repetitive and very regular:

 * add a new constructor, private to the companion object, with the original fields (necessary, unless we want to deal with MiMa)
 * add an apply factory method to the companion (necessary, unless we want to burden the class users with boilerplate)

Are these parts automatable using Macros? We would need the macro(s) to generate as many of these members as there are `Option`s in the case class (well, only the `Option`s that are on the right side of the case class constructor definition).

sjrd:
Anything that needs to add members visible during typechecking cannot be done with macros.

_overviews/tutorials/binary-compatibility-for-library-authors.md
julienrf: I don’t think the pattern I suggested is clumsy. It is similar to the “builder” pattern we often see around.

Anyway, there are probably several possible variations. The one I suggest here provides a public-facing API that (I think) is simpler because it contains fewer overloads of the `apply` methods.

bjornregnell: Well, documentation space is not scarce :) so I suggest to explain both possibilities: withX or overloaded apply constructors. It's up to the library author. And if the library user is "refused" a more concise constructor then it's just a self-made extension away...

julienrf:
@sideeffffect do you want to rework some parts or should we merge it as it is? (I am fine with both)

sideeffffect:
I thought we want to first finish the _MiMa vs "private" primary constructors_ debate.

https://contributors.scala-lang.org/t/private-primary-constructor-vs-mima/6050/5

But I'm not against merging this documentation before. What do you guys think?

julienrf:
That’s a good point.

@smarter, what do you think about Sébastien’s suggestion of emitting the constructors as “package private” constructors? If we can do that, then I guess we would not need to add the MiMa exception anymore.

smarter:
That would also be a binary breaking change which I assume mima would flag unless we special-case it in mima.

julienrf:
I assume MiMa already treats package private definitions like private definitions. If this is not the case, that should be fixed indeed. But then MiMa would not flag changes to the class constructors if they are emitted as package private. Of course, code compiled with a new version of the compiler (emitting private case class constructors as package private constructors at the bytecode level) would be binary incompatible with code compiled with a previous compiler, so library authors would have to add a MiMa exception for that. But at least, the emitted bytecode would be more correct because it would become impossible to call the constructor from Java code.

julienrf:
I created https://github.com/scala/bug/issues/12711 and https://github.com/lampepfl/dotty/issues/16651 to track the issue with private constructors that are still public.

We could add a note about that in the documentation, what do you think?

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,85 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+
+If you try to use `Person` as an extractor in a match expression, it will fail with a message like “method unapply cannot be accessed as a member of Person.type”. Instead, you can use it as a typed pattern:
+
+~~~ scala
+alice match
+  case person: Person => person.name
+~~~
+Later in time, you can amend the original case class definition to, say, add an optional `address` field. You
+ * add a new field `address` and a custom `withAddress` method,
+ * add the former constructor signature as a secondary constructor, private to the companion object. This step is necessary because the public `apply` method in the companion object calls the former constructor, which was effectively public in the bytecode produced by the compiler.
```
julienrf, line 222:
```suggestion
 * add the former constructor signature as a secondary constructor, private to the companion object. This step is necessary because the compilers currently emit the private constructors as public constructors in the bytecode (see [#12711](https://github.com/scala/bug/issues/12711) and [#16651](https://github.com/lampepfl/dotty/issues/16651)).
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,85 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+{% tab 'Scala 3 Only' %}
+```scala
+case class Person private (name: String, age: Int, address: Option[String]):
+  ...
+  // Add back the former primary constructor signature
+  private[Person] def this(name: String, age: Int) = this(name, age, None)
+  def withAddress(address: Option[String]) = copy(address = address)
+```
+{% endtab %}
+{% endtabs %}
```
julienrf, line 234:
```suggestion
{% tabs case_class_compat_2 class=tabs-scala-version %}
{% tab 'Scala 2' %}
~~~ scala
case class Person private (name: String, age: Int, address: Option[String]) {
  ...
  // Add back the former primary constructor signature
  private[Person] def this(name: String, age: Int) = this(name, age, None)
  def withAddress(address: Option[String]) = copy(address = address)
}
~~~
{% endtab %}
{% tab 'Scala 3' %}
~~~ scala
case class Person private (name: String, age: Int, address: Option[String]):
  ...
  // Add back the former primary constructor signature
  private[Person] def this(name: String, age: Int) = this(name, age, None)
  def withAddress(address: Option[String]) = copy(address = address)
~~~
{% endtab %}
{% endtabs %}
```

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,132 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+ * define a private `unapply` function in the companion object (note that by doing that the case class loses the ability to be used as an extractor in match expressions)
+ * for all the fields, define `withXXX` methods on the case class that create a new instance with the respective field changed (you can use the private `copy` method to implement them)
+ * create a public constructor by defining an `apply` method in the companion object (it can use the private constructor)
+
+Example:
+
+{% tabs case_class_compat_1 class=tabs-scala-version %}
+
+{% tab 'Scala 2' %}
+~~~ scala
```
sideeffffect, line 191: I've noticed that the other tutorials use triple ` instead of triple ~.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,132 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+// Mark the primary constructor as private
+case class Person private (name: String, age: Int) {
+  // Create withXxx methods for every field, implemented by using the copy method
```
sideeffffect, line 194: Scala 2 is a whole different game. I think we also have to make the `copy` method private. And consistently across the changes to the case class.

_overviews/tutorials/binary-compatibility-for-library-authors.md
```diff
@@ -173,6 +173,132 @@ You can find detailed explanations, runnable examples and tips to maintain binar
...
+{% endtab %}
+{% endtabs %}
+Later in time, you can amend the original case class definition to, say, add an optional `address` field. You
+ * add a new field `address` and a custom `withAddress` method,
+ * add the former constructor signature as a secondary constructor, private to the companion object. This step is necessary because the compilers currently emit the private constructors as public constructors in the bytecode (see [#12711](https://github.com/scala/bug/issues/12711) and [#16651](https://github.com/lampepfl/dotty/issues/16651)).
+
+{% tabs case_class_compat_2 class=tabs-scala-version %}
+{% tab 'Scala 2' %}
+~~~ scala
+case class Person private (name: String, age: Int, address: Option[String]) {
```
sideeffffect, line 260: For Scala 2 we also have to replicate the "old" `apply` function in the companion object.

_overviews/tutorials/binary-compatibility-for-library-authors.md
julienrf: It does not really matter but ` didn’t work with GitHub’s suggestions.

sideeffffect: Fair enough :smile_cat: I don't seem to be able to figure out escaping in GitHub's MarkDown either...

_overviews/tutorials/binary-compatibility-for-library-authors.md
julienrf: Oh, good point, I forgot that `copy` is not private in Scala 2. I’ve reverted the changes that included Scala 2 then.

_overviews/tutorials/binary-compatibility-for-library-authors.md
julienrf: Reverted.

julienrf:
I did another pass and made some final adjustments. I think the guide is ready to go now.

Screenshot 2023-01-13 at 09-51-29 Binary Compatibility for library authors

bishabosha:
This doesn't mention defining a new apply method in the companion (and keeping the old one) after adding the address field

bjornregnell:
Minor thing, fix word ordering to be "this makes the copy method private".

julienrf:
Thank you for the review, @bishabosha and @bjornregnell, I’ve addressed your comments in a818702f

PR #2664, opened 2023-01-01
Add code tabs to fun-write-method-returns-function

doehyunbaek:
With c875c656a62e1a07cd09bdb5eb6d7650c3ca90ab I updated Gemfile.lock.

Is it needed or better not tracked in git? I assumed the former from the previous commits to the file and if i misunderstood it I'll revert and force push the change.

With 60a87c4d20434a17fea9cf1ba0bb9c24ca2a36ab, I updated code tab, per #2481.

I tested each code with 2.13.10 repl and 3.2.1 repl and tested the website with jekyll

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -94,4 +94,4 @@ DEPENDENCIES
   webrick (~> 1.7)
 
 BUNDLED WITH
-   2.4.2
+   2.3.10
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -54,6 +54,8 @@ GEM
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
     mercenary (0.4.0)
+    nokogiri (1.13.10-arm64-darwin)
+      racc (~> 1.4)
     nokogiri (1.13.10-x86_64-darwin)
       racc (~> 1.4)
     nokogiri (1.13.10-x86_64-linux)
@@ -81,6 +83,7 @@ GEM
     yell (2.2.2)
 
 PLATFORMS
+  arm64-darwin-22
   x86_64-darwin-21
   x86_64-linux
 
@@ -91,4 +94,4 @@ DEPENDENCIES
   webrick (~> 1.7)
 
 BUNDLED WITH
-   2.3.10
+   2.4.2
--- a/_overviews/scala3-book/fun-write-method-returns-function.md
+++ b/_overviews/scala3-book/fun-write-method-returns-function.md
@@ -20,48 +20,72 @@ Once again we start with a problem statement:
 Given that statement, you can start building `greet`.
 You know it’s going to be a method:
 
+{% tabs fun-write-method-returns-function-1 %}
+{% tab 'Scala 2 and 3' %}
 ```scala
 def greet()
 ```
+{% endtab %}
+{% endtabs %}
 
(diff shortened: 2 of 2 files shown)


_overviews/scala3-book/fun-write-method-returns-function.md
```diff
@@ -116,27 +156,53 @@ A first thing you know is that you want to create a method that (a) takes a “d
...
 val frenchGreeting = (name: String) => println(s"Bonjour, $name")
 ```
+{% endtab %}
+{% endtabs %}
 
 Now all you need to do is (a) return `englishGreeting` if the `desiredLanguage` is English, and (b) return `frenchGreeting` if the `desiredLanguage` is French.
 One way to do that is with a `match` expression:
 
+{% tabs fun-write-method-returns-function-14 class=tabs-scala-version %}
+{% tab 'Scala 2' for=fun-write-method-returns-function-14 %}
```
julienrf, line 193: We don’t need the `for` attribute anymore.

julienrf (approved):
Thank you very much @doehyunbaek for your help!

I just noticed one minor possible improvement.

_overviews/scala3-book/fun-write-method-returns-function.md
doehyunbaek: Done with [9651d75](https://github.com/scala/docs.scala-lang/pull/2664/commits/9651d757bd5163e9b1bbbc311117aaa908d8209d).

doehyunbaek:
I added the for attribute because of [this mention](https://github.com/scala/docs.scala-lang/blob/008fc4beee8b0d83ddaba4c2e1eabbc56a953003/_overviews/contribute/add-guides.md#L153-L162) and should it also be updated?

doehyunbaek:
I rebased onto origin/main and force-pushed branch.

Gemfile.lock
```diff
@@ -91,4 +94,4 @@ DEPENDENCIES
   webrick (~> 1.7)
 
 BUNDLED WITH
-   2.3.10
+   2.4.2
```
julienrf, line 97: I am not sure about this version change for the bundler. Here we use 2.3.10: https://github.com/scala/docs.scala-lang/blob/00302d7083b5320c54e65203cf7e16ee94fd3949/Dockerfile#L4

julienrf:
Thank you for updating the PR. I don’t think the documentation should be updated, the advice explaining the possibility of adding the `for` property when that helps readability still applies.

Gemfile.lock
doehyunbaek: Can you give me a brief advice on when to commit Gemfile.lock file change? It seems always is not the option. So is it never? or for some portion of the lock file(e.g. except bundler version?

doehyunbaek: Updated with [fc15ab3](https://github.com/scala/docs.scala-lang/pull/2664/commits/fc15ab3079958eb8354833bf980d39dab2f7be36), would still appreciate your advice.

julienrf: The changes to the lockfile should always be committed, but the file should be generated with the right version of the bundler. Could you please make sure to have bundler 2.3.10 on your system, or to use the Docker approach to automatically use the correct version of the bundler?

doehyunbaek: I actually have bundler 2.3.10 installed at my local mac. Strange it changed the lockfile. 

I will try to commit lockfile with correct bundler version as specified in Dockerfile and will use the Docker approach if I'm not able to do that.

PR #2666, opened 2023-01-03
find the corresponding english page to list languages

bishabosha:
This solves an issue that if the first alphabetically ordered page in a multipage-overview is not translated, then that language will not appear in the language dropdown.

Instead, we now find the english page that has the same `id` (read as language-independent URL) to source the translations for that individual page, rather than for the whole overview

This problem manifested in the `ru` translation of the scala 3 book, which does not have a translation for `_overviews/scala3-book/ca-context-bounds.md`, i.e. the first alphabetically ordered English page in the collection. See the change here:

before:
Screenshot 2023-01-03 at 16 20 13

after:
Screenshot 2023-01-03 at 16 20 28

### DIFF
--- a/_includes/sidebar-toc-multipage-overview.html
+++ b/_includes/sidebar-toc-multipage-overview.html
@@ -52,9 +52,12 @@
           {% endfor %}
         </ul>
       {% elsif page.language and orphanTranslation == false %}
+        {% comment %}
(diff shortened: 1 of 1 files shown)


julienrf (approved):
Thank you Jamie!

PR #2668, opened 2023-01-04
Correct complex number toString implementation

ahmedali-durga:
For Complex numbers with negative imaginary parts, existing representation doesn't include "-" and shows real and imaginary parts conjoined


Philippus (approved):

julienrf:
The current implementation does the right thing: https://scastie.scala-lang.org/H9wCd1X3Tl6dNPGodXFBuA

This is because when the imaginary part is negative, there is automatically a `-` sign printed before the number.

Philippus:
Indeed, I tried it out and agree with Julien.

PR #2669, opened 2023-01-07
Update _tour/pattern-matching.md

raulsorrentino:
The `VoiceRecording` case must not exist in the code to cause the warning.


julienrf (approved):
Indeed, sorry for the mistake, thank you for the fix!

PR #2692, opened 2023-02-06
docs: migrate CONTRIBUTING.md from Dotty repo

ckipp01:
This change adds the extra information that is found in
https://github.com/lampepfl/dotty/blob/main/CONTRIBUTING.md into the
actual Scala 3 Contributing guide in an effort to get all of this
information in a single place. After this is merge I'll do a follow-up
pr to the `CONTRIBUTING.md` in Dotty to then point to this guide.

### DIFF
--- a/_overviews/scala3-contribution/procedures-checklist.md
+++ b/_overviews/scala3-contribution/procedures-checklist.md
@@ -7,31 +7,124 @@ previous-page: procedures-testing
 next-page: arch-intro
 ---
 
-Once you solved an issue, you likely want to see your change added to the [Scala 3 repo][lampepfl/dotty].
-To do that, you need to prepare a [pull request][pull-request] with your changes. We recommend you
-follow these guidelines, [also consult the full requirements][full-list]:
+Once you solved the issue you were working on, you'll likely want to see your
+changes added to the [Scala 3 repo][lampepfl/dotty]. To do that, you need to
+prepare a [pull request][pull-request] with your changes. Assuming that the team
+is aware of what you've been working, here are some final steps that you'll want
+to keep in mind as you create your PR.
 
 ### 1. Sign the CLA
 
-Make sure you have signed the [Scala CLA][cla], if not, sign it.
+Make sure you have signed the [Scala CLA][cla]. If you have any questions about
+what this is and why it's required you can read further about it [here][cla].
 
-### 2: Is It Relevant?
+### 2. Make sure your work is on its own branch
 
-Before starting to work on a feature or a fix, it's good practice to ensure that:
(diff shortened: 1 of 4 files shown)


_overviews/scala3-contribution/procedures-checklist.md
```diff
@@ -7,35 +7,134 @@ previous-page: procedures-testing
...
+   by a detailed commit message for (as discussed in the following sections).
+   For more info, see the article: [Git Workflow][git-workflow]. Additionally,
+   every commit should be able to be used in isolation—that is, each commit must
+   build and pass all tests.
+
+2. The first line should be a descriptive sentence about what the commit is
+   doing. It should be possible to fully understand what the commit does by just
+   reading this single line. It is **not ok** to only list the ticket number,
+   type "minor fix" or similar. If the commit has a corresponding ticket,
+   include a reference to the ticket number, prefixed with "Closes #", at the
```
ckipp01, line 68: So maybe more of a process question. Is this done for specific reason? Like is there any sort of automation set up that needs this prefixed. The only reason I ask is that if you follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), which many projects do, this is actually supposed to be in the footer. I realize I used the footer method in my example so I either need to change my example or change this. Maybe @anatoliykmetyuk you know more about the Dotty process?

_overviews/scala3-contribution/start-intro.md
```diff
@@ -58,3 +79,7 @@ hello world
 [java11]: https://www.oracle.com/java/technologies/javase-jdk11-downloads.html
 [adopt]: https://adoptopenjdk.net/
 [compat]: /overviews/jdk-compatibility/overview.html
+[scala-cli]: https://www.lightbend.com/contribute/cla/scala
```
julienrf, line 82: Typo here

julienrf (approved):
Thank you @ckipp01 for unifying the documentation!

_overviews/scala3-contribution/procedures-checklist.md
anatoliykmetyuk: The reason we put the reference to the issue being fixed as the first line is purely historical. To my awareness, we do not follow the conventional commits format you linked.

anatoliykmetyuk (approved):

_overviews/scala3-contribution/procedures-checklist.md
ckipp01: Alright, then I'll just leave this as is. If there is no process, then slightly pushing people to a format this is followed elsewhere is probably a good idea.

Thanks!

PR #2691, opened 2023-02-05
translate some "untranslated" titles in Chinese.

benluo:

PR #2693, opened 2023-02-07
Zh cn/overviews/scala3 book/taste summary

benluo:


eed3si9n:
Considering [e720e7a](https://github.com/scala/docs.scala-lang/pull/2693/commits/e720e7a9c552508b0e7809a3cf950d9f6b87f31c) was merged as https://github.com/scala/docs.scala-lang/pull/2691, only the other commit needs to be reviewed, which looks ok.

eed3si9n (approved):

PR #2698, opened 2023-02-14
Fix setup instructions

julienrf:
I think these setup instructions are too complicated but we need to fix them so that they work with the current state of things.

Ideally, we should have a single script that would download the correct binary based on the user OS and architecture.

Screenshot 2023-02-14 at 10-19-27 Getting Started

Screenshot 2023-02-14 at 10-19-44 Getting Started

### DIFF
--- a/_overviews/getting-started/index.md
+++ b/_overviews/getting-started/index.md
@@ -51,7 +51,7 @@ Install it on your system with the following instructions.
 {% tab macOS for=install-cs-setup-tabs %}
 Run the following command in your terminal, following the on-screen instructions:
 {% include code-snippet.html language='bash' codeSnippet=site.data.setup-scala.macOS-brew %}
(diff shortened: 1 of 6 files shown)


_overviews/getting-started/index.md
```diff
@@ -52,15 +52,22 @@ Install it on your system with the following instructions.
 Run the following command in your terminal, following the on-screen instructions:
 {% include code-snippet.html language='bash' codeSnippet=site.data.setup-scala.macOS-brew %}
 {% altDetails cs-setup-macos-nobrew  "Alternatively, if you don't use Homebrew:" %}
```
bishabosha, line 54:
```suggestion
{% altDetails cs-setup-macos-nobrew  "Alternatively for Apple Silicon, or if you don't use Homebrew:" %}
```

bishabosha (approved):
I would approve, but for completeness the alt-details label should indicate that Apple Silicon users click that

bishabosha (approved):

PR #2700, opened 2023-02-15
fix: Not compiling example in Quotes page

Iltotore:
Fix the `mirrorFields[T]` example at line 351.

The following example:

```scala
def mirrorFields[T: Type](using Quotes): List[String] =
  Type.of[T] match
    case '[field *: fields] =>
      Type.show[field] :: mirrorFields[fields]
    case '[EmptyTuple] =>
      Nil
    case _ =>
      compiletime.error("Expected known tuple but got: " + Type.show[T])

mirrorFields[EmptyTuple]         // Nil
mirrorFields[(Int, String, Int)] // List("Int", "String", "Int")
mirrorFields[Tuple]              // error: Expected known tuple but got: Tuple
```

Does not compile because `compiletime.error` (which should be used in inline methods) is used instead of `quotes.reflect.report.errorAndAbort`.

The following example compiles (with `scala.quoted.*` imported):

```scala
def mirrorFields[T: Type](using Quotes): List[String] =
  Type.of[T] match
    case '[field *: fields] =>
      Type.show[field] :: mirrorFields[fields]
    case '[EmptyTuple] =>
      Nil
    case _ =>
      quotes.reflect.report.errorAndAbort("Expected known tuple but got: " + Type.show[T])
```

but it does not when used outside of a macro:
```scala
mirrorFields[EmptyTuple]         // Not found `Type[EmptyTuple]`
mirrorFields[(Int, String, Int)] // Not found `Type[EmptyTuple]`
mirrorFields[Tuple]              // Not found `Type[EmptyTuple]`
```

Because the given `Type` instance needs a `Quotes` instance:
```scala
[error] I found:
[error] 
[error]     quoted.Type.of[(Int, String, Int)](/* missing */summon[quoted.Quotes])
[error] 
[error] But no implicit values were found that match type quoted.Quotes.
[error]   println(mirrorFields[(Int, String, Int)])
```

The most minimal example I managed to make (the one committed) is this one:

```scala
inline def mirrorFields[T]: List[String] = ${mirrorFieldsImpl[T]}

def mirrorFieldsImpl[T: Type](using Quotes): Expr[List[String]] =
  
  def rec[A : Type]: List[String] = Type.of[A] match
    case '[field *: fields] =>
      Type.show[field] :: rec[fields]
    case '[EmptyTuple] =>
      Nil
    case _ =>
      quotes.reflect.report.errorAndAbort("Expected known tuple but got: " + Type.show[A])

  Expr(rec)
```
```scala
mirrorFields[EmptyTuple]         // Nil
mirrorFields[(Int, String, Int)] // List("scala.Int", "java.lang.String", "scala.Int")
mirrorFields[Tuple]              // error: Expected known tuple but got: Tuple
```

I quickly read the rest of the page and I might have found similar issues in other examples.

### DIFF
--- a/_overviews/scala3-macros/tutorial/quotes.md
+++ b/_overviews/scala3-macros/tutorial/quotes.md
@@ -349,17 +349,23 @@ Note that the previous case could also be written as `case '{ ($ls: List[u]).map
 Types represented with `Type[T]` can be matched on using the patten `case '[...] =>`.
 
 ```scala
-def mirrorFields[T: Type](using Quotes): List[String] =
-  Type.of[T] match
+inline def mirrorFields[T]: List[String] = ${mirrorFieldsImpl[T]}
+
+def mirrorFieldsImpl[T: Type](using Quotes): Expr[List[String]] =
+  
+  def rec[A : Type]: List[String] = Type.of[A] match
     case '[field *: fields] =>
-      Type.show[field] :: mirrorFields[fields]
+      Type.show[field] :: rec[fields]
     case '[EmptyTuple] =>
       Nil
     case _ =>
-      compiletime.error("Expected known tuple but got: " + Type.show[T])
+      quotes.reflect.report.errorAndAbort("Expected known tuple but got: " + Type.show[A])
 
+  Expr(rec)
+```
+```scala
 mirrorFields[EmptyTuple]         // Nil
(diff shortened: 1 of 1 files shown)


bishabosha (approved):
Thanks!

PR #2704, opened 2023-02-24
Improve static site documentation, siteroot

Dedelweiss:
The goal of this PR is improve the documentation to create static site. 
So that beginners have all the tools in hand to have a static site that they can then improve with the "settings" page.
- More explanation about site-root
    - How to setup
    - In build.sbt
    - Must be scoped to Compile / doc / scalacOptions

### DIFF
--- a/_overviews/scala3-scaladoc/static-site.md
+++ b/_overviews/scala3-scaladoc/static-site.md
@@ -38,7 +38,18 @@ getting-started.html
 Scaladoc can transform both files and directories (to organize your documentation into a tree-like structure). By default, directories have a title based on the file name and have empty content. It is possible to provide index pages for each section by creating `index.html` or `index.md` (not both) in the dedicated directory.
 
 Before generating your static site you need to set the `-siteroot` value in your doc `scalacOptions`. The value of this is the directory that holds your docs.  The root URL for the generated documentation will also be `<site-root>`.
-The configuration is as follows:
+
+For example if you have a directory called `docs` and you'd like that to be treated as your site root :
+
+```
+.
+└── docs/
+    └── _docs/
+        ├── index.html
+        └── getting-started.html
+```
+
+Then the configuration would be as follows:
 
 ```
 Compile / doc / scalacOptions ++= Seq("-siteroot", "docs")
--- a/_overviews/scala3-scaladoc/static-site.md
+++ b/_overviews/scala3-scaladoc/static-site.md
@@ -37,6 +37,15 @@ getting-started.html
 
 Scaladoc can transform both files and directories (to organize your documentation into a tree-like structure). By default, directories have a title based on the file name and have empty content. It is possible to provide index pages for each section by creating `index.html` or `index.md` (not both) in the dedicated directory.
 
+Before generating your static site you also need to set the `-siteroot` to the name of your `<site-root>`,which means that when you generate documentation for your project, the root URL for the generated documentation will be `<site-root>`.
+The configuration is as follows:
+
+```
+Compile / doc / scalacOptions ++= Seq("-siteroot", "docs")
+```
+
+Note that the `Compile / doc` part of the line refers to the `doc` task in the `Compile` configuration, which generates documentation for the project's main sources. So, the `scalacOptions` set in this line will only be used during the generation of documentation for the main sources, not during compilation or testing. Refer to the settings page for more information
+
 Keep in mind that viewing your site locally with all the features it offers, like search or snippets, require a
 local server. For example if your output directory was `output` you could use a python server to view everything
 by doing the following and opening `localhost:8080`:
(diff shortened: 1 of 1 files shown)


_overviews/scala3-scaladoc/static-site.md
```diff
@@ -37,6 +37,15 @@ getting-started.html
 
 Scaladoc can transform both files and directories (to organize your documentation into a tree-like structure). By default, directories have a title based on the file name and have empty content. It is possible to provide index pages for each section by creating `index.html` or `index.md` (not both) in the dedicated directory.
 
+Before generating your static site you also need to set the `-siteroot` to the name of your `<site-root>`,which means that when you generate documentation for your project, the root URL for the generated documentation will be `<site-root>`.
+The configuration is as follows:
```
ckipp01, line 41:
```suggestion
Before generating your static site you need to set the `-siteroot` value in your doc `scalacOptions`. The value of this is the directory that holds your docs.  The root URL for the generated documentation will also be `<site-root>`.
The configuration is as follows:
```

_overviews/scala3-scaladoc/static-site.md
```diff
@@ -37,6 +37,15 @@ getting-started.html
...
+
+```
+Compile / doc / scalacOptions ++= Seq("-siteroot", "docs")
+```
+
+Note that the `Compile / doc` part of the line refers to the `doc` task in the `Compile` configuration, which generates documentation for the project's main sources. So, the `scalacOptions` set in this line will only be used during the generation of documentation for the main sources, not during compilation or testing. Refer to the settings page for more information
```
ckipp01, line 47:
```suggestion
Note that the `Compile / doc` part of the line refers to the `doc` task in the `Compile` configuration, which generates documentation for the project's main sources. So, the `scalacOptions` set in this line will only be used during the generation of documentation for the main sources, not during compilation or testing. Refer to the settings page for more information
```

I actually don't think we really need to specify this here. It's pretty sbt specific, so I'm not sure if we should be explaining it here or not. However, if we do keep it there is a slight mistake in here that might cause even more confusion.

> not during compilation

Actually the `Compile / ` part of this is actually scoping it to that configuration like you said earlier. So it's actually applied to the `doc` task  _during_ compilation. But again, I don't think we really need to explain those details here.

ckipp01 (changes requested):
Thanks for sending this in @Dedelweiss! Just a couple nits.

Dedelweiss (commented):
Done

_overviews/scala3-scaladoc/static-site.md
Dedelweiss: Do you think it's okay if I detail it like this? Or should I just keep the sentence as you did.
For example if you have a directory called `docs` and you'd like that to be treated as your site root :

```
.
└── docs/
    └── _docs/
        ├── index.html
        └── getting-started.html
```

Then the configuration would be as follows:

_overviews/scala3-scaladoc/static-site.md
```diff
@@ -37,6 +37,13 @@ getting-started.html
 
 Scaladoc can transform both files and directories (to organize your documentation into a tree-like structure). By default, directories have a title based on the file name and have empty content. It is possible to provide index pages for each section by creating `index.html` or `index.md` (not both) in the dedicated directory.
 
+Before generating your static site you need to set the `-siteroot` value in your doc `scalacOptions`. The value of this is the directory that holds your docs.  The root URL for the generated documentation will also be `<site-root>`.
+The configuration is as follows:
```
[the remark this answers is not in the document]
Dedelweiss: Do you think it's okay if I detail the example like this or keep the sentence as you did?
For example if you have a directory called `docs` and you'd like that to be treated as your site root :

```
.
└── docs/
    └── _docs/
        ├── index.html
        └── getting-started.html
```

Then the configuration would be as follows:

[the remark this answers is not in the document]
ckipp01: Actually sure, let's be super explicit and just add this in here as well. It may seem a bit duplicated, but at least then it will make it super clear that `docs` here is what `<site-root>` is up above.

_overviews/scala3-scaladoc/static-site.md
```diff
@@ -38,7 +38,18 @@ getting-started.html
 Scaladoc can transform both files and directories (to organize your documentation into a tree-like structure). By default, directories have a title based on the file name and have empty content. It is possible to provide index pages for each section by creating `index.html` or `index.md` (not both) in the dedicated directory.
 
 Before generating your static site you need to set the `-siteroot` value in your doc `scalacOptions`. The value of this is the directory that holds your docs.  The root URL for the generated documentation will also be `<site-root>`.
-The configuration is as follows:
+
+For example if you have a directory called `docs` and you'd like that to be treated as your site root :
```
ckipp01, line 42:
```suggestion
For example if you have a directory called `docs` and you'd like that to be treated as your site root:
```

ckipp01 (approved):
Perfect. Thanks @Dedelweiss! Can you rebase on main? That should get CI to pass. Then LGTM

Dedelweiss:
I think it's good 👍

PR #2710, opened 2023-02-27
fix(ci): update htmlproofer flags since it was updated

ckipp01:


fsalvi (approved):
You were faster than me. Thanks Chris!

PR #2713, opened 2023-02-28
Update lower-type-bounds.md

phrmoy:
Xlint is not a valid compiler option in Scala3.


SethTisue:
I'd suggest something like "on Scala 2, use this option... the lint may be added to Scala 3 in the future", that kind of thing?

PR #2714, opened 2023-03-01
Fix: Prevent the code text from sticking out of the div in Settings page documentation

Dedelweiss:
On the settings page, the code for the -social-links flag is sticking out.
On this PR, I just added a overflow scroll to the p elements with code.
Before:
Screenshot 2023-03-01 at 16 11 04

After:
Screenshot 2023-03-02 at 16 24 15

Reviewed-by: @ckipp01

### DIFF
--- a/_sass/layout/documentation.scss
+++ b/_sass/layout/documentation.scss
@@ -46,6 +46,7 @@
 
   p {
     line-height: 1.5;
+    overflow: scroll;
   }
 
   ol {
--- a/_sass/layout/type-md.scss
+++ b/_sass/layout/type-md.scss
@@ -172,6 +172,7 @@
             color: #859900;
             @include border-radius($border-radius-medium);
             padding: 2px 6px;
+            word-wrap: break-word;
         }
     }
 
--- a/_sass/layout/type-md.scss
+++ b/_sass/layout/type-md.scss
@@ -172,7 +172,6 @@
             color: #859900;
             @include border-radius($border-radius-medium);
             padding: 2px 6px;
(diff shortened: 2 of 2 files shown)


ckipp01 (commented):
Thanks for the pr @Dedelweiss! Since this is code, do you think instead of breaking we may just want instead have an overflow that you can scroll left and right with? I typically don't like that, but in this case it might make the most sense since you don't want to potentially break on code in a place that's showing actual code, causing confusion. Wdyt?

ckipp01:
I think my personal preference is your first one with the overflow on the single line. It _may_ not be super obvious, but at the same time it doesn't break the screen and provides the ability to work.

Dedelweiss:
Desc updated

ckipp01 (approved):
Great, thanks @Dedelweiss!

julienrf:
On Chrome the result does not look good:

Screenshot from 2023-03-13 15-12-09

julienrf:
It seems `overflow: auto` is a better fit because the scrollbars are shown only if the content does not fit:

Screenshot from 2023-03-13 15-17-36

julienrf:
Actually on Firefox I see no scrollbars :thinking: 

Screenshot from 2023-03-13 15-19-58

julienrf:
I’ve reverted the PR for now. I am not sure what should be the correct solution.

Dedelweiss:
Okay sorry, I will see for a better solution.

julienrf:
No problem, thank you for your help! Getting something that works consistently across browsers is not that easy…

PR #2716, opened 2023-03-06
Fix: Increase the number of search results

### LINKED ISSUE #2715: Search functionality seems to be missing things
I noticed today when searching for specific things on the website the search wasn't returning what I was expecting. To give an example let's say I'm interested in learning about the new Scaladoc features in Scala 3. [This section](https://docs.scala-lang.org/scala3/guides/scaladoc/index.html) to be specific. However, when the user searches `scaladoc` the items returned don't include this `scaladoc` portion of the website:

Screenshot 2023-03-02 at 09 14 38

I would expect this to be returned in this query.

Dedelweiss:
As the Scala documentation grows (the new Scala3 version, earlier versions, etc.) the search areas become more diverse, and a search should be as comprehensive as possible so that even the most novice user can easily find what they are looking for. 
This is why in this PR, I have increased the number of search results.

Fixes #2715

### DIFF
--- a/resources/js/functions.js
+++ b/resources/js/functions.js
@@ -338,7 +338,7 @@ $(document).ready(function() {
       resultsContainer: document.getElementById('result-container'),
       json: '/resources/json/search.json',
       searchResultTemplate: '<li><a href="{url}">{title}</a></li>',
-      limit: 5,
+      limit: 7,
     });
 
     $("#blog-search-bar").on("change paste keyup", function() {


ckipp01 (commented):
Ah good find! I didn't even realize we were hitting up on the limit. By changing this does it mean that no matter what 7 items will always be shown, or only when there are more that 5 matching?

Dedelweiss:
I think it is only when more than 5 elements are match.
Screenshot 2023-03-06 at 11 03 39

ckipp01 (approved):
Ok I'm cool with this, but let's get a couple more opinions.

julienrf (approved):
I also agree with this change. It would be good to also have a way to see the complete list of results. Is that possible?

Dedelweiss:
So maybe I'm wrong but I tested locally and nothing changes when I change this value. Is it the fact that it is local, or do I have to use the algolia API key. I don't know but I'll look into it.

SethTisue:
Let's optimistically merge and see what happens.

Dedelweiss:
So, I have just made some corrections and research. Is it possible to reopen this pull request or should I create a new one?

SethTisue:
PRs cannot be reopened. Followup PRs are in this area are welcome.

PR #2719, opened 2023-03-07
Fix: Increase the number of results of the doc search bar

Dedelweiss:
This PR corrects PR #2716 that I made.
In this one I set the search bar value of the blog to 5 as it was before and I set the search bar value of the documentation to 6 with distinct to true.
See the algolia documentation for:
- [hitsPerPage](https://www.algolia.com/doc/api-reference/api-parameters/hitsPerPage/)
- [distinct](https://www.algolia.com/doc/api-reference/api-parameters/distinct/?client=javascript)
### Before:
Screenshot 2023-03-07 at 10 02 42

### After:
Screenshot 2023-03-14 at 16 01 03

Also for the idea of having a paging system to be able to see other results. I've just seen that there is an open issue for this for DocSearch.
https://github.com/algolia/docsearch/issues/883

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -106,7 +106,7 @@ if ($("#doc-search-bar").length) {
     apiKey: 'fbc439670f5d4e3730cdcb715c359391',
     indexName: 'scala-lang',
     inputSelector: '#doc-search-bar',
-    algoliaOptions: { 'facetFilters': ["language:en"] },
+    algoliaOptions: { 'facetFilters': ["language:en"], 'hitsPerPage': 7 },
     debug: false // Set debug to true if you want to inspect the dropdown
   });
 }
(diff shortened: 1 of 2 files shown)


ckipp01:
In your after image it shows two exact same `Scaladoc` results. Are those linking to the same page? Also is that the same page that is listed under Style Guide? Since we'd want to avoid multiple duplicates in the same result.

ckipp01 (approved):
Thanks for digging into this @Dedelweiss. This looks good to me.

LGTM

SethTisue:
Thank you Lucas and Chris!

Dedelweiss:
> Thank you Lucas and Chris!
Thanks to you. I'll keep in touch with the pagination issue created for [DocSearch](https://github.com/algolia/docsearch/issues/883) as it would indeed be interesting to implement it.

Issue #2680, opened 2023-01-29
two or three paragraphs seem missed in _overviews/scala3-book/taste-methods.md


benluo:
After Line 131 in _overviews/scala3-book/taste-methods.md, there should be at least two paragraphs more.

bishabosha:
Please could you describe more precisely what you mean? It doesn't appear to me that there is missing/deleted text. These two paragraphs are talking about the extension methods example snippet that immediately follows.

Do you mean that there is not enough text, and there should be more text dedicated to explaining extension methods?

benluo:
In Chinese version, file  "_zh-cn/overviews/scala3-book/taste-methods.md" from line 154

> 如果没有IDE的帮助，那段代码可能很难阅读，但这个代码要明显得多：
> 
> {% tabs method_10 %}
> {% tab 'Scala 2 and 3' for=method_10 %}
> 
> ```scala
> engage(
>   speedIsSet = true,
>   directionIsSet = true,
>   picardSaidMakeItSo = true,
>   turnedOffParkingBrake = false
> )
> ```
> {% endtab %}
> {% endtabs %}
> 
> ## 扩展方法
> 
> _扩展方法_ 允许您向封闭类添加新方法。
> 例如，如果要将两个名为 `hello` 和 `aloha` 的方法添加到 `String` 类中，请将它们声明为扩展方法：
> 
> {% tabs extension0 %}
> {% tab 'Scala 3 Only' %}
> 
> ```scala
> extension (s: String)
>   def hello: String = s"Hello, ${s.capitalize}!"
>   def aloha: String = s"Aloha, ${s.capitalize}!"
> 
> "world".hello    // "Hello, World!"
> "friend".aloha   // "Aloha, Friend!"
> ```
> 
> {% endtab %}
> {% endtabs %}
> 

But all these lines missed in English version.

bishabosha:
ah this makes more sense, thanks!

gkepka:
[An example](https://docs.scala-lang.org/scala3/book/taste-methods.html#:~:text=extension) of using extension methods has been added in the meantime.

PR #2720, opened 2023-03-07
Fix: Added the missing paragraphs in taste-methods.md english version

Dedelweiss:
In this PR I have added the paragraphs missing in the English version but not in the Chinese version.

Fixes: #2680


ckipp01 (commented):
Thanks for the PR @Dedelweiss although I don't think this is really what we want. Looking at the issue they do seem to be out of sync, but what you have here is already covered on this page as far as I can tell. So I'd say let's close this pr for now and we can discuss on the issue if necessary about if we need to sync the pages or not.

PR #2727, opened 2023-03-09
Update some compiler options

som-snytt:
Updated for a few major features.

It would be nice to automate or at least automatically verify.

### DIFF
--- a/_overviews/scala3-migration/options-intro.md
+++ b/_overviews/scala3-migration/options-intro.md
@@ -18,4 +18,4 @@ To do so you can refer to the [Lookup Table](options-lookup.html).
 
 You can also discover the new Scala 3 compiler options, that have no equivalent in Scala 2.13, in the [New Compiler Options](options-new.html) page.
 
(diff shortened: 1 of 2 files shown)


julienrf (approved):
Thank you very much @som-snytt! The changes look good to me, but I would like someone from the Scala 3 compiler team to also have a look at them.

prolativ (approved):
If we want to keep things ordered alphabetically, then  `-Vdebug`, `-Vlog:<phases>` and `-Wself-implicit` should be moved. Otherwise LGTM

som-snytt:
Improved alphabetic order, added new options (they keep adding options), removed old aliases as a matter of policy.

Checked order against `scalac -help`, `scalac -V`, etc.

Tragically, I moved `-V` and `-W` to the front and lost the useful diff. Sorry about that.

Worth motivating that order, aside from the alphabet. As a user, I want more info (`-V`) or maybe more safety (`-W`). If I'm an advanced user, I might use an advanced option (`-X`). In the privacy of my own home or cubicle, I might add a private option (`-Y`) if no one is looking over my shoulder.

julienrf:
Thank you!

PR #2732, opened 2023-03-10
bugfix

rebekah:
The examples didn't make sense to me... so I played with them a bit realized they were broken and talked with @BalmungSan on discord. He suggested this resolution, which seems concise and appropriate. Here is the code in working(and tested) examples: https://github.com/rebekah/scala-3-play/blob/main/src/main/scala/ExploreContextBounds.scala

### DIFF
--- a/_overviews/scala3-book/ca-context-bounds.md
+++ b/_overviews/scala3-book/ca-context-bounds.md
@@ -8,28 +8,91 @@ previous-page: ca-given-using-clauses
 next-page: ca-given-imports
 ---
 
-In many situations the name of a [context parameter]({% link _overviews/scala3-book/ca-given-using-clauses.md %}#using-clauses) doesn’t have to be mentioned explicitly, since it’s only used by the compiler in synthesized arguments for other context parameters.
+In many situations the name of a [context parameter]({% link _overviews/scala3-book/ca-given-using-clauses.md %}#using-clauses) does not have to be mentioned explicitly, since it is only used by the compiler in synthesized arguments for other context parameters.
 In that case you don’t have to define a parameter name, and can just provide the parameter type.
 
 
 ## Background
 
-For example, this `maximum` method takes a _context parameter_ of type `Ord`, only to pass it on as an argument to `max`:
+For example, consider a method `maxElement` that returns the maximum value in a collection:
 
 {% tabs context-bounds-max-named-param class=tabs-scala-version %}
 
 {% tab 'Scala 2' %}
 ```scala
-def maximum[A](xs: List[A])(implicit ord: Ord[A]): A =
-  xs.reduceLeft(max(_, _)(ord))
+def maxElement[A](as: List[A])(implicit ord: Ord[A]): A =
+  as.reduceLeft(max(_, _)(ord))
 ```
 {% endtab %}
 
 {% tab 'Scala 3' %}
 ```scala
-def maximum[A](xs: List[A])(using ord: Ord[A]): A =
-  xs.reduceLeft(max(_, _)(using ord))
+def maxElement[A](as: List[A])(using ord: Ord[A]): A =
+  as.reduceLeft(max(_, _)(using ord))
 ```
+{% endtab %}
+
+{% endtabs %}
+
+The method `maxElement` takes a _context parameter_ of type `Ord[A]` only to pass it on as an argument to the method
+`max`.
+
+For the sake of completeness, here are the definitions of `max` and `Ord` (note that in practice we would use the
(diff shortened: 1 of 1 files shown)


julienrf:
I think it would be better to provide the definition of `max`. I believe the current code does work if `max` is defined as follows:

~~~ scala
def max[A](using Ord[A]): (A, A) => A =
  (a1, a2) => if a1 > a2 then a1 else a2
~~~

BalmungSan:
@julienrf yeah totally, when I looked at the page I was really lost since I didn't find `max` and the code didn't make sense to me at all.
I assume it was defined before but I didn't want to check multiple pages to find out.

rebekah:
> ```scala
> def max[A](using Ord[A]): (A, A) => A =
>   (a1, a2) => if a1 > a2 then a1 else a2
> ```

@julienrf This max method doesn't appear to use the Ord instance in any way. Am I missing something? Furthermore, when I try to use this code I get the error: ``` value > is not a member of A, but could be made available as an extension method```

rebekah:
> > ```scala
> > def max[A](using Ord[A]): (A, A) => A =
> >   (a1, a2) => if a1 > a2 then a1 else a2
> > ```
> 
> @julienrf This max method doesn't appear to use the Ord instance in any way. Am I missing something? Furthermore, when I try to use this code I get the error: ` value > is not a member of A, but could be made available as an extension method`

BalmungSan:
@rebekah so the funny thing is that both parts of your question are connected :)
Such `max` method should be indeed using `Ord`, specifically by the use of `>` which should be added as an extension method to any value of type `A` given the presence of the `Ord[A]`

But, it is not doing that because that code is also missing an import like:

```scala
import scala.math.Ordering.Implicits._
```

So, the lesson of the day is that code examples in docs should be complete and compiled, otherwise they may confuse newcomers reading them; which is exactly the opposite of what a doc should do :)

julienrf:
I 100% agree with you @rebekah. That page about context bounds explains the syntax that is handled by the compiler, and in some sense we don’t _need_ to know what the definition of `max` or `Ord` is to understand the syntactic transformation performed by the compiler, but I agree that it would be way better to provide a self-contained, executable, example. I propose the following, then.

Scala 3:

~~~ scala
trait Ord[A]:
  /** @return true if a1 is greater than a2, false otherwise */
  def greaterThan(a1: A, a2: A): Boolean

/** @return the maximum value between a1 and a2 */
def max[A](a1: A, a2: A)(using ord: Ord[A]): A =
  if ord.greaterThan(a1, a2) then a1 else a2

/** @return the maximum element in a collection */
def maxElement[A](as: List[A])(using ord: Ord[A]): A =
  as.reduceLeft(max(_, _)(using ord))
~~~

Note that since the `ord` parameter is a context parameter, we don’t need to pass it explicitly to the `max` method:

~~~ scala
def maxElement[A](as: List[A])(using Ord[A]): A =
  as.reduceLeft(max(_, _))
~~~

Note that since we don’t explicitly use that context parameter, we don’t even need to give it name (that sentence does not apply to Scala 2).

In that case, we can shorten the syntax even further by using a context bound:

~~~ scala
def maxElement[A: Ord](as: List[A]): A =
  as.reduceLeft(max(_, _))
~~~

And, optionally, even further:

~~~ scala
def maxElement[A: Ord](as: List[A]): A =
  as.reduceLeft(max)
~~~

In this last form, we replace the function literal `max(_, _)` with the _eta-expansion_ of the method `max`. (This is not really related to the usage of context bounds).

Scala 2:

~~~ scala
trait Ord[A] {
  def greaterThan(a1: A, a2: A): Boolean
}

def max[A](a1: A, a2: A)(implicit ord: Ord): A =
  if (ord.greaterThan(a1, a2)) a1 else a2

def maxElement[A](as: List[A])(implicit ord: Ord[A]): A =
  as.reduceLeft(max(_, _)(ord))
~~~

Or:

~~~ scala
def maxElement[A: Ord](as: List[A]): A =
  as.reduceLeft(max(_, _))
~~~

Or even:

~~~ scala
def maxElement[A: Ord](as: List[A]): A =
  as.reduceLeft(max[A])
~~~

We should also add a note that in practice there is already a trait `Ordering[A]` in the standard library, which provides a methods `max`, and there is already a method `max` on `List`.

rebekah:
@julienrf Wow, yes that would be fantastic!

rebekah:
thank you

julienrf:
Thank _you_ for starting this discussion!

rebekah:
@julienrf you are welcome!

PR #2742, opened 2023-03-16
add code tab in _zh-cn/overviews/scala3-book/fun-hofs.md

benluo:


jxnu-liguobin (approved):

julienrf:
After a second thought, I think it is fine to keep the code tabs, thank you!

PR #2754, opened 2023-03-21
add code tab in types-union.md

benluo:


_overviews/scala3-book/types-union.md
julienrf, line 1: Since this file describes a feature that has no Scala 2 equivalent, we should follow the same style as https://github.com/scala/docs.scala-lang/blob/main/_overviews/scala3-book/ca-multiversal-equality.md?plain=1, ie we inform the readers at the very beginning that all the code examples are Scala 3 only (and we don’t use code tabs).

Could you please add the following at the beginning at the file?

~~~ markdown
<span class="tag tag-inline">New In Scala 3</span>

> The union type is a new language feature that was introduced in Scala 3.
> Because it has no equivalent in Scala 2, all code examples
> in this lesson assume you are using Scala 3.
~~~

julienrf (changes requested):
Thank you @benluo, however since the file describes a feature that has no equivalent in Scala 2 we follow a different style, see my comment below.

julienrf:
Hey @benluo,
We have decided to streamline the way we “label” pages of documentation that are specific to a Scala version by using metadata in the YAML front matter instead of tabs around the code examples. Please see the PR #2768 for more details. I also went ahead and migrated the file types-union.md in that PR.
Thank you again for all your contributions!

PR #2755, opened 2023-03-21
add code tab in types-adts-gadts.md

benluo:


_overviews/scala3-book/types-adts-gadts.md
julienrf, line 1: I think this file also deserves a comment at the top to explain that it will use a new feature of Scala 3, and that in Scala 2 you have to manually use the “desugared” style (shown at the bottom)

julienrf:
I went ahead and updated this file in #2768 as well, so I am closing this PR. I appreciate your help, I should have clarified earlier the style to use when a page's content is specific to a Scala version.

PR #2760, opened 2023-03-30
Add Scala 2 version of case classes that can evolve

julienrf:
The same approach as with Scala 3 also works with Scala 2 if you add the `-Xsource:3` compiler option.


bishabosha (approved):

sideeffffect:
Just to be clear, with `-Xsource:3` and case class constructor `private`, both `copy` and `apply` are private too?

julienrf:
> Just to be clear, with `-Xsource:3` and case class constructor `private`, both `copy` (in class) and `apply` (in companion) are private too?

Yes.

sideeffffect:
But to be super duper clear, that means that adding `-Xsource:3` is a breaking change (assuming the project in question is using case classes), right?

julienrf:
Indeed.

PR #2765, opened 2023-04-10
Update types-union doc.md

Randle-Lanre:
added indicator that this is scala3 only for those new to Scala


julienrf:
Thank you for opening a PR @Randle-Lanre. However, it overlaps with the ongoing PR #2754. Would you be interested in contributing to that PR instead? Please follow the suggestion I wrote in https://github.com/scala/docs.scala-lang/pull/2754#discussion_r1154407764.

PR #2768, opened 2023-04-12
Add more guidelines for labelling code examples for Scala 2/3

julienrf:
Also, apply the recommended approach to a couple of existing pages.

image

image

Before:

image


bishabosha (changes requested):
You kept the "Scala 2 and 3" tab on `ca-given-using-clauses.md` which is a bit strange

julienrf:
I kept that one because it shows what the code looks like when we don’t use given/using. I also kept a similar code example with “Scala 2 and 3” in the Union Types page, which shows an alternative to union types.

But I agree that in the given/using page I could remove the code tab, though, if you think it is confusing.

bishabosha (approved):

PR #2777, opened 2023-04-26
Add example for error handling in ExecutionContext

som-snytt:
It was helpful to me to work out this example, as a refresher. Therefore, it may be helpful to beginners, in case there are still beginners out there using `Future` for the first time.

The existing words are correct, but might require a few hours of experimentation (or debugging) to comprehend.

### DIFF
--- a/_overviews/core/futures.md
+++ b/_overviews/core/futures.md
@@ -1140,7 +1140,9 @@ See the factory methods [`fromExecutor`](https://www.scala-lang.org/api/current/
 and [`fromExecutorService`](https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext$.html#fromExecutorService(e:java.util.concurrent.ExecutorService,reporter:Throwable=%3EUnit):scala.concurrent.ExecutionContextExecutorService).
 
 Since it is necessary to set the [`UncaughtExceptionHandler`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Thread.UncaughtExceptionHandler.html)
-for executing threads, as a convenience, `fromExecutor` will create a correctly configured context when passed a `null` executor.
+for executing threads, as a convenience, when passed a `null` executor,
+`fromExecutor` will create a context that is configured the same as `global`,
+but with the supplied reporter for handling exceptions.
 
 The following example demonstrates how to obtain an `ExecutionContext` with custom error handling
 and also shows the result of different exceptions, as described above:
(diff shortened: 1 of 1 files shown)


som-snytt:
The failure is unrelated
```
For the Links > Internal check, the following failures were found:

* At ./_site/sips/quote-pattern-type-variable-syntax.html:604:

  internally linking to Explicit-type-variables, which does not exist

```
Tried rebasing.

som-snytt:
The example should be specialized for Scala 3.

julienrf:
Hey, indeed the failure was unrelated to your PR, I fixed it on `main`.

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+      def failing(x: Any): Int = throw new NumberFormatException("test")
+      def interrupt(x: Any): Int = throw new InterruptedException("test")
+      def erroring(x: Any): Int = throw new AssertionError("test")
+      // warning: return statement uses an exception to pass control to the caller of the enclosing named method nonlocally
+      def nonlocally(xs: List[Int]): Int = { xs.foreach(x => if (x > 0) return x); -1 }
+      def testCrashes()(implicit ec: ExecutionContext): Future[Int] = Future.unit.map(crashing)
+      def testFails()(implicit ec: ExecutionContext): Future[Int] = Future.unit.map(failing)
+      def testInterrupted()(implicit ec: ExecutionContext): Future[Int] = Future.unit.map(interrupt)
+      def testError()(implicit ec: ExecutionContext): Future[Int] = Future.unit.map(erroring)
+      def testNonLocal()(implicit ec: ExecutionContext): Future[Int] = Future(List(42)).map(nonlocally)
```
julienrf, line 1166: I think the code would be clearer if the methods `crashing`, `failing`, etc. took no parameters and in the tests you would do something like `Future.unit.map(_ => crashing())`.

I was also going to say that it would even clearer with just `Future { crashing() }` but I do think it is important to highlight that exceptions thrown in the function passed to `map` also receive the special treatment. Maybe it would be good to clarify that explicitly, actually.

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+and also shows the result of different exceptions, as described above:
+
+{% tabs exceptions %}
+{% tab 'Scala 2 and 3' for=exceptions %}
+    import scala.concurrent._, duration._
+    import scala.util.chaining._
+    import java.util.concurrent.ForkJoinPool
+    import java.util.concurrent.TimeUnit.SECONDS
+
+    class C {
```
julienrf, line 1155: Is it really useful to define a class `C`? Why not put those methods directly in the object `Test` below?

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+and [`fromExecutorService`](https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext$.html#fromExecutorService(e:java.util.concurrent.ExecutorService,reporter:Throwable=%3EUnit):scala.concurrent.ExecutionContextExecutorService).
+
+Since it is necessary to set the [`UncaughtExceptionHandler`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Thread.UncaughtExceptionHandler.html)
+for executing threads, as a convenience, `fromExecutor` will create a correctly configured context when passed a `null` executor.
+
+The following example demonstrates how to obtain an `ExecutionContext` with custom error handling
+and also shows the result of different exceptions, as described above:
+
+{% tabs exceptions %}
+{% tab 'Scala 2 and 3' for=exceptions %}
```
julienrf, line 1149: The code is not idiomatic for Scala 3, it would be great to use distinct tabs for Scala 2 and Scala 3 here once we settle on the content.

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+
+    object Test extends App {
+      val c = new C
+      def check(f: => Future[Int]): Future[Int] =
+        try Await.ready(f, Duration(1L, SECONDS)).tap(res => println(s"ready $res"))
+        catch { case e: Exception => println(s"threw $e"); Future.never }
+      def reporter(t: Throwable) = println(s"reported $t")
+      locally {
+        import ExecutionContext.Implicits._
+        check(c.testFails())       // a failed Future
```
julienrf, line 1177: I see that in some tests below what you wrote in the comment corresponds to what was written to the standard output. It would be good to do the same here, I think.

```suggestion
        check(c.testFails())       // ready Failure(NumberFormatException(test))
```

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+        check(c.testCrashes())     // threw java.util.concurrent.TimeoutException
+        check(c.testInterrupted()) // a failed Future
+          .failed
+          .foreach(e => println(s"failed ${e.getCause}")) // failed java.lang.InterruptedException: test
```
julienrf, line 1181: I wonder if we could make the code more readable by changing the signature of `check` to the following:

~~~ scala
def check(f: Future[Int]): Unit
~~~

And printing the “cause” of failures directly in the implementation of `check`. I also think we don’t want `f` to be a by-name parameter.

julienrf (changes requested):
Thank you very much @som-snytt for adding this clarification! I have left a few comments.

_overviews/core/futures.md
```diff
@@ -1121,11 +1122,88 @@ behind this is to prevent propagation of critical and control-flow related
...
+That is, "fatal" means that the error is not recoverable for the `ExecutionContext`
+and is also not intended to be handled by user code. By contrast, application code may
+attempt recovery from a "failed" `Future`, which has completed but with an exception.
+
+An execution context can be customized with a reporter that handles fatal exceptions.
+See the factory methods [`fromExecutor`](https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext$.html#fromExecutor(e:java.util.concurrent.Executor,reporter:Throwable=%3EUnit):scala.concurrent.ExecutionContextExecutor)
+and [`fromExecutorService`](https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext$.html#fromExecutorService(e:java.util.concurrent.ExecutorService,reporter:Throwable=%3EUnit):scala.concurrent.ExecutionContextExecutorService).
+
+Since it is necessary to set the [`UncaughtExceptionHandler`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Thread.UncaughtExceptionHandler.html)
+for executing threads, as a convenience, `fromExecutor` will create a correctly configured context when passed a `null` executor.
```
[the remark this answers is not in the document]
som-snytt: Indeed, it might be "incorrectly configured"!

_overviews/core/futures.md
som-snytt: I ran out of steam. Because I removed the non-local example (which didn't make sense), the only difference is the import style, and maybe braces for locally. Thanks for the review, I tried to take all the suggestions, as usual, any remaining errata are my own.

julienrf: Thank you for all the changes so far. I would like to still make more suggestions by directly pushing commits to your branch. I should get some time for that next week.

som-snytt: Thanks for making it useful if that's possible.

julienrf:
Hey @som-snytt, I took the liberty to make some changes to your initial PR, I hope this is fine with you. What do you think of the result?

Here is the rendering of the Scala 3 version:

image

som-snytt (commented):
Thanks. I see indentation doesn't matter after all. (I wasn't sure.) I don't have approval privilege on this repo, but I approve this message.

julienrf:
Thank you for the review! I’ve updated the imports according to your feedback. What do you think of the result?

julienrf (approved):

PR #2798, opened 2023-05-11
use dotty website code highlighting

bishabosha:
I made a few changes to what was on the dotty website, but in general improvements are:
- quotes and splices
- various numeric literals (hex, floating + exponent, `_` digit separators)
- enums
- triple quote string literals (specifically that mix other quotes inside)
- escaped characters

critically json in string literals renders correctly for the toolkit tutorials:
Screenshot 2023-05-11 at 10 32 32

maybe at some point we can try to update the official highlight-js for scala with these changes

### DIFF
--- a/resources/js/hljs-scala3.js
+++ b/resources/js/hljs-scala3.js
@@ -271,7 +271,7 @@ function highlightDotty(hljs) {
   // Methods
   const METHOD = {
     className: 'function',
-    begin: `((${modifiers}|transparent|inline|infix) +)*def`, end: / =\s|\n/,
+    begin: `((${modifiers}|transparent|inline|infix) +)*def `, end: / =\s|\n/,
     excludeEnd: true,
     relevance: 5,
     keywords: withSoftKeywords('inline infix transparent'),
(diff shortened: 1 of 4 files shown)


resources/js/hljs-scala3.js
julienrf, line 1: The end result looks great, but is it possible to publish that package on npm and reference it from there rather than copying it here?

julienrf: Who is it maintained by? Are there any tests?

bishabosha: Do you mean our own custom package? or the official hljs CDN?

But maybe its better to incubate here (temporarily) so we can improve any issues?

julienrf (approved):
This fixes a lot of issues, but please make sure the syntax highlighting implementation is eventually contributed upstream to highlight.js so that we don’t maintain multiple copies of it.

resources/js/hljs-scala3.js
bishabosha: to be clear, it comes originally from https://github.com/lampepfl/dotty/blob/main/scaladoc/resources/dotty_res/scripts/hljs-scala3.js

PR #2804, opened 2023-05-17
Update the binary compatibility guarantees for Scala 3.

sjrd:
In addition, be a more assertive that we strongly *intend* compatibility of produced artifacts to be preserved, while still making it clear that we cannot *guarantee* it as it is not mechanically checkable.


SethTisue (approved):

_overviews/core/binary-compatibility-of-scala-releases.md
```diff
@@ -7,31 +7,79 @@ partof: binary-compatibility
...
 
 We check binary compatibility automatically with [MiMa](https://github.com/lightbend/migration-manager). We strive to maintain a similar invariant for the `behavior` (as opposed to just linkage) of the standard library, but this is not checked mechanically (Scala is not a proof assistant so this is out of reach for its type system).
 
+Note that for Scala.js and Scala Native, binary compatibility issues result in *linking errors* as opposed to run-time exceptions.
```
julienrf, line 14: It is not super clear what those linking errors are, especially just after we mentioned the type `LinkageError`. Do you think it would be possible to change the wording to say something like it “results in errors at build time”?

_overviews/core/binary-compatibility-of-scala-releases.md
```diff
@@ -7,31 +7,79 @@ partof: binary-compatibility
...
+These are strict constraints, but they have worked well for us since Scala 2.10.x.
+They didn't stop us from fixing large numbers of issues in minor releases.
+So far, that policy is still applicable for Scala 2, although [there exists a proposal to remove the forward compatibility guarantee](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html).
+
+For Scala 3, the minor version is the *second* number in a version, e.g., 2 in v3.2.1.
+The third number is the *patch* version.
+The major version is always 3.
+
+Scala 3 guarantees both backward and forward compatibility across *patch* releases within a single minor release.
+In particular, this applies within an entire LTS series (which will start with Scala 3.3.x).
```
julienrf, line 36: I think we should describe (or link to) what LTS means.

_overviews/core/binary-compatibility-of-scala-releases.md
```diff
@@ -7,31 +7,79 @@ partof: binary-compatibility
...
+The *policies* above extend to libraries compiled by particular Scala compiler versions.
+Every effort is made to preserve the binary compatibility of artifacts produced by the compiler.
+*However*, that cannot be mechanically checked.
+It is therefore possible, due to bugs or unforeseen consequences, that recompiling a library with a different compiler version affects its binary API.
+We cannot *guarantee* that it will never happen.
+
+We recommend that library authors use [MiMa](https://github.com/lightbend/migration-manager) themselves to verify compatibility of minor versions before releasing.
+
+#### TASTy and Pickle Compatibility
+*Binary* compatibility is a concept relevant at link time of the the target platform (JVM, Scala.js or Scala Native).
```
julienrf, line 59:
```suggestion
*Binary* compatibility is a concept relevant at link time of the target platform (JVM, Scala.js or Scala Native).
```

julienrf (commented):
Thank you, this is a great improvement!

I know the original document is only about binary compatibility, but I think it would be useful to also cover source compatibility. I think we could say that we can not guarantee backward source compatibility across compiler releases, but we have some tools (the open community build) to detect such incompatibilities and mitigate them, especially within an LTS series.

Kordyjan (approved):
Looks good.
I only suggest describing the distinction between JVM's `LinkageError`s and build-time linking errors from other platforms in more detail.

_overviews/core/binary-compatibility-of-scala-releases.md
```diff
@@ -7,31 +7,79 @@ partof: binary-compatibility
...
+#### Checking
+For the Scala library artifacts (`scala-library`, `scala-reflect` and `scala3-library`), these guarantees are mechanically checked with [MiMa](https://github.com/lightbend/migration-manager).
+
+The *policies* above extend to libraries compiled by particular Scala compiler versions.
+Every effort is made to preserve the binary compatibility of artifacts produced by the compiler.
+*However*, that cannot be mechanically checked.
+It is therefore possible, due to bugs or unforeseen consequences, that recompiling a library with a different compiler version affects its binary API.
+We cannot *guarantee* that it will never happen.
+
+We recommend that library authors use [MiMa](https://github.com/lightbend/migration-manager) themselves to verify compatibility of minor versions before releasing.
```
xuwei-k, line 56:
```suggestion
For the Scala library artifacts (`scala-library`, `scala-reflect` and `scala3-library`), these guarantees are mechanically checked with [MiMa](https://github.com/lightbend/mima).

The *policies* above extend to libraries compiled by particular Scala compiler versions.
Every effort is made to preserve the binary compatibility of artifacts produced by the compiler.
*However*, that cannot be mechanically checked.
It is therefore possible, due to bugs or unforeseen consequences, that recompiling a library with a different compiler version affects its binary API.
We cannot *guarantee* that it will never happen.

We recommend that library authors use [MiMa](https://github.com/lightbend/mima) themselves to verify compatibility of minor versions before releasing.
```

https://github.com/lightbend/mima/commit/906a81775d0972200291b9b5eff73a4bd86b8792

sjrd:
Updated to address the reviews.

julienrf (approved):
The content looks good to me. I still think we should clarify the source compatibility guarantees, but that can be achieved in a separate PR.

PR #2806, opened 2023-05-19
Update link to scala 3 contributing guide

bishabosha:
this has the right direct


bishabosha:
closed because it should really go to https://dotty.epfl.ch/docs/contributing/index.html, not https://dotty.epfl.ch/docs/contributing/getting-started.html

PR #2818, opened 2023-05-26
Early init migration for another common use case

lrytz:


_overviews/scala3-migration/incompat-dropped-features.md
```diff
@@ -226,6 +226,45 @@ class Fizz private (val name: String) extends Bar {
...
+~~~ scala
+class Adder {
+  var sum = 0
+  def add(x: Int): Unit = sum += x
+  add(1)
+}
+class LogAdder extends Adder {
+  private var added: Set[Int] = Set.empty
+  override def add(x: Int): Unit = { added += x; super.add(x) }
+}
```
sjrd, line 242: Did you mean to actually show the early init here?
```suggestion
class LogAdder extends {
  private var added: Set[Int] = Set.empty
} with Adder {
  override def add(x: Int): Unit = { added += x; super.add(x) }
}
```

lrytz: Ah, yeah :-)

sjrd (approved):

PR #2819, opened 2023-05-26
Refactor: Add documentation for redirectTo/redirectFrom

Dedelweiss:
I think it would be interesting to include documentation about the redirectTo/redirectFrom properties. 
I've added documentation that's a bit more detailed than the properties before because I thought it was important to add a context for use. 
These properties are quite important in a project to avoid breaking the various links.

I wasn't sure what **This setting is not exported to the template engine.** meant so I didn't put it in doubt but I'd be interested to know.
Added some markdown correction too.

### DIFF
--- a/_overviews/scala3-scaladoc/static-site.md
+++ b/_overviews/scala3-scaladoc/static-site.md
@@ -97,6 +97,35 @@ Predefined properties:
 - **hasFrame** when set to `false` page will not include default layout (navigation, breadcrumbs, etc.) but only token HTML wrapper to provide metadata and resources (js and css files). **This setting is not exported to the template engine.**
 - **layout** - predefined layout to use, see below. **This setting is not exported to the template engine.**
 
+Redirection properties:
+
+In addition to the predefined properties, Scaladoc also supports redirection properties, which allow you to redirect from one page to another. This can be useful when you move a page to a new location but want to keep the old URL working.
+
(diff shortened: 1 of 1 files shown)


_overviews/scala3-scaladoc/static-site.md
```diff
@@ -97,6 +97,19 @@ Predefined properties:
...
+
+- **redirectFrom** - redirect from another page.
+
+You can use them when you want to change the location of the page without breaking the links. Example:
+
+```
+---
+redirectTo: /docs/scala3-scaladoc/static-site.html
+---
+```
```
julienrf, line 112: `redirectFrom` is the most used one, so I think we should mention it before `redirectTo`.

`redirectTo` is used in an otherwise empty .md file, when we want to redirect to an external website (thus, it is impossible in such a case to use `redirectFrom`).

It would be good to explain a bit more how to use `redirectFrom`. Typically, when you move a page to a new location but want the old URL to still work, you add `redirectFrom: /absolute/path/to/old/url.html` in the frontmatter and Scaladoc generates an empty page at `/absolute/path/to/old/url.html` with a browser-based redirection to the new location. It would be good to also mention the Jekyll plugin [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) that inspired the feature.

julienrf (commented):
Thank you for the clarification. I think the content is at the right place in the docs, but I’ve suggested some changes, see below.

_overviews/scala3-scaladoc/static-site.md
Dedelweiss: Thank you @julienrf, it it better ?

julienrf (approved):
Great, thank you for the update!

PR #2824, opened 2023-05-27
add code tabs in TWO ca-given-imports.md

benluo:


sjrd (changes requested):

julienrf (changes requested):
Hey @benluo, thank you for the PR but since given imports and multiversal equality are features specific to Scala 3, we decided to label the whole page as “Scala 3 specific” instead of each code example (hence the absence of tabs on code snippets). You can see the guidelines [here](https://docs.scala-lang.org/contribute/add-guides.html#scala-2-vs-scala-3).

Could you please remove the tabs on the pages `ca-given-imports` and `ca-multiversal-equality`? Also in the future it would be easier to open separate PRs for changes in the documentation and changes in translations. Thank you!

PR #2842, opened 2023-06-12
Follow up on page renames

eed3si9n:
**Problem**
Build is failing on CI, unable to resolve some of the dead links.

**Solution**
This redirects some of the Dotty pages to the new pages.


julienrf:
Thank you very much Eugene but I started #2833 already. I am sorry that you didn’t see it.

PR #2843, opened 2023-06-12
remove `_` in BigDecimal() in first-look-at-types.md

jinfei1125:
The `_` literal will result in following error:
 <console>:6: error: Invalid literal number
       val a = BigDecimal(123_456.789)

### DIFF
--- a/_overviews/scala3-book/first-look-at-types.md
+++ b/_overviews/scala3-book/first-look-at-types.md
@@ -176,7 +176,7 @@ When you need really large numbers, use the `BigInt` and `BigDecimal` types:
 {% tab 'Scala 2 and 3' for=type-bigint %}
 ```scala
 val a = BigInt(1_234_567_890_987_654_321L)
-val b = BigDecimal(123_456.789)
+val b = BigDecimal(123456.789)
 ```
 {% endtab %}
 {% endtabs %}


Philippus:
I think underscores in numeric literals these were introduced in Scala 2.13. Perhaps you are using Scala 2.12?

jinfei1125:
> I think underscores in numeric literals these were introduced in Scala 2.13. Perhaps you are using Scala 2.12?

@Philippus Yep you're right! I'm in scala 2.12 now. Is `_` commonly used? I can't easily find the official documentation about it by googling...

jinfei1125:
Ah, it's a separator for readability purposes. Although, for the spirit of simplicity for official documentation and support multiple versions of scala, I suggest we remove `_`. But if you think it's okay to leave a version-specific example there, I can close this pull request. @Philippus

SethTisue:
If this were just a 2.12 vs 2.13 thing, I'd be open to removing a trivial incompatibility. But the page you have edited is in the Scala 3 edition of the book.

SethTisue:
Hmm... I reopened thinking the problem was fixable by having separate Scala 2 and 3 versions of the code sample, but then I realized that would give the incorrect impression that the Scala 3 version wouldn't compile on 2.13.

I think it's better we leave this alone. If you want to learn Scala 2.12, the best thing to do would be to get a Scala 2.12 book, otherwise you'll be bedeviled by all sorts of small incompatibilities like this.

PR #2851, opened 2023-06-30
Update generate.md

corey-at-protenus:
3 changes: 

1. note that installing via coursier may be needed.
2. correct link to "your first Scala program"
3. Include the current options for search engines and lazy developers

Can split into multiple if requested. That third change might be contentious :)

### DIFF
--- a/_overviews/scaladoc/generate.md
+++ b/_overviews/scaladoc/generate.md
@@ -32,7 +32,7 @@ For additional information about configuring Scaladoc in sbt, see the [Generate 
 
 ## Using scaladoc command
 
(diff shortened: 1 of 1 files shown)


julienrf:
Thank you! Indeed I am not sure about the third point, which might be hard to keep up to date. Also, there are specific pages for the scaladoc tool in Scala 3, which has a page listing the available options: https://docs.scala-lang.org/scala3/guides/scaladoc/settings.html.

I suggest removing the current options of scaladoc from this PR, and opening separate PRs to:
- cross-reference the `scaladoc` documentation for Scala 2 from the `scaladoc` documentation for Scala 3 (and _vice versa_)
- update the list of options of the Scala 3 version

corey-at-protenus:
@julienrf thanks for your patience. I've updated this PR according to your feedback :)

PR #2852, opened 2023-07-01
Add anchors for headings in documentation

kiendang:
to easily get url to a specific section in the doc

examples from https://docs.scala-lang.org/tour/singleton-objects.html

1.

h2

2.

h1


julienrf (approved):
Thank you for this contribution, I think it will be very useful!

PR #2860, opened 2023-07-09
Ambiguous link guidance

j-mie6:
When working with Scaladoc, it can be annoying to work out the correct way of dealing with ambiguous links.

This PR adds a section to the "Scaladoc for Library Authors" page to address this, clearing up exactly how this should be done. There appear to be a couple of bugs/holes in the disambiguator, so I've added them there as cautionary tales too.

(Compiler should now point at this page instead of its own quick guide, implemented in https://github.com/scala/scala/pull/10458, pending this merging).

### DIFF
--- a/_overviews/scaladoc/for-library-authors.md
+++ b/_overviews/scaladoc/for-library-authors.md
@@ -239,6 +239,19 @@ The markup for list blocks looks like:
 - DRY - don't repeat yourself. Resist duplicating the method description in the
   `@return` tag and other forms of repetitive commenting.
 
+## Resolving Ambiguous Links within Scaladoc Comments
+<!-- FIXME: this is wrong in a few ways -->
+Disambiguating terms and types: Suffix terms with '$' and types with '!' in case both names are in use:
+ - `[[scala.collection.immutable.List!.apply class List's apply method]]` and
+ - `[[scala.collection.immutable.List$.apply object List's apply method]]`
+Disambiguating overloaded members: If a term is overloaded, you can indicate the first part of its signature followed by *:
+ - `[[[scala.collection.immutable.List$.fill[A](Int)(=> A):List[A]* Fill with a single parameter]]]`
+ - `[[[scala.collection.immutable.List$.fill[A](Int, Int)(=> A):List[List[A]]* Fill with a two parameters]]]`
+Notes:
+ - you can use any number of matching square brackets to avoid interference with the signature
+ - you can use `\\.` to escape dots in prefixes (don't forget to use * at the end to match the signature!)
+ - you can use `\\#` to escape hashes, otherwise they will be considered as delimiters, like dots.
+
 ## More details on writing Scaladoc
 
 Further information on the formatting and style recommendations can be found in
--- a/_overviews/scaladoc/for-library-authors.md
+++ b/_overviews/scaladoc/for-library-authors.md
@@ -251,7 +251,7 @@ class bar {
 }
 ```
 
-If one references `foo` via `[[foo]]`, then the Scaladoc will complain and offers up both
+If one references `foo` via `[[foo]]`, then the Scaladoc will complain and offer both
 alternatives. Fixing this means elaborating the signature _enough_ so that it becomes unambiguous.
 There are a few things to be aware of in general:
 
@@ -260,8 +260,9 @@ There are a few things to be aware of in general:
 * You must fully qualify any types you are using: assume that you have written your program without
   any import statements!
 
-Then, to disambuate between objects and types (where both are in use), postfix with a `$` for objects
-and `!` for types. For example:
+Then, to disambiguate between objects and types, append `$` to designate a term name
+and `!` for a type name. Term names include members which are not types, such as `val`, `def`, and
+`object` definitions. For example:
  - `[[scala.collection.immutable.List!.apply class List's apply method]]` and
  - `[[scala.collection.immutable.List$.apply object List's apply method]]`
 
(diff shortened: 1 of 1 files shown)


_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+
+```scala
+import scala.collection.mutable.ListBuffer
+class bar {
+    def foo(x: Int): Boolean = ???
+    def foo(x: ListBuffer[Int], y: String): Int = ???
+}
+```
+
+If one references `foo` via `[[foo]]`, then the Scaladoc will complain and offers up both
```
som-snytt, line 254: "offers up" -> "offer"

_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+alternatives. Fixing this means elaborating the signature _enough_ so that it becomes unambiguous.
+There are a few things to be aware of in general:
+
+* You must not use a space in the description of the signature: this will cause Scaladoc to
+  think the link has ended and move onto its description.
+* You must fully qualify any types you are using: assume that you have written your program without
+  any import statements!
+
+Then, to disambuate between objects and types (where both are in use), postfix with a `$` for objects
```
som-snytt, line 263: disambiguate

terms and types. ~(where both are in use)~ is redundant.

"append `$` to designate a term name" etc.

If you think it necessary: "Term names include members which are not types, such as `val`, `def`, and `object` definitions." Pairings such as `class C(val c: Int); def C(c: Int) = new C(c)` would be prone, IIUC.

_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+  that allows you to cut off the signature when it is umambiguous.
+* You must specify the names of the arguments and they must be _exactly_ as written in the
+  function definition:
+  - `[[bar.foo(Int)*]]` is **illegal** (no name)
+  - `[[bar.foo(y:Int)*]]` is **illegal** (wrong name)
+  - `[[bar.foo(x: Int)*]]` is **illegal** (space! Scaladoc sees this as `bar.foo(x:`)
+  - `[[bar.foo(x:Int):Boolean]]` is **illegal** (no `*`)
+  - `[[bar.foo(x:Int)*]]` is **legal** and unambiguous
+  - `[[bar.foo(x:Int*]]` is **legal**, the `Int` is enough to disambiguate so no closing paren needed
+* The enclosing scope (package/class/object etc) of the method must use `.`, but within the arguments
```
som-snytt, line 280: Does this mean: Within the Scaladoc for method and class parameters and for method return types, `\.` must be used to fully qualify a name

_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+  and return type `\.` must be used instead:
+  - `[[bar.foo(x:ListBuffer[Int],y:String)*]]` is **illegal** (no qualification on `ListBuffer`)
+  - `[[bar.foo(x:scala.collection.mutable.ListBuffer[Int],y:String)*]]` is **illegal** (non-escaped dots!)
+  - `[[bar.foo(x:scala\.collection\.mutable\.ListBuffer[Int],y:String)*]]` is **legal**
+  - `[[bar.foo(x:scala\.collection\.mutable\.ListBuffer[Int]*]]` is **legal**, the first argument is
+    enough to disambiguate.
+* When generics are involved, any number of additional square brackets may be used to avoid the
+  signature accidentally closing the link:
```
som-snytt, line 288: I understand from the example.

The number of leading left brackets in a link determine the number of right brackets required to end the link. If a generic signature contains a sequence of right brackets that accidentally [inadvertently] closes the link, just add enough enclosing brackets to avoid it.

I can't believe all this info. No wonder I could never get Scaladoc to work.

_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+  - `[[baz(x:List[List[A]])*]]` is **illegal** (it is read as `baz(x:List[List[A`)
+  - `[[[baz(x:List[List[A]])*]]]` is **legal** (the `]]` is no longer a terminator, `]]]` is)
+
+Current holes:
```
som-snytt, line 292: Known limitations

_overviews/scaladoc/for-library-authors.md
```diff
@@ -239,6 +239,62 @@ The markup for list blocks looks like:
...
+  * it is unclear how `#` is to be handled within argument/return types: `#` doesn't seem to work
+    and neither does `\#`: there are no tests for this, so perhaps this is a bug.
```
som-snytt, line 294: Since this is not an internal document, delete "no tests" comments.

Suggest: `#` syntax is not supported in the Scaladoc for parameters and return types.

Disambiguation of an implicit parameter list is not supported, because it would require writing a space after `implicit`.

som-snytt (approved):
Tweaks and typos. Lots of good information.

If I select "request changes", then I'm on the hook to re-review?

It's already very useful info. The text can be amended later, if desired. Or now, if possible.

_overviews/scaladoc/for-library-authors.md
j-mie6: Yes, Scaladoc definitely requires a bit of love, especially in this aspect!

SethTisue:
@julienrf does the Center have a Scaladoc expert who would like to take a look? (if not, let's just hit merge?)

som-snytt:
Scala Center's charter should mandate that it have a Scaladoc expert on call 24/7. A Scaladoc doc.

@j-mie6 for future reference, if you at-notify someone in a commit message, they receive spam on every rebase. If your typo was intentional, thanks for that!

I will use that clever hack, @SethTissue! Maybe we can elicit further input from @oderski.

j-mie6:
Let's say that the typo _was_ intensional. Didn't actually realise that a tag in a commit actually... Well... Tagged.

som-snytt:
In case I misremembered or github notifications have changed, I will alert you on my next PR and we will learn the truth. Certainly, mentioning a ticket will spam the comments on the issue. Also, if I remember to try that experiment, "Apologies in Advance"!

j-mie6:
Why is there no 😱 react GitHub...

julienrf (approved):
@j-mie6 Thank you!

PR #2863, opened 2023-07-14
mark value classes as scala 2 only

bishabosha:


sjrd (changes requested):
Value classes are still a thing in Scala 3. *Implicit* value classes are typically replaced by `extension` methods, but not other value classes.

bishabosha:
> Value classes are still a thing in Scala 3. _Implicit_ value classes are typically replaced by `extension` methods, but not other value classes.

Well we decided that we would promote users to use opaque types in Scala 3, it doesnt say that its illegal

julienrf (approved):

PR #2866, opened 2023-07-14
fix packaging and imports

bishabosha:
it used old syntax for givens, and one of the code tabs had a duplicate ID


julienrf (approved):

julienrf:
Thank you Jamie!

PR #2867, opened 2023-07-14
port java interaction to scala 2

bishabosha:


_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
+{% tabs foo-definition %}
+{% tab Java %}
 ```java
-// java
-public class JavaClass {
-  public static List<String> getStrings() {
-    return new ArrayList<String>(List.of("a", "b", "c"));
-  }
+public interface Foo {
+  static java.util.List<String> getStrings();
```
julienrf, line 53: I don’t think abstract static methods exist in Java.

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
+  val javaList: java.util.List[String] = Foo.getStrings()
+  val scalaSeq: mutable.Seq[String] = javaList.asScala
   for s <- scalaSeq do println(s)
 ```
+{% endtab %}
+{% endtabs %}
 
-Of course that code can be shortened, but the individual steps are shown here to demonstrate exactly how the conversion process works.
-
+In the above code `scalaSeq` is a wrapper over `javaList` that adapt's it to Scala's `mutable.Seq` collection.
```
julienrf, line 90:
```suggestion
In the above code `scalaSeq` is a wrapper over `java.util.List` that adapts it to Scala's `mutable.Seq` collection.
```

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
+To demonstrate this, here’s a Java API that returns an `Optional[String]` value:
 
-public class JavaClass {
-  static Optional<String> oString = Optional.of("foo");
-  static Optional<String> oEmptyString = Optional.empty();
+{% tabs bar-definition %}
+{% tab Java %}
+```java
+public interface Bar {
+  static java.util.Optional<String> optionalString();
```
julienrf, line 105: Same comment here about abstract static methods.

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
 }
 ```
+{% endtab %}
+{% endtabs %}
 
-Now in your Scala code you can access those fields.
-If you just access them directly, they’ll both be `Optional` values:
+First import the `scala.jdk.OptionConverters` object, and then use the `toScala` method to convert the `Optional` value to a Scala `Option`:
```
julienrf, line 111: What you really import is the converters _within_ the object `OptionConverters`

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
+  def wag = println("Tail is wagging")
 
+def useJavaInterfaceInScala =
+  val d = Dog()
+  d.speak
+  d.wag
+  d.run
+```
+{% endtab %}
+{% endtabs %}
```
julienrf, line 192: Should we also mention the fact that “nilary” methods in Java can be called both with and without parens (ie, `d.run` and `d.run()` are both valid)?

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
+For example, suppose that a Scala API returns a `List[String]` like this:
+
+{% tabs baz-definition class=tabs-scala-version %}
+{% tab 'Scala 2' %}
 ```scala
-// scala
-class ScalaClass:
-  val strings = List("a", "b")
+object Baz {
+  val strings: List[String]
```
julienrf, line 205: Abstract vals in objects are invalid in Scala too. I think it is important to always show valid code examples, even if I understand your idea here. We could either add ` = ???`, or turn `Baz` into a `trait`.

_overviews/scala3-book/interacting-with-java.md
```diff
@@ -39,209 +39,277 @@ Note that the Java examples in this section assume that you’re using Java 11 o
...
 
+{% tabs qux-definition class=tabs-scala-version %}
+{% tab 'Scala 2' %}
 ```scala
-// scala
-object ScalaObject:
-  val someString = Option("foo")
-  val noneString: Option[String] = None
+object Qux {
+  val optString: Option[String]
```
julienrf, line 250: Same here.

julienrf (changes requested):

_overviews/scala3-book/interacting-with-java.md
bishabosha: right, that makes sense. Really I only wanted to list the signature rather than show source code, not sure the best way

bishabosha: I added in an implementation.

_overviews/scala3-book/interacting-with-java.md
bishabosha: I also added in an implementation.

_overviews/scala3-book/interacting-with-java.md
bishabosha: done

_overviews/scala3-book/interacting-with-java.md
bishabosha: I added an explanation

_overviews/scala3-book/interacting-with-java.md
bishabosha: I added in an implementation.

_overviews/scala3-book/interacting-with-java.md
bishabosha: I also added in an implementation.

julienrf (approved):
Thank you Jamie!

PR #2873, opened 2023-07-17
port library authors guide to Scala 3

bishabosha:
guide already mentions scala 3.3, but use code tabs for the code example


julienrf (approved):
Thank you!

PR #2874, opened 2023-07-22
minor fixes to concurrency page

jamesma100:
If the user has been following along with a Scala REPL, `eventualInt` will have already been completed, and thus will return a result. Calling the function again and wrapping it in a Future will deterministically produce the intended result of an uncompleted Future.


bishabosha (approved):
Thanks for the suggestion!

And to clarify the other occurrences of `eventualInt` above in the page still make sense to keep because the section is talking about the time it takes to complete the future

PR #2898, opened 2023-09-07
fix outdated default value of minThreads

floh0:
Fix an outdated default value for the `scala.concurrent.context.minThreads` VM attribute used in the global execution context.

This value was correct when this page was written but it is not the case anymore, since this commit:
https://github.com/scala/scala/commit/e230409c13de167b0f4010464c74328ff91d9043

The default value is now "1" as described in the JavaDoc of the global execution context:
https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/ExecutionContext.scala#L127

Or in the implementation directly:
https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/impl/ExecutionContextImpl.scala#L92


bishabosha (approved):
thanks!

PR #2908, opened 2023-09-26
Update http-client-upload-file.md

spamegg1:
- in "Multi-part requests" the `file1` and `file2` have to be of type `File` instead of `Path`

### DIFF
--- a/_overviews/toolkit/http-client-upload-file.md
+++ b/_overviews/toolkit/http-client-upload-file.md
@@ -49,8 +49,8 @@ If the web server can receive multiple files at once, you can use a multipart bo
 ```scala
 import sttp.client4.quick._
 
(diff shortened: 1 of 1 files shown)


SethTisue:
@szymon-rd do you remember who wrote the http-client tutorial? since that person could best review this

szymon-rd:
Sadly, I don't. Blame disappeared on squash. However, it seems to work with just Path. Are there some cases where it doesn't?

spamegg1:
Weird, it's working now:
workingNow
It was giving type errors at `file1` and `file2` before. That was probably a Metals issue? (I did receive Scala-cli and Metals updates since then.) 

It works with `.toFile()` (`File`) or without it (`Path`), so it seems `multipartFile` can work with both types.

PR #2911, opened 2023-10-06
Fix code in `Scala Tour > Singleton Objects > Companion Objects`

VMironiuk:
Imported `Pi` from `scala.math` to resolve the "Not found: Pi" compilation error.


Philippus (changes requested):
Can you make the imports in the same style for scala 2 and 3?

Philippus (approved):

Philippus:
Thanks!

PR #2914, opened 2023-10-10
fix typo and bump scala 2.13 version

johnduffell:
I discovered that the suggested code did not compile, because it uses `libraryDependency` (doesn't exist) rather than `libraryDependencies`

This PR fixes that typo, and also updates the example version from 2.13.11 to 2.13.12 (latest)

### DIFF
--- a/_overviews/scala3-migration/tutorial-macro-mixing.md
+++ b/_overviews/scala3-migration/tutorial-macro-mixing.md
@@ -106,8 +106,8 @@ lazy val example = project.in(file("example"))
 
 lazy val `example-compat` = project.in(file("example-compat"))
   .settings(
-    scalaVersion := "2.13.11",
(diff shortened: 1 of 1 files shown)


_overviews/scala3-migration/tutorial-macro-mixing.md
```diff
@@ -106,8 +106,8 @@ lazy val example = project.in(file("example"))
 
 lazy val `example-compat` = project.in(file("example-compat"))
   .settings(
-    scalaVersion := "2.13.11",
-    libraryDependency += "org.scala-lang" % "scala-reflect" % scalaVersion.value
+    scalaVersion := "2.13.12",
+    libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value
```
johnduffell, line 110: this is the key fix

sjrd (approved):

PR #2918, opened 2023-10-24
Escaping asterisk(forthcoming)

dongxuwang:
It was supposed to be like:

\* = forthcoming; support available in [nightly builds](https://stackoverflow.com/q/40622878/86485)

instead of:
* = forthcoming; support available in [nightly builds](https://stackoverflow.com/q/40622878/86485)


bishabosha (approved):
Thanks! I tested this locally

SethTisue:
thank you

PR #2922, opened 2023-10-31
Fix typo in variances.md

adpi2:
We need the `new` keyword to create an anonymous class


odersky:
LGTM, thanks for the quick fix!

Issue #2648, opened 2022-11-21
Documentation states that "type inference proceeds one parameter list at a time" but this is not true in Scala 3.


smarter:
This is stated in https://docs.scala-lang.org/tour/multiple-parameter-lists.html but while this mostly applies to Scala 2, it doesn't apply at all to Scala 3 where type variables are inferred as late as possible (but not later than needed). I discuss the differences between Scala 2 and 3 here in my talk on type inference: https://youtu.be/lMvOykNQ4zs?t=1216

julienrf:
Could you suggest what should be the appropriate explanation in Scala 2 and in Scala 3?

smarter:
I would just completely remove the "Drive type inference" section, or make it Scala 2 specific.

PR #2924, opened 2023-11-14
FAQ for type inference

som-snytt:
Documents https://github.com/scala/bug/issues/7743

The tour still needs tweaking. per https://github.com/scala/docs.scala-lang/issues/2648

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -196,6 +196,19 @@ be written using the `_` syntax.
 
 See also [SLS 6.23.2](https://scala-lang.org/files/archive/spec/2.13/06-expressions.html#placeholder-syntax-for-anonymous-functions).
 
+### Why doesn't my dotted expression work?
+
+When methods have type parameters, the compiler makes a best effort to infer them
+if they are not supplied explicitly in a method application.
+Type parameter inference is improved in Scala 3.
+
+```
+scala> List("ab").toSet.head.head
+                             ^
+       error: value head is not a member of type parameter B
+```
+For more context, see [this section of the tour](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#drive-type-inference).
+
 ### Can I chain or nest implicit conversions?
 
 Not really, but you can [make it work](https://stackoverflow.com/a/5332804).
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -196,18 +196,21 @@ be written using the `_` syntax.
 
 See also [SLS 6.23.2](https://scala-lang.org/files/archive/spec/2.13/06-expressions.html#placeholder-syntax-for-anonymous-functions).
 
-### Why doesn't my dotted expression work?
+### Why couldn't Scala infer the correct type in my code?
 
-When methods have type parameters, the compiler makes a best effort to infer them
-if they are not supplied explicitly in a method application.
-Type parameter inference is improved in Scala 3.
+It is difficult to generalize about type inference, because various features of the language
+affect how your code is construed. There may be several ways to rewrite your code to make
+the types fall out naturally.
 
-```
-scala> List("ab").toSet.head.head
-                             ^
-       error: value head is not a member of type parameter B
-```
-For more context, see [this section of the tour](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#drive-type-inference).
+The most straightforward workaround is to supply explicit types in your code.
+
+That may involve specifying an explicit type to a definition, or a type argument to a method.
+
+Type inference is greatly improved in Scala 3. If Scala 2 doesn't compile your code, it's worth trying with Scala 3.
+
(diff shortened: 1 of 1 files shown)


SethTisue:
@som-snytt wdyt about the following version?

### Why couldn't Scala infer the correct type for my code?

This is difficult to generalize about, as there could be a variety of causes and possible solutions.

Type inference was greatly improved in Scala 3. If you're using Scala 2, you might try your example in Scala 3 to see if it's handled any better.

The most common workaround is simply to add an explicit type to your code.

Sometimes breaking up a parameter list into multiple sections helps, as per [this language tour section](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#drive-type-inference).

Questions about type inference often involve `toSet`. For discussion of examples involving `toSet`, see:

* https://github.com/scala/bug/issues/7743
* https://stackoverflow.com/questions/5544536/in-scala-2-type-inference-fails-on-set-made-with-toset

SethTisue:
@BalmungSan this might interest you as well

som-snytt:
### If it's called dotty, then why doesn't my dotted expression work?

### If I'm implying the correct type for my code, then why can't Scala infer it?

  Are you using "imply" and "infer" correctly?

### Expectation or Anticipation

I see my PR was more of a placeholder.

The github states are draft (WIP, drafts cannot be merged) and open green (ready for review), so if you're reviewing, it should be green. I don't know why there isn't a third state for "shared draft" that precedes green. That would mean, even if somebody accidentally grants an accepting review, don't merge it, as we're still mulling.

BalmungSan:
Personally, I really like this FAQ addition.

My proposal for the title may be:

## Why couldn't Scala infer the correct type for my dot-chain expression?

Additionally, I think providing a concrete example of the suggested solutions would be useful, as well as explaining why that fixes the problem.

Like:

```scala
// Chane this:
List("ab").toSet.head.head

// Into this:
val set = List("ab").toSet
set.head.head
// This works because ...

// Another option is:
List("ab").toSet[String].head.head
// This also works because ...
```

SethTisue:
> Additionally, I think providing a concrete example of the suggested solutions would be useful, as well as explaining why that fixes the problem

I'm consciously choosing not to do this. In general, I seek to keep FAQ entries short and sweet. A brief answer, and then perhaps a links to whatever other sources on the Internet (or on our own documentation site) that expound further.

BalmungSan:
> I'm consciously choosing not to do this. In general, I seek to keep FAQ entries short and sweet. A brief answer, and then a link or two or three to whatever other sources on the Internet

Oh right, forgot about that.
In that case I still expect whatever link is chosen to have such examples and explanations :)
Be it the official **Scala** documentation or a **StackOverflow** question / answer as other entries.

som-snytt:
OK I will update with something like Seth's text. I'm having trouble with any title that says "dot". We say "dotted name" or similar for `com.acme.foo`, but "what do you call it when an expression uses a selection?" "Selection" seems too "specialized" a word for a FAQ. One reason this is a FAQ is that it's not obvious why I can't rewrite expressions in obvious ways and have it just work. You can say Liskov till you're blue in the face, referential transparency blah-blah, but if simple syntax tweaks break my code, then I will be confused and angry. I don't think a link to a ticket suffices, but a link to the Big Book of Scala can be supplied later.

som-snytt:
ok I updated the section title.

### toSet or not toSet, that is the question

_overviews/FAQ/index.md
```diff
@@ -196,18 +196,21 @@ be written using the `_` syntax.
...
-For more context, see [this section of the tour](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#drive-type-inference).
+The most straightforward workaround is to supply explicit types in your code.
+
+That may involve specifying an explicit type to a definition, or a type argument to a method.
+
+Type inference is greatly improved in Scala 3. If Scala 2 doesn't compile your code, it's worth trying with Scala 3.
+
+Sometimes, using multiple parameter lists helps inference, as explained in [this section of the language tour](https://docs.scala-lang.org/tour/multiple-parameter-lists.html#drive-type-inference).
+
+For paradigmatic questions about type inference involving `toSet`, see the discussions on [this ticket](https://github.com/scala/bug/issues/7743) and a related [Q&A](https://stackoverflow.com/questions/5544536/in-scala-2-type-inference-fails-on-set-made-with-toset).
```
SethTisue, line 213: overall LGTM but I ask for one change: s/paradigmatic/common/

som-snytt: ok I accept pitching it to the commons

som-snytt:
I neglected to mention that I inlined the links, to make it ASAP (short as preferred).

I changed `paradigmatic question` to `stupid` before the final edit.

BalmungSan (approved):

som-snytt:
It should say, "If you are here, you should ask BalmungSan".

Issue #2933, opened 2023-12-13
Missing reference for "other advanced types" ?


vreuter:
The "other types" [page](https://docs.scala-lang.org/scala3/book/types-others.html) of the Scala 3 book lists a variety of "other advanced types" and links to the [reference documentation](https://docs.scala-lang.org/scala3/reference/index.html) "for more details on these types." For most of the listed sorts of types, there is reference easily accessible on the linked page, but for 2 of them there's nothing apparent (at least not when Ctrl+F searching on the page for these sorts of types):
- [ ] Singleton types
- [ ] Refinement types

It seems that either reference should be added for these on the linked page, or that more specific link(s) should be provided on the Scala 3 book page. Did I miss something, though?

vreuter:
<img width="1362" alt="Capture d’écran 2023-12-13 à 12 09 28" src="https://github.com/scala/docs.scala-lang/assets/12036370/a2208486-2d97-49e7-82d4-a179d56dcb87">

bishabosha:
Thanks for the feedback! that is indeed a missing information that is only visible in the [spec](https://scala-lang.org/files/archive/spec/3.4/03-types.html)

vreuter:
> Thanks for the feedback! that is indeed a missing information that is only visible in the [spec](https://scala-lang.org/files/archive/spec/3.4/03-types.html)

Many thanks for the link @bishabosha

gkepka:
Completed by #2934

PR #2934, opened 2023-12-14
Add info pointer for singleton and refinement types, #2933

vreuter:


SethTisue:
We don't normally link from ordinary documentation — especially tutorial documentation like the Scala 3 Book — to the language specification, which is so much more formal and technical. We might link to the spec if no other reference documentation was available, but in this case, reference documentation is available and we already link to it.

(Does the reference documentation link to the relevant section of the spec? If not, it could and should.)

SethTisue:
Oh, i see on the ticket that maybe @bishabosha doesn't agree. wdyt @bishabosha ?

vreuter:
> Does the reference documentation link to the relevant section of the spec? If not, it could and should.)

Thanks for the context distinction b/w book and language spec and for suggesting the specific section links @SethTisue . I've added the more specific links here: https://github.com/scala/docs.scala-lang/pull/2934/commits/29c0b72dd852904e16df94fb9118f9e074e50626

vreuter:
> Ideally I would suggest that you write a dedicated section, maybe you would like to come back and do that?

@bishabosha would you mean in the Book (e.g., https://docs.scala-lang.org/scala3/book/types-intersection.html) or in the Reference (e.g., https://docs.scala-lang.org/scala3/reference/new-types/intersection-types.html#)?

PR #2939, opened 2023-12-22
Fix 404s

Philippus:


bishabosha (commented):
I think this is generated and overridden by a cron process in scala/improvement-proposals

Philippus:
Indeed, I see I tried to do the same in https://github.com/scala/docs.scala-lang/pull/2937 already. :)

Philippus:
Created this PR: https://github.com/scala/improvement-proposals/pull/70.

PR #2955, opened 2024-01-24
remove 404 link in Scala on Android tutorial

bishabosha:
it was a Google Drive link to a pre built demo of the application. It's not critical to reading the tutorial.

@makingthematrix is this ok, or do you want to make another link?


makingthematrix:
@bishabosha : Yes, I think it's the best to just delete that link and the sentence.

Instead, maybe we could add a new paragraph about [Fx Calculator](https://github.com/makingthematrix/scalaonandroid/tree/main/fxcalculator)? It's a more advanced app based on the same tech stack.

SethTisue:
> Instead, maybe we could add a new paragraph about Fx Calculator

PR welcome!

PR #2960, opened 2024-01-31
CI: bump actions versions

SethTisue:
GitHub is whining (rightly, I guess) about this at e.g. https://github.com/scala/docs.scala-lang/actions/runs/7731097673

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,14 +4,14 @@ jobs:
   build:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


SethTisue:
Even with the bumps, we still get:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: coursier/setup-action@v1.3.4. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

Issue #925, opened 2017-11-10
Pages with TOCs not navigable when window width <= 768px


aaron-kaplan:
The table of contents is not visible on my phone, which makes it impossible to navigate the Scala Tour. I can reproduce this behavior on desktop by resizing the browser window to 768 pixels wide.

[description edited after clarification below]

heathermiller:
Yes, this is by design. I wouldn't say that it's "impossible to navigate the site" though, is it? You just don't get a table of contents for the page you're viewing... You still have all other navigation (top navbar, footer, etc)

aaron-kaplan:
I should have said that it makes it impossible to navigate the tour. The top nav and footer have links to other parts of the site, but I don't see links to other pages of the tour. The intro page has in-line links to many of the pages of the tour, but not all of them. Am I missing something?

heathermiller:
Aha, yes ok, you're right. The Scala Tour is a multi-page document that actually needs the table of contents to navigate. I guess it's the only exception on the site. Good point–so we should figure out a way to show a TOC for at least this one document...

marconilanna:
> The Scala Tour is a multi-page document that actually needs the table of contents to navigate. I guess it's the only exception on the site.

@heathermiller it also affects multiple overview documents:

* http://docs.scala-lang.org/overviews/reflection/overview.html
* http://docs.scala-lang.org/overviews/macros/usecases.html
* http://docs.scala-lang.org/overviews/quasiquotes/setup.html
* http://docs.scala-lang.org/overviews/collections/introduction.html
* http://docs.scala-lang.org/overviews/parallel-collections/overview.html
* http://docs.scala-lang.org/overviews/scaladoc/overview.html

SethTisue:
the tour sections have "previous" and "next" links now, but that doesn't help the other pages.

rkoeninger:
These lines are what cause the TOC to be hidden when the page width is less than 768px:

https://github.com/scala/docs.scala-lang/blob/44840f559bff75de755b04c457838d046c5edbd4/_sass/layout/toc.scss#L6-L8

Should this just be removed? If it were, the TOC would always be visible, even on small screens. TOC isn't that big, not sure why it would need to be hidden.

And forcing it to appear only for certain pages would be complicated. (Maybe there would be a flag in the article's markdown that causes a class to be added or removed from the TOC div and the css in the media query includes that class, or `:not()`?)

Or just always show the TOC.

gkepka:
Completed by #2969

PR #2969, opened 2024-02-14
Always show TOC and move it to bottom on mobile

Dieken:
Fixed issue #925.


Dieken:
CI failure isn't caused by this PR.

bishabosha:
hey, sorry I forgot to look at this

bishabosha:
I think it's not obvious when reading the page that the TOC even exists because it's after everything, maybe users won't scroll far enough to notice it.

perhaps there should be a banner at the top/bottom that links to the TOC? (or a sticky one that follows you as you read)

Screenshot 2024-02-23 at 10 31 28

bishabosha (changes requested):
I think it should be more clear how to find the toc on mobile

bishabosha:
still it could be a follow-up PR to add the links

Dieken:
> I think it should be more clear how to find the toc on mobile

It’s better to convert the whole table of content into a menu button on narrow screen, touching or clicking the menu button will expand or collapse the table of content.

But that change is too difficult for me, this patch put the toc bottom, it’s already much better now:

* the first page is reached from header menu, so it’s fine to not see the toc immediately, actually the toc is too long, it must’t be placed on the top of content.

* when finish one chapter, the reader will notice the toc.

* if the reader doesn’t jump on chapters, he can just follow the prev/next links, same experience as before, because the toc never shows before the content.

So, it’s good enough for me, very usable now:-)

PR #2970, opened 2024-02-14
Clearer points in favour of macro mixing

johnduffell:
This PR updates the scala 3 migration docs to emphasis the benefits of macro mixing rather than indicating that it's experimental.

--

The background is as follows:

We are trying to update our code gradually to scala 3.  In one case we have a large multi project build which produces many semi-related jar files for deployment into AWS lambda.

I am trying to update this module by module, but I quickly hit an issue where play-json had scala 2.13 and 3 versions.

Normally I would just force version 3 and use `-Ytasty-reader`, however in this case because it's a macro library, the modules remaining on scala 2.13 didn't compile any more.

I proposed adding macro mixing to the library, but this was pushed back on the basis that it wasn't necessary and is not promoted by the docs.  See the issue below:
https://github.com/playframework/play-json/issues/954

I think this is not really accurate (is it?) and I think this is going to be a blocker for all of our larger repos in our company, and possibly other organisations.

### DIFF
--- a/_overviews/scala3-migration/compatibility-metaprogramming.md
+++ b/_overviews/scala3-migration/compatibility-metaprogramming.md
@@ -63,7 +63,10 @@ You can learn all the new metaprogramming concepts by reading the [Macros in Sca
 ## Cross-building a Macro Library
 
 You have written a wonderful macro library and you would like it to be available in Scala 2.13 and Scala 3.
-There are two different approaches, the traditional cross-building technique and the more experimental mixing macro technique.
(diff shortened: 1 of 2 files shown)


bishabosha (approved):
I was surprised to see the phrasing "Experimental" in the old doc. The only caveat there is if the macro implementation method was compiled by Scala 3 then you can't use things like quasiquotes or eval, but if the macro method is in a separate 2.13 compiled module then this is no issue

PR #2971, opened 2024-02-14
Fixed broken "Scala 3" sub menu on mobile

Dieken:
CSS doesn't support id selector containing whitespace, see

https://github.com/scala/docs.scala-lang/blob/d626b36e217a99ee078cdcfbc30d23386dc7fb00/resources/js/functions.js#L90


Dieken:
CI failure isn't caused by this PR.

Dieken:
> Another option could be replace ` ` with `__`, `_` or `-` possibly?

All are fine, the js script doesn’t hard code the id, I prefer simple word “scala3”.

Thanks for your merging!

PR #2972, opened 2024-02-14
Fixed broken code highlight on Safari of iOS 15.7.8

Dieken:
On iOS, the "type `A` \*/" is rendered as "type `A` //", seems highlight.js is incompatible with iOS Safari, wrongly replaces "*/" with "//".

I'm not able to find the root cause, thus change to inline comment to bypass the issue.

### DIFF
--- a/_overviews/scala3-book/ca-context-bounds.md
+++ b/_overviews/scala3-book/ca-context-bounds.md
@@ -44,12 +44,12 @@ existing method `max` on `List`, but we made up this example for illustration pu
 
 {% tab 'Scala 2' %}
 ```scala
-/** Defines how to compare values of type `A` */
+// Defines how to compare values of type `A`
 trait Ord[A] {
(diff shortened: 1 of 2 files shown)


Dieken:
image

Upload the screenshot to show the issue.

bishabosha:
im not sure what your original comment means, e.g. here is what I see on iOS:

IMG_A62D691D6578-1

Dieken:
> im not sure what your original comment means, e.g. here is what I see on iOS (it is iOS 17.2.1):
> 
So it's a bug in webkit engine of iOS 15.7.8, I tried Chrome on iOS, same issue as the screenshot above,  the trailing "*/" is rendered as "//" and the **whole** code snippet is rendered as comment.

Dieken:
Any more comment?

bishabosha:
there are 175 of occurrences of `/*` in the docs, including the non-english translations of this page. Did you see any others that had the same issue?

PR #2985, opened 2024-02-29
Link out to wiki page for the concept of a "generator"

hongtron:
The term "generator" is used in a couple spots in the docs (e.g. in [For Comprehensions](https://github.com/scala/docs.scala-lang/blob/main/_tour/for-comprehensions.md) in the Tour of Scala), but is never defined. As an alternative to this change, I welcome (and would even prefer) proposed definitions for the term, if the Scala concept does not completely line up with the wiki content - but I do not feel qualified to venture a definition myself.


_overviews/scala3-book/control-structures.md
```diff
@@ -236,7 +236,7 @@ for i <- ints do println(i)
 {% endtabs %}
 
 
-The code `i <- ints` is referred to as a _generator_.
+The code `i <- ints` is referred to as a [_generator_](https://en.wikipedia.org/wiki/Generator_(computer_programming)).
```
bishabosha, line 239: really the only thing "generator" means in scala is the syntax `p <- e` itself. That wikipedia definition would be inappropriate in this case

bishabosha: so the definition of generator is `p <- e` where `e` is an expression, and `p` is a pattern

bishabosha: a for-comprehension itself (i.e. the whole `for (exprs) yield e` fulfils the wikipedia generator definition better

bishabosha: this is the definition in [the spec](https://scala-lang.org/files/archive/spec/3.4/06-expressions.html#for-comprehensions-and-for-loops)

> A generator `p <- e` produces bindings from an expression `e` which is matched in some way against pattern `p`.

bishabosha:
Hey, thank you for opening this PR! We can definitely do more to make a reusable definition to explain what is meant by "generator". Probably the best is from [this page](https://docs.scala-lang.org/scala3/book/control-structures.html)

> The code `i <- ints` is referred to as a generator.
 
Perhaps we can copy that to [the tour](https://docs.scala-lang.org/tour/for-comprehensions.html) and [a taste of scala](https://docs.scala-lang.org/scala3/book/taste-control-structures.html#for-loops-and-expressions)

My suggestion for improvement would be

> The code `i <- ints` is referred to as a generator. In any generator `p <- e`, the expression `e` can generate zero or many bindings to the pattern `p`.

hongtron:
Thank you for your feedback! I've pushed some changes to incorporate your definition, and update the other spots you called out.

I also made two other wording changes, which I've broken into separate commits for emphasis (one regarding the phrase "semicolon-separated," and one replacing "filter" with "guard"). Let me know if you think they make sense.

_tour/for-comprehensions.md
```diff
@@ -12,7 +12,7 @@ redirect_from:
   - "/tutorials/tour/sequence-comprehensions.html"
 ---
 
-Scala offers a lightweight notation for expressing _sequence comprehensions_. Comprehensions have the form `for (enumerators) yield e`, where `enumerators` refers to a semicolon-separated list of enumerators. An _enumerator_ is either a generator which introduces new variables, or it is a filter. A comprehension evaluates the body `e` for each binding generated by the enumerators and returns a sequence of these values.
+Scala offers a lightweight notation for expressing _sequence comprehensions_. Comprehensions have the form `for (enumerators) yield e`, where `enumerators` refers to a list of enumerators. An _enumerator_ is either a generator, or it is a guard (see: [Control Structures](/scala3/book/control-structures.html)). A comprehension evaluates the body `e` for each binding generated by the enumerators and returns a sequence of these values.
```
bishabosha, line 15:
```suggestion
Scala offers a lightweight notation for expressing _sequence comprehensions_. Comprehensions have the form `for (enumerators) yield e`, where `enumerators` refers to a list of enumerators. An _enumerator_ is either a generator, or it is a guard (see: [Control Structures](/scala3/book/control-structures.html#for-loops)). A comprehension evaluates the body `e` for each binding generated by the enumerators and returns a sequence of these values.
```

bishabosha (approved):
Thanks for doing this!

PR #2986, opened 2024-03-01
ignore more GitHub links in HTMLProofer

bishabosha:


sjrd (approved):

SethTisue:
for context in case anyone comes back to this, HTMLProofer keeps failing because GitHub sometimes return 404s on valid URLs, and it's become too consistently annoying to allow. it's a bit sad to miss the link checking on this but oh well

PR #2987, opened 2024-03-01
Update links to use scala/scala3 instead of lampepfl/dotty

hamzaremmal:
Part of https://github.com/lampepfl/dotty/pull/19839 and https://github.com/lampepfl/dotty/issues/19848


bishabosha:
this should wait right?

hamzaremmal:
It is ready for approval, but it should not be merged until the transfer is done.

hamzaremmal:
@bishabosha it would be great if you can review this before tomorrow :-)

bishabosha (approved):

PR #2989, opened 2024-03-02
Fix tab name

artemkorsakov:
Tab names must be unique to avoid the following error:

Page: https://docs.scala-lang.org/scala3/book/methods-main-methods.html

Screen:
image


artemkorsakov:
@bishabosha @sjrd please, review this PR

bishabosha:
its strange that you get this error because I do not

bishabosha:
ok it looks like https://github.com/scala/docs.scala-lang/pull/2991 already solved it

artemkorsakov:
Yes, it's the same as https://github.com/scala/docs.scala-lang/pull/2991

PR #2992, opened 2024-03-04
improve description about enumerator in for-comprehensions

laglangyue:
Student of Scala may have some doubts when learning for-enumerators, and may not understand the execution order of `if enumerator`. They may think that if is similar to match grammer, they think `if i + j == v` will execute before `j <-...`.
look at this example
```scala
    val a = Array(1, 2, 3)
    val value = for
      i <- 0 to 5;
      j = a(i) if i % 2 == 0
     yield (j)
```
Actually, it should be 
```
    val a = Array(1, 2, 3)
    val value = for
      i <- 0 to 5 if i % 2 == 0
      j = a(i) 
     yield (j)
```

### DIFF
--- a/_zh-cn/tour/for-comprehensions.md
+++ b/_zh-cn/tour/for-comprehensions.md
@@ -35,7 +35,8 @@ twentySomethings.foreach(name => println(name))  // prints Travis Dennis
 ```scala mdoc
 def foo(n: Int, v: Int) =
    for (i <- 0 until n;
-        j <- i until n if i + j == v)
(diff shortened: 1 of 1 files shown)


bishabosha:
@scala/docs-zh

jxnu-liguobin (approved):

laglangyue:
sorry，I check this PR again，My friend  doesn't read the documentation very well.
This section has explained very clearly how if works in for-comprehensions.
image

PR #3004, opened 2024-03-30
Correct 2.13 Java 22 compatibility

octylFractal:
The Java 22 compatibility was only merged into 2.13.13.


Philippus:
@SethTisue can you verify/clarify this? As you're also working on https://github.com/scala/docs.scala-lang/pull/3005.

SethTisue (approved):
Good catch, thanks! I also applied the same fix farther down the page.

PR #3007, opened 2024-04-15
Fix wrong naming in variables and method

gojideth:
image
I noticed some errors with the arguments, they were not following the naming in Spanish. Also it should call `imprimirNombre` instead of `printName` and I fixed a small typo.


eed3si9n (approved):
Thanks!

PR #3009, opened 2024-04-18
Update docs for `-release` to make it clear it accepts recent Java versions

rtyley:
The old displayed values of `6, 7, 8, 9` are confusing for developers contemplating updating to recent Java versions like `21` - [eg](https://github.com/guardian/mobile-save-for-later/pull/109/files#r1559390842)  :

> I was getting an error that -release:21 is not a valid compiler option when I ran the code. Looking at the docs: https://docs.scala-lang.org/overviews/compiler-options/ my understanding is that this is not supported for Java 21.

That's an incorrect understanding from the docs, as newer Java versions definitely are supported by the `-release` Scala compiler parameter - so it's best to clarify that!

In this case, the developer was getting the error because they were accidentally running with an older Java version (the compiler error could probably be clearer there too).

### DIFF
--- a/_data/compiler-options.yml
+++ b/_data/compiler-options.yml
@@ -260,7 +260,7 @@
       type: "String"
       arg: "release"
       default:
-    description: "Compile for a specific version of the Java platform. Supported targets: 6, 7, 8, 9"
+    description: "Compile for a specific version of the Java platform. Supported targets: 8, 11, or any higher version listed at https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html"
     abbreviations:
     - "--release"
   - option: "-sourcepath"


som-snytt (approved):
This is OK but likely to be overwritten if someone runs the automation.

rtyley:
> This is OK but likely to be overwritten if someone runs the automation.

Ah, thank you! I didn't realise this file was auto-generated by automation - searching for the filename, I now see [a comment](https://github.com/scala/docs.scala-lang/issues/1711#issuecomment-578364758) that mentions the [`generateDocsData`](https://github.com/scala/scala/blob/ec97e821e03facf216626ff959dfb39d358ec883/build.sbt#L1215-L1221) command:

> over in the scala/scala repo, running sbt followed by generateDocsData generates a compiler-options.yml file, and then the web page is generated from that 

I guess `generateDocsData` isn't run as part of the release process for Scala releases - I tried running it, and got quite a big diff on `compiler-options.yml`.

PR #3010, opened 2024-04-21
Add `Matching function expressions`

arainko:
It seems like no one on the internet knows how to do this properly, so after drunkenly  stumbling upon the correct syntax I decided to throw this small section together.


som-snytt (approved):
IDK but it is standalone

arainko:
...and I got to actually put it to real use [just yesterday](https://github.com/arainko/ducktape/blob/4604d0769920e8f9e484ddf096c12a4e62bbe292/ducktape/src/main/scala/io/github/arainko/ducktape/internal/ParseRenamer.scala), what a happy little coincidence

SethTisue:
Thank you!

PR #3012, opened 2024-05-02
Fix illegal type of expr in scala3 macros overview

road21:


road21:
@SethTisue @julienrf
Could you please take a look?

PR #3014, opened 2024-05-07
Toolkit tutorials: update Scala version

SethTisue:
a language newcomer on Discord hit this today — the Toolkit examples didn't work because the user used a newer sttp milestone that was built on Scala 3.3 and thus didn't work with 3.2


adpi2:
> @Kordyjan could keeping this up-to-date be part of the Scala 3 release steps?

sttp is probably going to stay on Scala 3.3 and all Scala 3.3.x versions are compatible (backward and forward). So maybe we don't need to make it part of the Scala 3 release.

adpi2 (approved):

PR #3016, opened 2024-05-07
in Toolkit sbt example, use flat single-project style

SethTisue:
a language newcomer on Discord was bitten by this today

note that the examples weren't just unnecessarily complicated, they also used `file("example")` instead of `file(".")`, so the user ended up having both a `root` and an `example` project, and they didn't understand that they needed to put their sources in `example/src`, because how on earth would they know that?

we can avoid issues like that by simply using the flat style


adpi2:
I am not convince about that because I think we should slowly discourage people from using the sbt flat style:
- the difference between flat, and `ThisBuild`, and `Global` and `project.in(file("."))` is non-obvious for beginners
- for single module project we should better use Scala CLI whose syntax is as flat as it can be

However I am okay to use `file(".")` instead of `file("example")`. It seems that's what we do in `scala/scala3.g8` [template](https://github.com/scala/scala3.g8/blob/main/src/main/g8/build.sbt).

SethTisue:
okay, I'll make the change

adpi2:
LGTM! Is it ready to be merged?

PR #3017, opened 2024-05-07
Update FAQ on initialization order

lrytz:


_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
-Naturally when a val is overridden, it is not initialized more than once.  So though x2 in the above example is seemingly defined at every point, this is not the case: an overridden val will appear to be null during the construction of superclasses, as will an abstract val.
+When a `val` is overridden, in fact its accessor method (the "getter") is overridden.
+So the access to `x2` in class `A` in fact invokes the overridden getter in class `C` which reads the underlying field `C.x2`.
+This field is not yet initialized during the construction of `A`.
 
-There is a compiler flag which can be useful for identifying this situation:
+## Mitigation
 
-**-Xcheckinit**: Add runtime check to field accessors.
+The [`-Ysafe-init` compiler flag](https://docs.scala-lang.org/scala3/reference/other-new-features/safe-initialization.html) in Scala 3 enables compiler warnings for accesses to uninitialized fields:
```
SethTisue, line 52: suggest mentioning here that the flag is experimental, like all `-Y` flags

_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
+      |      ├── abstract class A {	[ Test.scala:1 ]
+      |      │   ^
+      |      └── println("A: " + x1 + ", " + x2)	[ Test.scala:5 ]
+      |                          ^^
 
-Using it on the opening example:
+In Scala 2, the `-Xcheckinit` flag adds runtime checks in the generated bytecode to identify accesses of uninitialized fields.
+The code then throws an exception rather than allowing a `null` (or `0` / `false` in the case of primitive types) to silently appear.
+Note that these runtime checks only test code that is actually exectued at runtime.
+The flag can be helpful to find accesses to uninitialized fields, but it should never be used in production due to its performance overhead.
```
SethTisue, line 68:
```suggestion
The flag can be helpful to find accesses to uninitialized fields, but it should not be used in production due to its performance overhead.
```

_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
+      |      │   ^
+      |      ├── abstract class A {	[ Test.scala:1 ]
+      |      │   ^
+      |      └── println("A: " + x1 + ", " + x2)	[ Test.scala:5 ]
+      |                          ^^
 
-Using it on the opening example:
+In Scala 2, the `-Xcheckinit` flag adds runtime checks in the generated bytecode to identify accesses of uninitialized fields.
+The code then throws an exception rather than allowing a `null` (or `0` / `false` in the case of primitive types) to silently appear.
+Note that these runtime checks only test code that is actually exectued at runtime.
```
SethTisue, line 67:
```suggestion
Note that these runtime checks only test code that is actually executed at runtime.
```

SethTisue (commented):
very nice improvements. I've picked a few nits

_overviews/FAQ/initialization-order.md
som-snytt: That's like saying use it, but don't like _use_ it. How about instead of experimental, we start saying "cutting-edge". Also, experimental has the narrow meaning for dotty that you need a flag and an experimental compiler to use the feature.

_overviews/FAQ/initialization-order.md
som-snytt: "must never", and spoken with Germanic authority. People run debug in production all the time just in case there's a crash, which there always is.

som-snytt:
@SethTisue I suggest we don't wait for Lukas and definitely don't wait for Seb, who would incur more rounds of nits even pickier than previously surfaced.

_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
+      override def add(x: Int): Unit = { added += x; super.add(x) }
+    }
+
+### Early definitions: deprecated
 
-Sometimes all you need from an interface is a compile-time constant.
+Scala 2 supports early definitinos, but they are deprecated in Scala 2.13 and unsupported in Scala 3.
+See the [migration guide](https://docs.scala-lang.org/scala3/guides/migration/incompat-dropped-features.html#early-initializer) for more information.
 
-Constant values are stricter than strict and earlier than early definitions and have even more limitations,
```
som-snytt, line 163: Omitting compile-time mechanisms to work around runtime issues seems important. Probably more is available in Scala 3 besides final vals.

Let's also respect that this was the original one-question FAQ. Maybe a minute of silence?

_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
+In Scala 2, you can define an abstract `val` or `def` instead.
 
-Note that early definitions are deprecated in Scala 2.13; they will be replaced by trait parameters in Scala 3. So, early definitions are not recommended for use if future compatibility is a concern.
+An exception during initialization of a lazy val will cause the right-hand side to be re-evaluated on the next access: see SLS 5.2.
 
-#### Use constant value definitions ####
-    abstract class A {
-      val x1: String
-      val x2: String = "mom"
+Note that using multiple lazy vals creates a new risk: cycles among lazy vals can result in a stack overflow on first access.
```
sjrd, line 125: Or dead lock in Scala 3 for thread-safe lazy vals.

_overviews/FAQ/initialization-order.md
```diff
@@ -36,129 +36,125 @@ Let's observe the initialization order through the Scala REPL:
...
 
-      println("A: " + x1 + ", " + x2)
-    }
-    class B extends A {
-      val x1: String = "hello"
-      final val x3 = "goodbye"
+### Use a nested object
```
sjrd, line 127: Isn't that basically the same solution as a lazy val?

sjrd (approved):

_overviews/FAQ/initialization-order.md
bishabosha: it will also be `-Wsafe-init` from 3.5

SethTisue:
@lrytz care to return to this? you can self-merge if you think the review feedback has been addressed

_overviews/FAQ/initialization-order.md
som-snytt: Just noticed `definitinos`! Either I want the t-shirt, or I want to name the next subatomic particle.

> Definitinos allow the Scala Quantum Matrix (SQuaM) to infer your classically ambiguous implicit.

som-snytt:
I merged with uncontroversial tweaks. I believe I added no puns. I'm willing to follow up if I typoed something at 2:30 am.

PR #3024, opened 2024-05-18

marcogalluzzi:
Referencing new Scala 3 keyword `using`


som-snytt:
The example code is correct. It's possible that we refer to givens generally as implicits. We still say "implicit scope" and not "given scope".

> This section describes changes to the implicit resolution that apply both to the new givens and to the old-style implicits in Scala 3.

In recent PRs about the relative priority of givens and implicits, they are distinguished by their specific names.

PR #3026, opened 2024-06-02
Update translated word for the "print" word in Thai language

fResult:
Update the translated word "print" from "พิมพ์" to "แสดงค่า" to make it clearer


SethTisue:
well I don't know Thai but I'm happy to assume this is an improvement. thank you!

PR #3028, opened 2024-06-14
change function name to match method name

performantdata:
Change the lambda's name to match that of the corresponding method name (`double`) in the subsequent example.

### DIFF
--- a/_overviews/scala3-book/fp-functions-are-values.md
+++ b/_overviews/scala3-book/fp-functions-are-values.md
@@ -20,7 +20,7 @@ You’ve seen this in multiple places in this book, whenever methods like `map` 
 ```scala
 val nums = (1 to 10).toList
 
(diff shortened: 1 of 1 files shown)


Philippus (commented):
I think `doubles` is the correct name, as you get doubles after the map has executed.
Maybe further down this line `val doubles = nums.filter(underFive).map(double)` could change to `val doublesUnderFive = nums.filter(underFive).map(double)`.

performantdata:
@Philippus I think `double` is correct, as a verb that names the action that the function performs. But I don't disagree with this:
> `val doubles = nums.filter(underFive).map(double)` could change to `val doublesUnderFive = nums.filter(underFive).map(double)`.

Go with that if you want.

`double` also appears as the function name further down the page.

bishabosha:
this change does not bring value.

PR #3031, opened 2024-06-24
Update and correct macro cross tutorial

som-snytt:
Fixes https://github.com/scala/scala3/issues/20591

Besides the use of jpath, the scala 2 macro impl must be public. I tried to verify the steps as presented, but did not quite have the patience; probably the sophisticated user will fill in any gaps.

It's hard not to turn it into an sbt tutorial.

It's nicer just to say `+ test`.

### DIFF
--- a/_overviews/scala3-migration/tutorial-macro-cross-building.md
+++ b/_overviews/scala3-migration/tutorial-macro-cross-building.md
@@ -24,10 +24,11 @@ In order to exemplify this tutorial, we will consider the minimal macro library 
 lazy val example = project
   .in(file("example"))
   .settings(
-    scalaVersion := "2.13.11",
+    scalaVersion := "2.13.14",
     libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


_overviews/scala3-migration/tutorial-macro-cross-building.md
```diff
@@ -191,9 +192,9 @@ object Macros:
   private def locationImpl(using quotes: Quotes): Expr[Location] =
     import quotes.reflect.Position
     val pos = Position.ofMacroExpansion
-    val file = Expr(pos.sourceFile.jpath.toString)
+    val path = Expr(pos.sourceFile.path)
```
som-snytt, line 195: They preferred getJPath.getOrElse on the related PR. But I think source must have a path, but may not have a file. A source must have a path because I need to be able to report errors at a location, even if the location is virtual. It's not a Scala thing but an API thing.

PR #3032, opened 2024-06-25
Grammar and wording

neuroevolutus:
This PR adds minor grammatical and wording changes to the Tour of Scala. One thing in particular that I believe would be worth double-checking is whether my change to `_tour/annotations.md` is correct with respect to Scala's semantics.


bishabosha (approved):
Reads well to me, thank you

PR #3035, opened 2024-06-25
finish migration to code tabs in migration guide

bishabosha:


_overviews/scala3-migration/incompat-type-inference.md
```diff
@@ -68,19 +68,30 @@ class Child extends Parent {
...
 import scala.language.reflectiveCalls
 
 val foo = new {
   def bar: Unit = ???
 }
+```
+{% endtab %}
+{% endtabs %}
 
+However the Scala 3 compiler does not infer structural types by default, and so infers type `Object` for `foo`, rather than `{ def bar: Unit }` as in Scala 2.13, so the following structural selection fail to compile:
```
sjrd, line 84:
```suggestion
However the Scala 3 compiler does not infer structural types by default.
It infers the type `Object` for `foo` instead of `{ def bar: Unit }`.
Therefore, the following structural selection fails to compile:
```

sjrd (approved):
One suggestion to split a long sentence into three. Otherwise LGTM.

PR #3041, opened 2024-07-09
Update scala-features.md

stensh1:
The sentence "Например, переменные создаются лаконично, а их типы понятны:" does not correspond to the code example, there are no "var" but "val" instead.


som-snytt:
vals and vars are both called variables in an informal context. Otherwise, "mutable variable" would be redundant. Other distinctions I would prefer be universally honored include "definition" vs "declaration", and "its" vs "it's". Also a constant is a constant only if it is a mathematically precise value that never changes.

SethTisue:
I don't know Russian but I agree that the `val` should stay a `val`.

PR #3042, opened 2024-07-17
A new page about Scala IDEs + reordering of Getting Started

makingthematrix:
This is a draft for a page listing Scala IDEs. The corresponding ticket: https://github.com/scala/scala-lang/issues/1657

For now I added entries for IntelliJ IDEA + Scala Plugin and VS Code + Metals. I changed the "Getting Started" button to a foldable menu where the old page is now under "Install Scala", and the other entry in the menu is "Scala IDEs" with the new page.

Please do not merge it yet. I would ask someone from Metals to first take a look at their entry. I will also ask someone from the copyproof team at JetBrains to look at ours. 

PS. Curiously, the pages "Getting Started with Scala in IntelliJ" and "Getting Started with Scala and sbt" are physically in the `getting-started` folder, but on the webpage they are under the "Tutorials" menu. Since now "Getting Started" is a menu as well, we could move them there.


_overviews/getting-started/scala-ides.md
```diff
@@ -0,0 +1,45 @@
...
+
+For more information, check out our tutorial [Getting Started with Scala in IntelliJ](/getting-started/intellij-track/building-a-scala-project-with-intellij-and-sbt.html)
+
+# Visual Studio Code + Metals
+
+[https://scalameta.org/metals](https://scalameta.org/metals)
+
+![](../../resources/images/getting-started/VSCodeMetals.png)
+
+Visual Studio Code, commonly called VS Code, is a source code editor developed by Microsoft. Similar to how IntelliJ IDEA requires IntelliJ Scala Plugin to support Scala, you can get support for Scala in VS Code by installing an extension: Metals by Scalameta. In contrast to IntelliJ IDEA + Scala Plugin, VS Code + Metals is aimed at people who like to tinker with settings and try out experimental Scala features. Besides, Metals - the extension - is also available for a variety of other source-code editors, such as Vim, Sublime Text, and Emacs, which means that you will get a similar experience in any of them.
```
tgodzik, line 44: I wouldn't say that VS Code is complicated in terms of settings, other editors rather. Maybe we should highlight that is a LSP server?

`is aimed at people who like to get feedback and code intelligence straight from the compiler, which enables them to also try out experimental Scala features`

tgodzik, line 44:
```suggestion
Visual Studio Code, commonly called VS Code, is a source code editor developed by Microsoft. Similar to how IntelliJ IDEA requires IntelliJ Scala Plugin to support Scala, you can get support for Scala in VS Code by installing an extension: Metals by Scalameta. In contrast to IntelliJ IDEA + Scala Plugin, VS Code + Metals is is aimed at people who like to get feedback and code intelligence straight from the compiler, which enables them to also try out experimental Scala features. Besides, Metals - as an [language server](https://microsoft.github.io/language-server-protocol/) - is also available to use with a variety of other source-code editors, such as Vim, Sublime Text, Zed, Helix and Emacs, which means that you will get a similar experience in any of them.
```

SethTisue:
I haven't really followed how the Getting Started pages have evolved, so I'm not the best reviewer for that aspect; @bishabosha might like to have a look.

As for the text of the page itself, I have some ideas about wordings I would like to alter. But I suggest we do the review in two phases: let's get to something that's mergeable; let's merge it; then let's improve the text from there, including putting it out for public comment.

And I actually, I don't see anything here that's not mergeable.

_overviews/getting-started/scala-ides.md
```diff
@@ -0,0 +1,45 @@
...
+keywords:
+- Scala
+- IDE
+- JetBrains
+- IntelliJ
+- VSCode
+- Metals
+---
+
+By Maciej Gorywoda
```
SethTisue, line 18: Although we very much appreciate you providing a first version of the page, I suggest removing the author credit. We don't generally include them on short factual pages; I think the text is likely to evolve a lot over time; and we have found that having an author credit discourages edits.

SethTisue:
Oh, mind including an appropriate edit to https://www.scala-lang.org/community/index.html ?

makingthematrix:
> Oh, mind including an appropriate edit to https://www.scala-lang.org/community/index.html , so people can find the new page from there?

I'm not sure where to put it. Maybe a new section between "Who's behind Scala?" and "Ambassadors"?

SethTisue:
at the top of "Community Libraries and Tools"?

makingthematrix:
@SethTisue : Done. https://github.com/scala/scala-lang/pull/1677

SethTisue:
(Let's not merge this too quickly — still want to give @bishabosha a chance to weigh in, if he has time.)

bishabosha:
My comment would be to recommend metals first - it is always accurate which is the most important thing for beginners, then IntelliJ next

makingthematrix:
@bishabosha : On the other hand, IntelliJ seems to be more beginner-friendly overall. Besides, if the order is not alphabetical then it will create an issue when we add more IDEs to the list. It will look like very much as if the order was about preference.
 
How about write it explicitely in the introduction that the order is alphabetical and all listed IDEs are treated equally?

SethTisue:
I think we should consider ourselves authorized to make an actual choice here and not fall back on alphabetical order or some such. In general on the website, we must be fair but we don't need to be entirely neutral. It benefits the language to do some gentle steering of users towards the things that we deem most significant.

I don't think I agree with this claim:

>  [Metals] is always accurate which is the most important thing for beginners

First, because I think (with moderate confidence, anyway) that it's just not _all_ that common, especially in beginner code, for IntelliJ to get something wrong. I don't see that possibility as a dominant, or the dominant, consideration.

Second, because a mismatch with the official compiler isn't the only kind of IDE trouble a user is likely to have. My impression is that beginners are _overall_ likelier to have a smoother experience with IntelliJ setup and usage.

Third, for marketing reasons. A great many Scala users still come from the Java world and IntelliJ is the gold standard there. To know that IntelliJ has good Scala support has long been, and remains, critical to adoption. Yes, it is _also_ important to let people know they can use VSCode or their favorite text editor, but having the Metals section right there accomplishes that.

tgodzik:
I mean, there are just two options here, I don't think the user will be unable to see the other and since they are actually on this page they will probably read both rather short entries and decide for themselves.

makingthematrix:
@SethTisue : So what's the next step? 

By the way, I see that the tests are failing with the error that the Chinese translation has dead links. I don't know what I need to do to fix it.

bishabosha (approved):
I'll accept the comment from @SethTisue - no more disagreements

SethTisue:
@makingthematrix it isn't only finding dead links in translated pages, it's also finding them in some English language pages. Does that help identify the cause? I can't dig into it right now but I could eventually if you and/or other reviewers are stumped about what fix is needed.

_overviews/getting-started/scala-ides.md
```diff
@@ -0,0 +1,43 @@
+---
+layout: singlepage-overview
+title: Scala IDEs
+
+partof: scala-ides
+
+permalink: /tutorials/:title.html
```
bishabosha, line 7: this permalink is the source of the HTMLproofer errors - the url is `/tutorials/scala-ides.html` - but it should probably be 
```suggestion
permalink: /getting-started/:title.html
```

SethTisue:
@bishabosha I committed your suggestion, but HTMLProofer is still complaining?

SethTisue:
gah. let's just merge it and deal with HTMLProofer separately

SethTisue:
the new page is located at https://docs.scala-lang.org/getting-started/scala-ides.html

SethTisue:
> As for further edits, I'll submit a followup PR myself once this merged 

https://github.com/scala/docs.scala-lang/pull/3090

PR #3043, opened 2024-07-20
Macro doc: fix example about reporting error

kraktus:
since `quotes.reflect.report` is used, the `given` must be given a variable name


kraktus:
nvm not needed

PR #3046, opened 2024-07-25
Update unified-types.md

shitpoet:
Add phrase:
> Note that `Long` to `Float` conversion is deprecated in new versions of Scala, because of the potential precision lost.


bishabosha (approved):
Thanks! While this is technically correct, perhaps its better to fix the example to another conversion that isn't lossy

som-snytt:
I had to visit the web page to understand the issue. I think bishabosha's previous edit `.toFloat` breaks the text, which is trying to show how easy and wonderful numeric widening is. Instead, it should show the types and also trumpet Scala's "gun buy-back program" for footguns.

SethTisue:
Merging since this is strictly an improvement, but it would be nice if a volunteer improved it further, as per above discussion.

PR #3047, opened 2024-07-25
Update tooling-syntax-rewriting.md

csgn:
I checked scalac command options as follows:

```bash
$ scalac --help
 -javaextdirs  Override java extdirs classpath.
    -language  Enable one or more language features.
  -new-syntax  Require `then` and `do` in control expressions.
   **-no-indent  Require classical {...} syntax, indentation is not significant.**
   -nowarn  Silence all warnings.
```

But in docs wrote as `-noindent`. So I change it to `-no-indent`


bishabosha (approved):
Thanks!

PR #3052, opened 2024-08-08
add Scala FAQ entry about nilary vs nullary methods

SethTisue:
My experience on Discord is that this is in fact asked often enough to merit a spot in the FAQ.

The style guide entry the answer points to could definitely use improvement. I've opened a separate PR on that: https://github.com/scala/docs.scala-lang/pull/3053. But in the meantime, I think it's still okay for the FAQ to link to the old version.

fyi @BalmungSan

PR #3053, opened 2024-08-09
rewrite style guide entry on `foo` vs `foo()`

SethTisue:
Note that the visibility of this will be increased by linking it to from the FAQ: https://github.com/scala/docs.scala-lang/pull/3052

review by @BalmungSan ? @Ichoran? @lrytz?

who knows, we might even nerd-snipe @djspiewak into having a look — I think he wrote the original

### DIFF
--- a/_style/naming-conventions.md
+++ b/_style/naming-conventions.md
@@ -205,7 +205,7 @@ or with no parameter lists at all:
     def foo2 = ...
 
 By convention, parentheses are used to indicate that a method has
-side effects.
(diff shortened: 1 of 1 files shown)


BalmungSan (approved):
Personally, I like the proposal. I don't like the word _"side effects"_ tho, but I really can't think of a better word that is not a full paragraph.

Maybe something like:
> Parameterless methods should not mutate the state of the called object in any way
?

PR #3056, opened 2024-08-12
Add Cask tutorials

gkepka:
The `scala-cli` part needs some attention. While code compiles and starts without any problems, the process end immediately after being started. In SBT the same problem can be solved by setting `fork := true`, but I wasn't able to find any similar solution for `scala-cli`.


sjrd:
> The `scala-cli` part needs some attention. While code compiles and starts without any problems, the process end immediately after being started. In SBT the same problem can be solved by setting `fork := true`, but I wasn't able to find any similar solution for `scala-cli`.

In order to troubleshoot and isolate the problem: have you tried adding a `while (true) {}` loop at the end of `main`?

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -0,0 +1,228 @@
...
+{% endtab %}
+{% tab 'Mill' %}
+Add the Scalatags dependency in `build.cs` file:
+```scala
+ivy"com.lihaoyi::scalatags::0.13.1"
+```
+{% endtab %}
+{% endtabs %}
+
+Now the example above can be rewritten to use a template. Cask will build a response out of the `doctype` automatically.
```
sjrd, line 166: Does this also automatically return the correct Content-Type? The example on the following page says that it does not. If it doesn't, we should already fix it here.

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -0,0 +1,228 @@
...
+
+object MyApp extends cask.MainRoutes {
+
+  private def getZoneIdForCity(city: String): Option[ZoneId] = {
+    import scala.jdk.CollectionConverters._
+    ZoneId.getAvailableZoneIds.asScala.find(_.endsWith("/" + city)).map(ZoneId.of)
+  }
+  
+  @cask.get("/time/:city")
+  def dynamicWithCity(city: String) = {
```
sjrd, line 182: What's the result type, here?

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -0,0 +1,228 @@
...
+
+  private def getZoneIdForCity(city: String): Option[ZoneId] =
+    import scala.jdk.CollectionConverters.*
+    ZoneId.getAvailableZoneIds.asScala.find(_.endsWith("/" + city)).map(ZoneId.of)
+
+  @cask.get("/time/:city")
+  def dynamicWithCity(city: String) =
+    val text = getZoneIdForCity(city) match
+        case Some(zoneId) => s"Current date is: ${ZonedDateTime.now().withZoneSameInstant(zoneId)}"
+        case None => s"Couldn't find time zone for city $city"
```
sjrd, line 216: Indentation issue:
```suggestion
    val text = getZoneIdForCity(city) match
      case Some(zoneId) => s"Current date is: ${ZonedDateTime.now().withZoneSameInstant(zoneId)}"
      case None => s"Couldn't find time zone for city $city"
```

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -0,0 +1,228 @@
...
+Access the endpoint at [http://localhost:8080/time](http://localhost:8080/time). You should see a result similar to the
+one below.
+
+```
+Current date is: 2024-07-22T09:07:05.752534+02:00[Europe/Warsaw]
+```
+
+## Using path segments
+
+You can use path segments to specify the returned data more precisely. Building on the example above, add the `:city`
```
sjrd, line 77: The first sentence doesn't really make sense to me. Could you elaborate?

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -0,0 +1,228 @@
...
+Current date is: 2024-07-22T09:08:33.806259+02:00[Europe/Paris]
+```
+
+and at [http://localhost:8080/time/Tokyo](http://localhost:8080/time/Tokyo) you will see:
+```
+Current date is: 2024-07-22T16:08:41.137563+09:00[Asia/Tokyo]
+```
+
+Cask endpoints can handle either fixed or arbitrary number of path segments. Please consult the 
+[documentation](https://com-lihaoyi.github.io/cask/index.html#variable-routes) for more details.
```
sjrd, line 136: The first sentence here is a bit obscure. Consider leaving only the second sentence.

Also, "Please consult" -> "Consult".

_overviews/toolkit/web-server-query-parameters.md
```diff
@@ -0,0 +1,77 @@
...
+```scala
+import java.time.{ZoneId, ZonedDateTime}
+
+object MyApp extends cask.MainRoutes:
+
+  private def getZoneIdForCity(city: String): Option[ZoneId] =
+    import scala.jdk.CollectionConverters.*
+    ZoneId.getAvailableZoneIds.asScala.find(_.endsWith("/" + city)).map(ZoneId.of)
+
+    @cask.get("/time")
```
sjrd, line 55: There are indentation issues in this snippet.

_overviews/toolkit/web-server-query-parameters.md
```diff
@@ -0,0 +1,77 @@
...
+type: section
+description: Handling query parameters with Cask
+num: 33
+previous-page: web-server-dynamic
+next-page: web-server-input
+---
+
+{% include markdown.html path="_markdown/install-cask.md" %}
+
+You can read the query parameters by adding to an endpoint method arguments that don't have corresponding path segments
```
sjrd, line 12: Consider starting with a sentence that reminds/tells the reader what "query parameters" even are.

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+num: 34
+previous-page: web-server-query-parameters
+next-page: web-server-websockets
+---
+
+{% include markdown.html path="_markdown/install-cask.md" %}
+
+## Handling form-encoded input
+
+Similarly to path segments and query parameters, form fields are read by using endpoint method arguments. Use `cask.postForm`
```
sjrd, line 14:
```suggestion
Similarly to path segments and query parameters, we read form fields using endpoint method arguments. Use `cask.postForm`
```
English does not like the passive voice. Avoid when you can.

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+annotation and set the HTML form method to `post`.
+
+In this example we create a form asking for name and surname of a user and then redirect to a greeting page. Notice the
+use of `cask.Response`. The default returned content type is `text/plain`, set it to `text/html` in order for browser to display
+the form correctly.
```
sjrd, line 19: It feels like we should have introduced this `cask.Response` subtlety before. We're mixing two issues here, it seems.

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+
+The `formMethod` endpoint reads the form data using `name` and `surname` parameters. The names of parameters must
```
sjrd, line 21:
```suggestion
The `formEndpoint` method reads the form data using `name` and `surname` parameters. The names of parameters must
```

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+  initialize()
+```
+{% endtab %}
+{% endtabs %}
+
+## Handling JSON-encoded input
+
+JSON fields are handled in the same way as form fields, except that `cask.PostJson` annotation is used. The topmost fields
+will be read into the endpoint method arguments and if any of them is missing or has an incorrect type, an error message
+will be returned with 400 response code.
```
sjrd, line 91:
```suggestion
will be read into the endpoint method arguments.
```
This could come later in the explanation. It is a bit much for the intro.

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+  --data '{"name":"John","surname":"Smith"}' \
+  http://localhost:8080/json
+```
+
+The response will be:
+```
+Hello John Smith
+```
+
+Deserialization is handled by uPickle JSON library. To deserialize an object, use `ujson.Value` type.
```
sjrd, line 132: I don't understand how this particular example deserializes? AFAICT it precisely *avoids* deserializing and instead presents the raw `ujson.Value`, doesn't it?

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+```
+"{\"name\":\"John\",\"surname\":\"Smith\"}"
+```
+
+## Handling JSON-encoded output
+
+Cask endpoint can return JSON objects returned by uPickle library functions. Cask will automatically handle the `ujson.Value`
+type and set the `Content-Type application/json` header.
+
+In this example we use a simple `TimeData` case class to send information about the time zone and current time in a chosen
```
sjrd, line 179:
```suggestion
In this example we use a `TimeData` case class to send information about the time zone and current time in a chosen
```

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+location. To serialize a case class into JSON you need to define a serializer in its companion object.
```
sjrd, line 180: There's "we" in the previous sentence and "you" in this one. We should be consistent throughout.

_overviews/toolkit/web-server-websockets.md
```diff
@@ -0,0 +1,113 @@
...
+</html>
+```
+
+The JavaScript code opens a WebSocket connection using the `ws://localhost:8080/websocket` endpoint. The `ws.onmessage`
+event handler is executed when server pushes a message to the browser and `ws.onclose` when the connection is closed. 
+
+Create an endpoint for serving static files using `@cask.staticResources` annotation and the endpoint for handling
+the WebSocket connection.
+
+{% tabs web-server-query-1 class=tabs-scala-version %}
```
sjrd, line 60:
```suggestion
{% tabs web-server-websocket-1 class=tabs-scala-version %}
```

_overviews/toolkit/OrderedListOfMdFiles
```diff
@@ -27,3 +27,10 @@ http-client-request-body.md
 http-client-json.md
 http-client-upload-file.md
 http-client-what-else.md
+web-server-intro.md
+web-server-static.md
+web-server-dynamic.md
+web-server-query-parameters.md
+web-server-input.md
+web-server-websockets.md
+web-server-cookies-and-decorators.md
```
sjrd, line 36: Missing EOL.

_overviews/toolkit/web-server-input.md
```diff
@@ -0,0 +1,232 @@
...
+
+{% tabs web-server-input-3 class=tabs-scala-version %}
```
sjrd, line 182:
```suggestion
{% tabs web-server-input-4 class=tabs-scala-version %}
```

_overviews/toolkit/web-server-static.md
```diff
@@ -0,0 +1,146 @@
+---
+title: How to serve a static file?
```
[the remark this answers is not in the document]
gkepka: I would like to preserve the "How to [do X]" title convention, used in all the other tutorials. I imagine that's what a person in a need for guidance would search for.

_overviews/toolkit/web-server-dynamic.md
gkepka: It does in case of using a Scalatags template. I added a sentence explaining that.

_overviews/toolkit/web-server-dynamic.md
gkepka: I rewrote it to something hopefully more understandable.

gkepka:
The problem with `scala-cli` turned out to be caused by the difference between how  files with `.sc` and `.scala` extensions are handled. It turns out that while the `.scala` files are executed as written, the contents of `.sc` files are wrapped in an auto-generated `main` function, ignoring all `main` functions defined in the code. The documentation for `scala-cli` makes a very good job at hiding that distinction, I managed to find it descibed [here](https://scala-cli.virtuslab.org/docs/reference/commands/#run).

_overviews/toolkit/web-server-input.md
gkepka: I wonder if I should add a part dedicated specifically to this.

_overviews/toolkit/web-server-dynamic.md
```diff
@@ -65,17 +71,17 @@ In the terminal, the following command will start the server:
 {% endtab %}
 {% endtabs %}
 
-Access the endpoint at [http://localhost:8080/time](http://localhost:8080/time). You should see a result similar to the
-one below.
+Access [the endpoint](http://localhost:8080/time). You should see a result similar to the one below.
```
sjrd, line 74: IMO the previous text, with the link appearing in the text, was better here. We should actively show the user that we're sending them to a `localhost` URL.

sjrd (commented):
This looks much better now, IMO!

There are a number of English grammar issues that we should fix. Let's discuss offline how you prefer me to report them.

_overviews/toolkit/web-server-static.md
```diff
@@ -0,0 +1,159 @@
...
+  initialize()
+```
+{% endtab %}
+{% endtabs %}
+
+In the example above, `@cask.staticFiles` instructs the server to look for files accessed at the `/static` path in the 
+`src/main/resources` directory. Cask will match any subpath coming after `/static` and append it to the directory path.
+If you access the `/static/hello.html` file, it will serve the file available at `src/main/resources/hello.html`.
+The directory path can be any path available to the server, relative or not. If the requested file cannot be found in the
+specified location, a 404 response with an error message will be returned instead.
```
sjrd, line 96:
```suggestion
specified location, the server will return a 404 response with an error message.
```

_overviews/toolkit/web-server-static.md
```diff
@@ -0,0 +1,159 @@
...
+
+The `Example` object inherits from the `cask.MainRoutes` class, providing the main function that starts the server. The `initialize()`
```
sjrd, line 98:
```suggestion
The `Example` object inherits from the `cask.MainRoutes` class. It provides the main function that starts the server. The `initialize()`
```

_overviews/toolkit/web-server-static.md
```diff
@@ -0,0 +1,159 @@
...
+  @cask.staticResources("/static")
+  def staticEndpoint(): String = "."
+
+  initialize()
+```
+{% endtab %}
+{% endtabs %}
+
+In the endpoint method, the location is set to `"."`, telling the server that the files are available directly in the
+resources directory. In general, you can use any nested location within the resources directory, for instance you could opt
```
sjrd, line 130:
```suggestion
resources directory. In general, you can use any nested location within the resources directory. For instance, you could opt
```

_overviews/toolkit/web-server-websockets.md
```diff
@@ -0,0 +1,118 @@
...
+num: 35
+previous-page: web-server-input
+next-page: web-server-cookies-and-decorators
+---
+
+{% include markdown.html path="_markdown/install-cask.md" %}
+
+You can create a WebSocket endpoint by using the `@cask.websocket` annotation. The endpoint method can return either a
+`cask.WsHandler` instance defining how the communication should take place, or a `cask.Response`, which rejects the 
+attempt at forming a WebSocket connection.
```
sjrd, line 14:
```suggestion
You can create a WebSocket endpoint with the `@cask.websocket` annotation. The endpoint method should return a
`cask.WsHandler` instance defining how the communication should take place. It can also return a `cask.Response`, which rejects the 
attempt at forming a WebSocket connection.
```

_overviews/toolkit/web-server-websockets.md
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+initialize()
+```
+{% endtab %}
+{% endtabs %}
+
+In the `cask.WsHandler` we define a `cask.WsActor` which reacts to events (of `cask.util.Ws.Event` type) and uses
+WebSocket channel to send messages. In this example, we receive the name of a city and return the current time there. If server
+receives an empty message, the connection is closed.
```
sjrd, line 118:
```suggestion
In the `cask.WsHandler` we define a `cask.WsActor`. It reacts to events (of type `cask.util.Ws.Event`) and uses the
WebSocket channel to send messages. In this example, we receive the name of a city and return the current time there. If the server
receives an empty message, the connection is closed.
```

sjrd (changes requested):
Final English grammar and style comments.

sjrd (approved):

PR #3068, opened 2024-09-06
FAQ: minor improvements to functions vs methods

SethTisue:
review by @BalmungSan ?

my initial motivation was to fix a place where Scala 3 was referred to in the future tense. but then while I was there I found I wanted to do some light rewriting

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -253,10 +253,13 @@ accept given parameters and
 [polymorphic functions]({{ site.scala3ref }}/new-types/polymorphic-function-types.html)
 have type parameters.
 
-Most code uses methods most of the time,
-unless a function value is actually needed. With
-[eta-expansion](https://stackoverflow.com/questions/39445018/what-is-the-eta-expansion-in-scala),
-a method can be converted as a function when needed.
+It's standard to use methods most of the time,
+except when function value is actually needed.
+[Eta-expansion](https://stackoverflow.com/questions/39445018/what-is-the-eta-expansion-in-scala),
+converts methods to functions when needed.
+For example, a method such as `map` expects a function,
+but even if you `def square` as shown above, you can
+still `xs.map(square)`.
 
 ### What's the difference between types and classes?
 
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -244,18 +244,19 @@ differ from a function value such as:
 
     val square: Int => Int = x => x * x
 
-For Scala 2, there is a [complete answer on Stack Overflow](https://stackoverflow.com/a/2530007/4111404)
+For **Scala 2**, there is a [complete answer on Stack Overflow](https://stackoverflow.com/a/2530007/4111404)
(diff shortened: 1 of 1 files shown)


_overviews/FAQ/index.md
```diff
@@ -244,18 +244,19 @@ differ from a function value such as:
...
+[Context functions]({{ site.scala3ref }}/contextual/context-functions.html)
+accept given parameters and
+[polymorphic functions]({{ site.scala3ref }}/new-types/polymorphic-function-types.html)
+have type parameters.
 
-Nevertheless, it is still recommended to use methods most of the time,
-unless you absolutely need a function. And, thanks to
-[eta-expansion](https://stackoverflow.com/questions/39445018/what-is-the-eta-expansion-in-scala)
-you rarely would need to define a function rather than a method.
+Most code uses methods most of the time,
```
BalmungSan, line 256: IMHO we should not only note that most code uses methods but also encourage them as the default, given the syntax, performance, etc.

_overviews/FAQ/index.md
```diff
@@ -244,18 +244,19 @@ differ from a function value such as:
...
+unless a function value is actually needed. With
```
BalmungSan, line 257: In retrospect, we should have added an example, something like:

> e.g. you will compose it with other functions using `andThen`.

_overviews/FAQ/index.md
```diff
@@ -244,18 +244,19 @@ differ from a function value such as:
...
+[eta-expansion](https://stackoverflow.com/questions/39445018/what-is-the-eta-expansion-in-scala),
```
BalmungSan, line 258: I'm not sure if _"With"_ is the best connector to use.
But, I am not known for my good level of written English 😅 it just feels off to me but that may as well be because I am not a native speaker.

BalmungSan (approved):

_overviews/FAQ/index.md
SethTisue: Hmm. I see what you mean, but my intent was actually to make the wording stronger, not to weaken it. I think the word "recommend" is sort of weak. Whereas I really want to say quite forcefully, methods are definitely the norm in this language.

Maybe I can find a different wording.

_overviews/FAQ/index.md
SethTisue: yes, good point, I can try to add something like hat

_overviews/FAQ/index.md
SethTisue: I actually hesitated over that too, you're right to call it out.

Thanks for the feedback; I'll ponder some further edits.

SethTisue:
@BalmungSan wdyt now? mergeable?

note that I really don't want to make this any longer. FAQ answers should be short. so I'm willing to make further edits but only if they aren't expansions. the entry already links to longer explanations.

BalmungSan (approved):

SethTisue:
fixup: cee9851ba1edaf550833031786593bbd0efe793d

PR #3082, opened 2024-10-06
add JDK 23 to build tool table

SethTisue:
@eed3si9n what should I put for sbt?

@lefou what should I put for mill?

@Gedochao should I add a scala-cli column?


eed3si9n:
We only run CI tests on JDK 17 (https://www.scala-sbt.org/1.x/docs/Setup.html). If anything later worked, it's purely by lucky accident, but people are free to choose whatever.

Friendseeker:
@eed3si9n 

For sbt, would it be an improvement to add CI tests for all combination of major OS (Ubuntu, Mac, Windows) & LTS Java version?

I use sbt with Java 21 locally and can confirm it works quite well. Some minor change in `ci.yml` to declare a testing matrix would do it (which I can work on).

Even if some of the newly added tests fail, given the massive userbase of sbt, the CI failures would likely be real issue that needs to be addressed.

eed3si9n:
@Friendseeker

> For sbt, would it be an improvement to add CI tests for all combination of major OS (Ubuntu, Mac, Windows) & LTS Java version?

I've created https://github.com/sbt/sbt/issues/7731. Let's discuss there.

SethTisue:
for the purposes of this web page, it's enough that there isn't some obvious showstopper that most people will hit

SethTisue:
@lefou what should I put for mill?

lefou:
Mill has no dedicated Java 23 test suite. Our full CI doesn't succeed on Java 23 due to some tests exercising older Scala versions that are knows to be incompatible with Java 21 and above. But Mill `0.11.12` runs with Java 23 without known issues. I guess older versions down to `0.11.5` might work as well.

lefou:
I just checked, I can build Mill's own code base (`0.11.x` branch) with Mill `0.11.8` on Java 23 too. I couldn't check older versions due to feature requirement in that project.

_overviews/jdk-compatibility/overview.md
```diff
@@ -35,6 +35,7 @@ Minimum working versions:
 
 | JDK         | sbt             | mill       |
 |:-----------:|:---------------:|:-----------|
+| 23          | 1.9.0           | ???        |
```
lefou, line 38:
```suggestion
| 23          | 1.9.0           | 0.11.8     |
```

SethTisue:
thank you @lefou 

leaving as draft until we hear from Piotr (there's no big rush)

SethTisue:
ping @Gedochao

Gedochao:
Sorry for the late answer, I missed this somehow.

> should I add a scala-cli column?

@SethTisue that would make sense, but I suspect you'd need the relevant minimum `scala-cli` versions for each of the LTS JDKs to fill the table.

`scala-cli` is built with Java 17, and running the JVM launcher (a power user case) technically requires Java 16+ (with 17+ being recommended).
Our tests suite also runs mostly on Java 17, with some choice tests running on 8, 11 and 22.

The average user would use a native launcher, which wouldn't care what version of Java is installed for running the CLI itself.

When building projects with a build server (as in, Bloop), Java 17+ is required. However, Scala CLI will download the appropriate JVM for Bloop as necessary, so this could be treated independently.

The most recent `scala-cli` version seems to run fine with Java 23, but to provide more precise data for the table (which specific `scala-cli` version is needed), I'd need to run some tests. I'll try to give more specific data when I can, I should be able to give you some answers this week, if not today.

If this is blocking this PR, we can do a separate follow-up.

Gedochao:
Ran some very basic tests.
Minimum Scala CLI versions for running user projects with the appropriate JDKs:
- Java 17: Scala CLI 1.0.0 (theoretically any pre-1.0.0 would work as well, even as early as 0.0.2) but I'd rather not put any of the earlier versions there)
- Java 21: Scala CLI 1.0.0 with Scala set to 3.3.1+, or 1.0.5 with tool defaults
- Java 23: Scala CLI 1.4.1 or higher (independently of Scala version)

I intend to add some tests to our test suite, + perhaps bump things from Java 22 to 23 on the CI, but perhaps the data above would be sufficient?

SethTisue:
> perhaps the data above would be sufficient?

Indeed, thank you!

SethTisue:
I'll (not immediately) do a follow up that adds JDK 24.

SethTisue:
followup adding JDK 24: https://github.com/scala/docs.scala-lang/pull/3140

PR #3083, opened 2024-10-14
JDK compat page: add Scala 3 LTS column

SethTisue:
realized recently that was a significant omission we ought to address

@Gedochao @sjrd need your help here — I left two cells blank because I don't know what to put, and I'm not sure some other entries are correct (see below)


SethTisue:
in addition to the blank cells, I'm also wondering if the Scala 3 LTS column values are actually correct. I no longer remember what the basis for the "3.3.4" and "3.3.5" entries was, and I now find them questionable

JDK 21 support landed in https://github.com/scala/scala3/pull/17536, in scala/scala3@5038c9c4401e9922e0d5fa11369981f722be0d9c which is in 3.3.1's history but not 3.3.0's , so I believe that 3.3.1 entry is correct for JDK 22

but what about 22 and 23? did we actually do anything to support those versions?

I suspect the "3.3.4" and "3.3.5" entries might have come about because those are the versions were we upgraded ASM. in Scala 2, it makes sense for the ASM upgrade PRs to be the vital ones, because the Scala 2 optimizer uses ASM to read bytecode

but Scala 3 doesn't have the optimizer. does it ever use ASM for reading bytecode, or does it only ever use it to generate bytecode? if the latter, than maybe 3.3.1 actually works fine on JDK 22 and 23 and the table ought to indicate that?

cc @Philippus, who has taken an interest in ASM upgrades

SethTisue:
in addition to the blank cells, I'm also wondering if the Scala 3 LTS column values are actually correct. I no longer remember what the basis for the "3.3.4" and "3.3.5" entries was

JDK 21 support landed in https://github.com/scala/scala3/pull/17536, in scala/scala3@5038c9c4401e9922e0d5fa11369981f722be0d9c which is in 3.3.1's history but not 3.3.0's , so I believe that 3.3.1 entry is correct for JDK 21

but what about 22 and 23? did we actually do anything to support those versions?

I suspect the "3.3.4" and "3.3.5" entries might have come about because those are the versions were we upgraded ASM. in Scala 2, it makes sense for the ASM upgrade PRs to be the vital ones, because the Scala 2 optimizer uses ASM to read bytecode

but Scala 3 doesn't have the optimizer. does it ever use ASM for reading bytecode, or does it only ever use ASM to generate bytecode? if the latter, than maybe 3.3.1 actually works fine on JDK 22 and 23 and the table ought to indicate that?

I see that https://www.scala-lang.org/news/3.3.4 does say "Support for JDK 22", but there's no link that goes with it. what is the basis for that claim? I would like to understand

cc @Philippus, who has taken an interest in ASM upgrades

cc @bishabosha who is responsible for https://github.com/scala/scala3/pull/18618 and might have some insight

bishabosha:
You are right that 3.3.0 didn't work with JDK 21, it crashed in a weird was due to the constant pool change - so in 3.3.1 we introduced more explicit reporting of breaking due to failures in reading class files

SethTisue:
@Gedochao @sjrd need your help here

if it feels like I'm asking for more work than is reasonable, perhaps we could find a low-effort solution, like filling in some cells to the best of our recollection and figure that if a mistake is ever found, we can fix it then?

sjrd:
From my tests, Scala 3.3.1 and 3.4.0 both support JDKs up to 23. I did not manage to make them crash, unlike Scala 3.3.0 which definitely fails with JDK 21+.

SethTisue:
Piotr has his hands full right now, so I'm just going to merge this – I'll run it by him at a more opportune time later

PR #3088, opened 2024-10-19
add FAQ about IDEs

SethTisue:
linking to the new page, written by @makingthematrix , that landed recently


SethTisue:
I'm just going to hit "merge" on this as it seems like a very straightforward change.

PR #3090, opened 2024-10-19
on IDEs page, give non-VS-Code editors their own section

SethTisue:
I also did some copyediting and rewording of this and that

review by @makingthematrix? (and by anyone else interested)


bishabosha (approved):
I think this is a good improvement

PR #1691, opened 2020-05-11
complete code snippet

honnix:
* Use `whitebox.Context` explicitly
* Show a complete code snippet (it was not obvious to me as a new macro developer how to replace those `???`)

Still I couldn't make the example work under 2.12.11 with scala-maven-plugin:

```
      <plugin>
        <groupId>net.alchim31.maven</groupId>
        <artifactId>scala-maven-plugin</artifactId>
        <configuration>
          <args>
            <arg>-language:experimental.macros</arg>
            <arg>-feature</arg>
          </args>
          <compilerPlugins>
            <compilerPlugin>
              <groupId>org.scalamacros</groupId>
              <artifactId>paradise_2.12.11</artifactId>
              <version>2.1.1</version>
            </compilerPlugin>
          </compilerPlugins>
        </configuration>
        <executions>
          <execution>
            <goals>
              <goal>compile</goal>
              <goal>testCompile</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
```

PR #3094, opened 2024-10-22
language tour: remove old 'automatic closures' page

SethTisue:
fixes #1691


Flowdalic:
FYI, https://docs.scala-lang.org/tour/automatic-closures.html still exists and is accessible. I am not sure if this should be the case.

SethTisue:
@fsalvi is this one of those things where the website publishing only adds files but doesn't delete them?

do you have any clue why the `redirect_from` in this PR isn't doing its job...?

PR #3105, opened 2024-11-18
`boundary`/`break` expression plus several examples of custom control structures

gkepka:


_overviews/scala3-book/control-structures.md
```diff
@@ -989,4 +990,178 @@ finally
...
+        if i == 7 && j == 9 then break("found")
+    "not found"
+
+println(result)
+```
+
+{% endtab %}
+{% endtabs %}
+
+Since `break` uses an exception to control the execution flow, mixing it with `try`/`catch` can lead to unexpected results:
```
gkepka, line 1017: This one looks like a potential huge footgun if someone doesn't notice `break` throws an exception, so I wanted to mention it. Please notify if the `try`/`catch` example is OK and should stay.

_overviews/scala3-book/control-structures.md
```diff
@@ -989,4 +990,178 @@ finally
...
+  println(i)
+  i = i + 1
+  until(i == 10)
+```
+
+{% endtab %}
+{% endtabs %}
+
+The following example uses a context parameter to store actions to be executed after a block of code is completed. The
+`with_defer` function defines a scope in which `defer` can be used to defer execution of actions.
```
gkepka, line 1124: Not sure about this because context parameters are not mentioned yet at this point. Maybe the custom control structures section should be moved to a separate page?

_overviews/scala3-book/control-structures.md
Ichoran: `Try` also swallows the control flow because internally it uses `try`/`catch` in a non-`boundary`/`break` compatible way.  It might also be worth explaining how to avoid catching it.  (E.g. `case e: Exception if !e.isInstanceOf[Break[?]]`.)

PR #3118, opened 2024-11-24
Add documentation for Pattern Matching on `String`

### LINKED ISSUE #2687: String interpolation doc page should cover pattern matching
I was surprised today to notice that https://docs.scala-lang.org/overviews/core/string-interpolation.html makes no reference to the use of string interpolation in patterns, e.g. with the `s` interpolator:

```scala
scala 2.13.10> "foo bar" match { case s"f${x}r" => x }
               ^
               warning: match may not be exhaustive.
val res1: String = oo ba
```

there is a bit of material on this https://docs.scala-lang.org/sips/string-interpolation.html but not very much (it only references a theoretical `xml` interpolator — understandable since the SIP predates any use of the feature in the Scala stdlib)

Friendseeker:
@SethTisue @sjrd 

c.c. https://discord.com/channels/632150470000902164/632628489719382036/1286949507128361093

Closes #2687

Btw I didn't build the site locally to test out the changes (I don't use Ruby so lots of hassle to setup things from scratch), but if necessary, I can do local testing.

Feel free to force push to my branch for any obvious improvements (i.e. syntax error, typo etc).

### DIFF
--- a/_tour/pattern-matching.md
+++ b/_tour/pattern-matching.md
@@ -138,7 +138,7 @@ println(showNotification(someVoiceRecording))  // prints You received a Voice Re
 
 The function `showNotification` takes as a parameter the abstract type `Notification` and matches on the type of `Notification` (i.e. it figures out whether it's an `Email`, `SMS`, or `VoiceRecording`). In the `case Email(sender, title, _)` the fields `sender` and `title` are used in the return value but the `body` field is ignored with `_`.
 
-## Matching on String
+## Matching on string
 
 The `s`-interpolator allows embedding variables in strings and is also useful for pattern matching.
 
--- a/_tour/pattern-matching.md
+++ b/_tour/pattern-matching.md
@@ -168,6 +168,39 @@ input match
 
 In this example, name and age extract parts of the string based on the pattern. This is helpful for parsing structured text.
 
+We can also use extractor objects for string pattern matching.
+
+{% tabs s-interpolator-pattern-matching-2 class=tabs-scala-version %}
+{% tab 'Scala 2' for=s-interpolator-pattern-matching-2 %}
+```scala
+object Int {
+  def unapply(s: String): Option[Int] = s.toIntOption
+}
+
+val input: String = "Alice is 25 years old"
+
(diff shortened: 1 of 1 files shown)


scarf005 (approved):
Looks good to be, tho maybe we could also document that extractor object is also supported: https://www.scala-lang.org/files/archive/api/2.13.x/scala/StringContext$s$.html#unapplySeq(s:String):Option[Seq[String]]

such that following works (and great for parsing)
```scala
scala> object Int:
     |   def unapply(s: String): Option[Int] = try Some(s.toInt)
     |   catch case _: NumberFormatException => None
     | 
// defined object Int
                                                                                                                 
scala> val input: String = "Alice is 25 years old"
val input: String = Alice is 25 years old
                                                                                                                 
scala> val (name, age) = input match
     |   case s"$name is ${Int(age)} years old" => (name, age)
     |   case _                                 => ("No match", -1)
     | 
val name: String = Alice
val age: Int = 25
```

_tour/pattern-matching.md
```diff
@@ -138,6 +138,69 @@ println(showNotification(someVoiceRecording))  // prints You received a Voice Re
...
+In this example, name and age extract parts of the string based on the pattern. This is helpful for parsing structured text.
+
+We can also use extractor objects for string pattern matching.
+
+{% tabs s-interpolator-pattern-matching-2 class=tabs-scala-version %}
+{% tab 'Scala 2' for=s-interpolator-pattern-matching-2 %}
+```scala
+object Int {
+  def unapply(s: String): Option[Int] = s.toIntOption
+}
```
adpi2, line 178: It seems confusing to use `Int` as an object name, because it shadows the `scala.Int` type. I suggest to use `Age` instead.

```suggestion
object Age {
  def unapply(s: String): Option[Int] = s.toIntOption
}
```

Friendseeker:
@gkepka Would you like to review this PR?

adpi2 (approved):
LGTM

SethTisue:
Thank you — a nice addition

PR #3126, opened 2024-12-10
Scala 2 Book migration - Control Structures

gkepka:


sjrd (changes requested):

_overviews/scala3-book/control-structures.md
```diff
@@ -1040,15 +1040,27 @@ match
...
 {% endtabs %}
 
+The `match` expression can also follow a period, which simplifies matching on results returned by chained method calls:
+
+{% tabs 'control-structures-33' %}
+{% tab 'Scala 3 Only' %}
+```scala
+List(1, 2, 3).map(_ * 2).headOption.match
+  case Some(value) => println(s"The head is: $value")
+  case None => println("The list is empty")
```
sjrd, line 1050: For the example to be compelling, it should be formatted as follows:
```suggestion
List(1, 2, 3)
  .map(_ * 2)
  .headOption
  .match
    case Some(value) => println(s"The head is: $value")
    case None => println("The list is empty")
```

PR #3130, opened 2024-12-13
Update pattern-matching.md

ErikPohl444:
Added a note about sealed traits because it is the first usage of them in the guide.

### DIFF
--- a/_tour/pattern-matching.md
+++ b/_tour/pattern-matching.md
@@ -92,7 +92,7 @@ case class VoiceRecording(contactName: String, link: String) extends Notificatio
 {% endtab %}
 {% endtabs %}
 
-`Notification` is a sealed trait which has three concrete Notification types implemented with case classes `Email`, `SMS`, and `VoiceRecording`. Now we can do pattern matching on these case classes:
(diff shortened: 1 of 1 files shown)


_tour/pattern-matching.md
```diff
@@ -92,7 +92,7 @@ case class VoiceRecording(contactName: String, link: String) extends Notificatio
 {% endtab %}
 {% endtabs %}
 
-`Notification` is a sealed trait which has three concrete Notification types implemented with case classes `Email`, `SMS`, and `VoiceRecording`. Now we can do pattern matching on these case classes:
+`Notification` is a sealed trait which has three concrete Notification types implemented with case classes `Email`, `SMS`, and `VoiceRecording`. (A [sealed trait](#Sealed-types) can be extended only in the same file as its declaration.) Now we can do pattern matching on these case classes:
```
gkepka, line 95:
```suggestion
`Notification` is a sealed trait which has three concrete Notification types implemented with case classes `Email`, `SMS`, and `VoiceRecording`. (A [sealed trait](/tour/pattern-matching.html#sealed-types) can be extended only in the same file as its declaration.) Now we can do pattern matching on these case classes:
```
Hopefully will help with the failing check.

gkepka (changes requested):

PR #3133, opened 2024-12-13
Update extractor-objects.md

ErikPohl444:
The customerid match case does not run in Scastie for Scala 3 without curly braces.

I've added them here.

Is this a problem with Scastie's Scala 3? If so, you can reject the PR and fix Scastie.

Is this a fix for the tutorial code for Scala 3?  If so, please accept this PR.

### DIFF
--- a/_tour/extractor-objects.md
+++ b/_tour/extractor-objects.md
@@ -49,9 +49,10 @@ object CustomerID:
     if stringArray.tail.nonEmpty then Some(stringArray.head) else None
 
 val customer1ID = CustomerID("Sukyoung")  // Sukyoung--23098234908
-customer1ID match
+customer1ID match {
   case CustomerID(name) => println(name)  // prints Sukyoung
   case _ => println("Could not extract a CustomerID")
+}
 ```
 {% endtab %}
 


Philippus:
I think this might be a bug? It works if I wrap it in something, like in a `def`:
``` 
(...) 
def matcher() =
    customer1ID match
      case CustomerID(name) => println(name) // prints Sukyoung
      case _ => println("Could not extract a CustomerID")
matcher()
```

ErikPohl444:
Do you think it is a bug in the documentation [my PR] or in Scastie?  

The matcher() def works for me, as well as curly braces.

If it is a Scastie bug, then I request someone closer to Scastie's to enter an issue for it relating to this.
If it is something which would benefit from a change in the tour, please let me know if you'd rather curly braces or a def statement.

Right now, the tour does not work in Scastie.

At the very least, I could submit a disclaimer in the tour for the code not working in Scastie for Scala 3.

Let me know your preference!

Philippus:
I think it might be a case of https://github.com/scalacenter/scastie/issues/929.

ErikPohl444:
Good catch.  I know as a newbie I spent a while on stack overflow trying to find out the proper syntax for this code to work in Scala 3, and finally just using curly braces.  It is a Scastie issue, and Scastie is recommended for people using the tour.  

How about I remove the curly braces from my pull request in the Scala 3 code blocks I've committed changes for, and I add a disclaimer that -- for people following along in Scastie -- they'll need to use the Scala 2 code block.

Does that sound fair to you?  That way the Scala 3 code block will maintain its no-braces status, and people like me wandering through the tour with Scastie will not wonder why the code doesn't work for us.

adpi2:
Thanks @ErikPohl444 for reporting this issue and proposing a mitigation plan.

This seems to be a important bug in Scastie and we should put some effort in fixing it.

> Does that sound fair to you? That way the Scala 3 code block will maintain its no-braces status, and people like me wandering through the tour with Scastie will not wonder why the code doesn't work for us.

This is appears to be an important bug in Scastie, and we should prioritize addressing it. Let's see if we manage to fix it before adding any disclaimer.

ErikPohl444:
Addressing the root cause is ideal.  Thanks!

PR #3134, opened 2024-12-21
clarify wording on pure-functions overview to allow immutable hidden state, and add an example

epic-64:
Updates the "pure functions" overview at https://docs.scala-lang.org/scala3/book/fp-pure-functions.html

- adds clarification accessing immutable hidden state is still considered pure
- adds an example of this in form of a closure

### DIFF
--- a/_overviews/scala3-book/fp-pure-functions.md
+++ b/_overviews/scala3-book/fp-pure-functions.md
@@ -13,7 +13,7 @@ Another feature that Scala offers to help you write functional code is the abili
 A _pure function_ can be defined like this:
 
 - A function `f` is pure if, given the same input `x`, it always returns the same output `f(x)`
-- The function’s output depends _only_ on its input variables and its implementation
+- The function’s output depends _only_ on its input variables and its implementation (and in case of a closure, any immutable data that it captures)
 - It only computes the output and does not modify the world around it
 
 This implies:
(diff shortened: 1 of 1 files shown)


epic-64:
there are still some typos, sorry, I will reopen a PR once they are fixed

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -57,7 +57,7 @@ Conversely, the following functions are _impure_ because they violate the defini
 
 Impure functions often do one or more of these things:
 
-- Read from hidden state, i.e., they access variables and data not explicitly passed into the function as input parameters
+- Read from hidden mutable state, i.e., they access non-constant data that was not explicitly passed into the function as input parameters
```
SethTisue, line 60: "non-constant" is not really a synonym for "mutable". I would suggest using the word "mutable" again even if it feels a bit repetitive; it's more important to be precise.

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -98,6 +98,21 @@ def double(i: Int): Int = i * 2
 
 {% endtabs %}
 
+The next example is bit more tricky. Here, `i` is not passed as a parameter, but instead referenced directly from the outside.
+This works in Scala because functions act as closures - they can capture the state around them. As long as that state is *immutable*, such a closure is still considered pure.
```
SethTisue, line 102: Not sure this is the best wording since we usually use "state" to talk about mutable state. Not all of the time. I would suggest "capture values from enclosing scopes".

_overviews/scala3-book/fp-pure-functions.md
```diff
@@ -129,7 +144,7 @@ If you understand that code, you’ll see that it meets the pure function defini
 
 The first key point of this section is the definition of a pure function:
 
-> A _pure function_ is a function that depends only on its declared inputs and its implementation to produce its output.
+> A _pure function_ is a function that depends only on its declared inputs, captured constants, and its implementation to produce its output.
```
SethTisue, line 147: suggest "closed-over values" rather than "captured constants"

PR #3135, opened 2024-12-25
Change the performance of the Queue prepend operation

pochka15:
If I'm not mistaken the `prepend` performance is incorrect. It should be constant instead of linear. The same performance is written in the russian lang docs ([Link](https://docs.scala-lang.org/ru/overviews/collections-2.13/performance-characteristics.html)).

I've also looked into sources and `prepended` seems to be a simple O(1) operation

001691@2x


adpi2 (approved):
I am not sure why it was reverted in #1880.

But it seems you are right and `C` is the correct performance of `Queue.prepended`.

PR #3139, opened 2025-01-04
Fix: Rename `coll` to `iter` for consistency in lazyMap examples

blank71:
https://docs.scala-lang.org/overviews/collections-2.13/views.html

~~fix(docs): Rename variable `iter` to `coll` for consistency with documentation~~

~~Replaced variable name `iter` with `coll` in the code examples to align with the surrounding documentation and improve clarity.~~


blank71:
I'm new to scala and I found this. 

But I found the problem above. Maybe I need to fix document except code. Fix `coll` to `iter`. At least I think it should avoid using `coll`.

```bash
% scala-cli version
Scala CLI version: 1.5.1
Scala version (default): 3.5.1
Your Scala CLI version is outdated. The newest version is 1.5.4
It is recommended that you update Scala CLI through the same tool or method you used for its initial installation for avoiding the creation of outdated duplicates.
```

```scala
scala> def lazyMap[T, U](coll: Iterable[T], f: T => U) = new Iterable[U]:
     |   def iterator = coll.iterator map f
     |
-- [E049] Reference Error: -----------------------------------------------------
2 |  def iterator = coll.iterator map f
  |                 ^^^^
  |Reference to coll is ambiguous.
  |It is both defined in method lazyMap
  |and inherited subsequently in anonymous class Object with Iterable[U] {...}
  |
  | longer explanation available when compiling with `-explain`
-- [E007] Type Mismatch Error: -------------------------------------------------
2 |  def iterator = coll.iterator map f
  |                                   ^
  |                                   Found:    (f : T => U)
  |                                   Required: U => U
  |
  | longer explanation available when compiling with `-explain`
2 errors found

scala> def lazyMap[T, U](iter: Iterable[T], f: T => U) = new Iterable[U]:
     |   def iterator = iter.iterator map f
     |
def lazyMap[T, U](iter: Iterable[T], f: T => U): Iterable[U]
```

SethTisue:
What's wrong with `coll`?

I can see the desire to shorten `Iterable` to `iter`, but personally I feel like `iter` is so strongly associated with `Iterator` that `coll` might actually be better here. There isn't any `Collection` type in the collections API, but `Seq` and `Set` and `Map` are all subtypes of `Iterable`.

blank71:
Thanks!!!

The first problem is that the code uses `iter` while the description uses `coll`, causing a discrepancy. This is I want to fix.

Some try, I changed the code from `iter` to `coll`. The old documentation used `coll`. Found the problem as described above. It reads like the problem is caused by an ambiguity between the `coll` argument and the `coll` that `Iterable` has.

I check the implementation of Iterable in `Iterable.scala` file. 

```scala
trait Iterable[+A] extends IterableOnce[A]
  with IterableOps[A, Iterable, Iterable[A]]
  with IterableFactoryDefaults[A, Iterable] {
  ...
  final protected def coll: this.type = this
  ...
}
```

All of this leads me to think that the use of `coll` should be avoided. Please let me know if there is another variable name that would work best instead of `iter`. Or is it possible to avoid the problem while keeping the use of `coll`?

SethTisue:
Ah, I see. Okay, let's go with `iter`. Thanks for your patience here.

blank71:
Thanks a lot for checking and merging!

PR #3141, opened 2025-01-07
update scalatest version

keyno63:
The main issue lies with the Japanese page.
The version specification for ScalaTest was incorrect, so I updated it.
However, sbt fails to resolve ScalaTest, resulting in an error.

I updated the version to the latest stable one (3.1.19).

I also updated the versions on other language pages as well.
This was likely not necessary, but I changed them for consistency, as having different versions across pages was a concern.


adpi2 (approved):
LGTM

PR #3142, opened 2025-01-07
fix add-guide file link in contribute

keyno63:
It seems that the link to contribute is incorrect.


keyno63:
It seems I misunderstood.
When the file is converted to HTML, it appears to be accessible using a relative path.
I updated it to a URL format to match the others.

SethTisue:
Thanks!

PR #3147, opened 2025-01-22
string-interpolation: fix headline level

Flowdalic:
Previously, the raw-interpolator was at the third headline level (###), while the s-interpolator and the f-interpolator where at the second headline level (##).

This changes and renames the second headline "Other Interpolators" to become a second level headline, under which we place all three string interpolators.


Philippus (approved):

Philippus:
Thank you!

PR #3148, opened 2025-01-22
string-interpolation: the raw-interpolator also performs variable substitution

Flowdalic:
Clarify that the raw-interpolator also performs variable substitution. This was not obvious from the text and it is easy to believe that raw means just that: the raw string, as is, without any substitutions performed.

This also moves the last sentence of the raw-interpolator section into the following section, where it logicaly belongs.

### DIFF
--- a/_overviews/scala3-book/string-interpolation.md
+++ b/_overviews/scala3-book/string-interpolation.md
@@ -211,9 +211,21 @@ res1: String = a\nb
 
 The raw interpolator is useful when you want to avoid having expressions like `\n` turn into a return character.
 
-In addition to the three default string interpolators, users can define their own.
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/string-interpolation.md
```diff
@@ -211,6 +211,20 @@ res1: String = a\nb
...
+{% tab 'Scala 2 and 3' for=example-11 %}
+```scala
+scala> val foo = 42
+scala> raw"a\n$foo"
+res1: String = a\n42
+```
+{% endtab %}
+{% endtabs %}
+
+## Advanced Usage
```
adpi2, line 226: This title is repeated below

Flowdalic: good find, fixed

sjrd (approved):

Issue #3158, opened 2025-02-10
Tour: Scastie integration not there


igneus:
https://docs.scala-lang.org/tour/basics.html says:

> Scastie is integrated with some of the code examples in this documentation; if you see a **Run** button in a code example below, click it to directly experiment with the code.

But in fact in the whole Tour there's no such button to be seen. It seems that the paragraph was originally written with ScalaFiddle in mind and then changed to Scastie without making the integration actually work.

- [2ec7e2207022bfeb640f22c216a9664bcd16a3b0](https://github.com/scala/docs.scala-lang/commit/2ec7e2207022bfeb640f22c216a9664bcd16a3b0#diff-b30e67834d09a5cd648dd7c964f3bec82b7a738c1b7370790b9815b214833fabL23) - paragraph updated
- b6d2fd5f98f7380ede80b4b67b688e00af8b5b8f - ScalaFiddle integration removed without replacement

Is it OK to simply remove the false promise, or is it desirable to implement the Scastie integration?

bishabosha:
Hey thanks for noticing, it would be more beneficial to remove the false hope, if some runnable examples ever materialised then we can add back in the notice

PR #3160, opened 2025-02-12
Tour: delete references to the long gone "Run" button

igneus:
ref #3158 

(I understand all the affected European languages enough to know exactly what I'm deleting, but of the Asian languages I don't know any. There I used Google Translate and trusted it blindly.)


adpi2 (approved):
Thank you

PR #3161, opened 2025-02-12
drop link to a removed feed

igneus:
Tests are currently broken, HTMLProofer emits hundreds of errors along the lines of

```
At ./_site/api/all.html:49:

  External link http://scala-lang.org/feed/blog.xml failed (status code 404)
```

That's because the referenced feed has been recently dropped from the site: https://github.com/scala/scala-lang/pull/1750


adpi2 (approved):
Thank you

PR #3169, opened 2025-03-03
Massively simplify getting started with Scala

joan38:
After multiple people telling me how complicated it is using Scala I was bummed to see how complicated are the instructions to install Scala.
Today I passed by an Apple store and asked my wife to record me to show how easy it is to get started with Scala: https://www.youtube.com/shorts/9Un9_rTP0yc

So I really think our "Getting Started" should be dead simple. I know we are proud to show coursier and all the thing we can do with it. But most people don't care, they just want to play with Scala right now and if it's more than one command we already lost half of them to some other language.


bishabosha:
nice video!

sjrd:
The problem with installing *only* scala-cli is that users won't actually be able to use "Scala" in general. They will get confused as soon as they clone and want to run a project built with sbt or Mill (which is virtually all projects).

Also, as I mentioned at https://github.com/VirtusLab/scala-cli/pull/3532, I disagree that committing runners into the repos is good practice. It doesn't work cross-platform: a repo that is created/managed on a Linux platform will not be usable on a Windows platform, and conversely. This actively harms approachability of Scala projects, IMO. This was (still is?) is serious issue with Mill projects, for example. It recommended people commit their `mill` launcher into the repo, but that made those repos completely unusable on Windows.

joan38:
Thanks for your comment @sjrd 
I can see how my view was skewed towards Linux and Mac platforms. Maybe for projects it would be a good habit to also install the bat file.

About installing build tools like SBT (I don't think coursier installs mill), I don't think it's worth complicating scala-lang.org's Getting Started for that. It's the responsibility of the readme of each project to explain how to build the project.
In my opinion the Getting Started page should only worry about Scala and not build tools.

Maybe we could have 2 tabs, install Scala CLI (default tab) and the Scala CLI launcher script in the second tab?
Coursier and build tools would be out of the picture. WDYT?

Ichoran:
I think it's a good idea to show just how easy it can be for cases where you want "easy".  There's no reason to complicate things for people who just want to try it out.  However, it does need to be clearly spelled out that scala-cli, while fantastic, is _not a full build environment_.

I use scala-cli a lot precisely because the install is such a non-issue.  I can write stuff in Scala and almost anyone can use it almost anywhere with almost zero effort.  This completely blows away Python, R, etc..  It's good to play that up, not hide it, just because "most people will want sbt and/or mill".

Just make it very clear what you're getting, and I don't think there's any problem.  If people cannot pay enough attention to **Most scala projects use a full build system, which this is not.  See below (or above) for instructions on how to install common build tools.**, I'm not sure it bodes well for their ability to use Scala at all.

Ichoran:
(In case it's not clear, I mean: we shouldn't delete the instructions on how to install Coursier.  They also need to be in an obvious place and made as clear as possible.  However, we can lead with the simplest, which is scala-cli.)

joan38:
> it does need to be clearly spelled out that scala-cli, while fantastic, is not a full build environment.

I might be mistaken but I would never think of going on scala-lang.org to get documentation on how to use SBT or Mill.
I'd be interested to see how many people get confused by going on scala-lang.org and not getting SBT or Mill doc.
To me there is no confusion on the expectation of the Getting Started of scala's official website to install nothing else than... Scala.

On the other hand right now people who land on the Getting Started of scala-lang.org have to learn about Coursier, need brew installed, then there is a table with sbt, amm, scalafmt... then some info how to `cs launch scala:2.13.16`.
Meanwhile I myself, a Scala dev for the last 12 years, don't have any of that installed on my laptop and seem to do fine. I don't install anything on purpose so that I level down with people onboarding on my projects.

SethTisue:
> we can lead with the simplest, which is scala-cli

I agree. I would like to see this PR merged in some form.

philwalk:
I love the video!

> it would be a good habit to also install the bat file

I assume that very few scala devs working on a Windows system work from a `CMD.EXE` command line (does anybody know?).
If not working from WSL (which is Linux), I'm guessing they tend to work from `MSYS2` or `cygwin`, or `git-bash`.

A batch file isn't adequate by itself, except for the subset of developers who do prefer `CMD.EXE`.
A wrapper helps, but example `scala` scripts tend to look like the following, and it won't work unless `scala.exe` is in the PATH.
```
#!/usr/bin/env -S scala
sys.props.foreach( (k, v) => printf("%-44s: %s\n", k, v) )
```
However, if `scala-cli.exe` is installed, the script can do this works fine, even on Windows:
```
#!/usr/bin/env -S scala-cli shebang
sys.props.foreach( (k, v) => printf("%-44s: %s\n", k, v) )
```

sjrd:
> I assume that very few scala devs working on a Windows system work from a `CMD.EXE` command line (does anybody know?).
> If not working from WSL (which is Linux), I'm guessing they tend to work from `MSYS2` or `cygwin`, or `git-bash`.

`cmd.exe` gives the best experience for sbt. The interactive shell doesn't work so well inside git bash. And WSL is a huge beast. cygwin is nowhere to be seen; it's always been the worst and only got a good run because it was the *only* thing 15 years ago.

philwalk:
> interactive shell doesn't work so well inside git bash

I work mostly in `WSL` and `MINGW64`, which seems quite solid, less experience with `git bash`.

_data/setup-scala.yml
scarf005, line 1: wouldn't this prevent users from using `cs`, `scala`, etc?

sideeffffect:
This is great, would be nice to see this merged. Are here any blockers?

PR #3180, opened 2025-03-27
FAQ: Scala 3 vs 2: tilt it farther towards 3

SethTisue:
I think some of the wordings that made sense circa 2021–3 are no longer the best wordings.

### DIFF
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -44,7 +44,7 @@ reasons for wanting to learn Scala.
 Don't sweat the decision too much.  You can't go far wrong either
 way. It isn't that hard to switch later, in either direction.
 
-That said, you should choose Scala 3 unless you have a specific reason
+Regardless, you should choose Scala 3 unless you have a specific reason
 to need 2. Scala 3 is the future, and it's the best version for
 falling in love with the language and everything it has to offer.
 Scala 3 has plenty of books, plenty of libraries, and high quality
--- a/_overviews/FAQ/index.md
+++ b/_overviews/FAQ/index.md
@@ -44,17 +44,15 @@ reasons for wanting to learn Scala.
 Don't sweat the decision too much.  You can't go far wrong either
 way. It isn't that hard to switch later, in either direction.
 
-Scala 2 remains a common and reasonable choice. Most Scala jobs are
-still Scala 2 jobs. Scala 2 also has a somewhat bigger library
-ecosystem, somewhat better tooling support, and somewhat wider
-availability of learning materials.
-
-That said, these gaps are getting narrower every month.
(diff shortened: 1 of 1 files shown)


BalmungSan (approved):
I also agree we should sell **Scala 3** more.

Two things:

1. You repeat _"That said"_ two times, my formal English is not that good to think of alternatives, but maybe there are better combinators, like using `However` for the second part?

2. Should we also mention `LTS` VS `next`? Maybe in another FAQ entry and link it?

sjrd (approved):

SethTisue:
> You repeat "That said" two times

oops, fixed

> Should we also mention LTS VS next

Hmm, I'm not sure if that's actually an FAQ. It does sometimes come up, but my hope is that people are normally finding their way to Scala Next by default.

In any case, I think it would be a separate entry.

_overviews/FAQ/index.md
```diff
@@ -44,17 +44,15 @@ reasons for wanting to learn Scala.
...
-still Scala 2 jobs. Scala 2 also has a somewhat bigger library
-ecosystem, somewhat better tooling support, and somewhat wider
-availability of learning materials.
-
-That said, these gaps are getting narrower every month.
-Plenty of books are available, a rich library ecosystem already
-exists, and there are more and more jobs using Scala 3.  You should strongly
-consider learning Scala 3. It's still fairly new, released in 2021.
-But it's the future, and it's the best version for
+Regardless, you should choose Scala 3 unless you have a specific reason
```
som-snytt, line 47: `s/Regardless/Irregardless`

_overviews/FAQ/index.md
```diff
@@ -44,17 +44,15 @@ reasons for wanting to learn Scala.
...
+to need 2. Scala 3 is the future, and it's the best version for
 falling in love with the language and everything it has to offer.
```
som-snytt, line 49: They always say an "older version" is better for falling in love, especially your first time.

_overviews/FAQ/index.md
```diff
@@ -44,17 +44,15 @@ reasons for wanting to learn Scala.
...
+Scala 3 has plenty of books, plenty of libraries, and high quality
+tooling.
+
+That said, many Scala jobs are still Scala 2 jobs. In most cases, the
+cause of that is simply inertia, especially at large shops. (But it can
+sometimes be due to availability of specific libraries.)
```
som-snytt, line 55: This makes Scala shops sound stodgy and cumbersome.

> Many legacy Scala shops need a kick in the pants to upgrade. Since they are eager for young talent, you need only tell them that you're "only comfortable coding in Scala 3" to flip that script.

PR #3181, opened 2025-03-27
Add generator expr.s to Scala for Python Devs

### LINKED ISSUE #3178: Scala for Python Developers should show equivalent of generator expressions
The [Scala for Python Developers](https://docs.scala-lang.org/scala3/book/scala-for-python-devs.html) page does have [a section](https://docs.scala-lang.org/scala3/book/scala-for-python-devs.html#comprehensions) comparing Scala's comprehensions to Python's list comprehensions, but (AFAICT) lacks any information about what Scala's analogue is to Python's [generator expressions](https://docs.python.org/3/tutorial/classes.html#generator-expressions), which are like list comprehensions but lazily evaluated.

From the answers to [a StackOverflow question](https://stackoverflow.com/q/6996093/2748899) about this, I got that in Scala, lazy evaluation or not is decided by the object being iterated over, because comprehensions just desugar to `object.map()` calls and the like, which of course get to determine the output type in full (unlike Python, where iterated-over objects only get to determine the element type).

It would be nice if one equivalent to a generator expression using a `.view` or something like that was briefly demonstrated, maybe with a short comment explaining this (although it seems like the existing examples are kept comment-free on purpose, so maybe not if the goal is to really keep it short - but in that case, maybe there could be a longer explanation further down the page).

smheidrich:
Fixes #3178

### DIFF
--- a/_overviews/scala3-book/scala-for-python-devs.md
+++ b/_overviews/scala3-book/scala-for-python-devs.md
@@ -40,7 +40,7 @@ At a high level, Scala shares these *similarities* with Python:
 - Both have a relatively simple, concise syntax
 - Both support a [functional style of programming][fp-intro]
 - Both are object-oriented programming (OOP) languages
-- Both have comprehensions: Python has list comprehensions and Scala has `for` comprehensions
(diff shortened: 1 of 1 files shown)


SethTisue:
Thank you!

PR #3207, opened 2025-08-28
docs(implicit-conversions): use new given syntax

scarf005:
given syntax was changed to use `:` over `with` since scala 3.6: https://docs.scala-lang.org/scala3/reference/contextual/givens.html


bishabosha (approved):
Thanks!

PR #3215, opened 2025-09-17
update JDK compat page for JDK 25 going GA

SethTisue:
but I left the "Tooling compatibility table" section alone for now since I don't have that information yet. I think we should merge this regardless.

Also I haven't added "JDK 26 (ea)" yet even though early-access builds do exist; that can come later.


SethTisue:
@Gedochao this page remains rather Scala 2 centric. I haven't aimed to correct that here, but maybe it's something that would interest you 🤷

PR #3230, opened 2025-11-11
various updates to JDK compat guide

SethTisue:
I have chosen not to add a JDK 26 row yet, as it's still so early in the JDK 26 cycle.

### DIFF
--- a/_overviews/jdk-compatibility/overview.md
+++ b/_overviews/jdk-compatibility/overview.md
@@ -10,7 +10,7 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
 
 ## Scala 3 compatibility
 
-At present, both Scala 3.3 LTS and Scala 3.4 through 3.7 still support JDK 8 and above.
+Scala 3 versions from 3.3 LTS through Scala 3.7 support JDK 8 and above.
 
 As per [this blog post](https://www.scala-lang.org/news/next-scala-lts-jdk.html),
 Scala 3.8 will have a new minimum JDK version of 17.


Gedochao (approved):
LGTM

_overviews/jdk-compatibility/overview.md
```diff
@@ -12,13 +12,13 @@ Sometimes new JVM and JDK (Java Development Kit) versions require us to update S
...
 Minimum Scala versions:
 
-| JDK         | 3        | 3 LTS    | 2.13      | 2.12      | 2.11       |
-|:-----------:|:--------:|:--------:|:---------:|:---------:|:----------:|
-| 25 (LTS)    | 3.7.1    | 3.3.6    | 2.13.17   | 2.12.21*  |            |
-| 21 (LTS)    | 3.4.0    | 3.3.1    | 2.13.11   | 2.12.18   |            |
-| 17 (LTS)    | 3.0.0    | 3.3.0    | 2.13.6    | 2.12.15   |            |
-| 11 (LTS)    | 3.0.0    | 3.3.0    | 2.13.0    | 2.12.4    | 2.11.12    |
-| 8 (LTS)     | 3.0.0    | 3.3.0    | 2.13.0    | 2.12.0    | 2.11.0     |
+| JDK         | 3.8*   | 3.4+     | 3.3 LTS  | 2.13      | 2.12      |
```
lrytz: I find the `3.4+` column confusing, because it lists versions like `3.0.0` and `3.7.1`. Also I'm not sure this table has the right format to fit in the information about 3.8 requiring Java 17.

Maybe leave it as is, and move the existing note that 3.8 requires Java 17 closer to the table / somewhere more prominent?

lrytz: The table is "minimum Scala versions" (that works on some Java version) but for 3.8 it's actually about "minimum Java version" (required for some Scala version), I guess that mix confuses me a bit

SethTisue: We had something similar in the Scala 2.11 column — deleted in this PR — where that column served both to indicate the minimum supported Scala versions on JDK 8 and 11, _and_ the maximum allowable JDK version (17+ doesn't work, so those cells were blank).

Tell you what, I'll leave the new column in place, but I'll _also_ take your suggestion to add explanatory text close by.

SethTisue:
I separated the tables for Scala 3 and Scala 2, and that allowed me to move the Scala 3 information that had been all the way at the bottom of the page, up to the top where it belongs.

PR #3233, opened 2025-11-18
run mdoc with newer Toolkit version (0.7.0, was 0.1.7)

SethTisue:


bishabosha:
this i think can be solved with `mdoc:fail` if example.com is expected to reject POST always, or swap to `httpbin.org` (which is flaky but at least has an endpoint)

PR #3249, opened 2026-01-05
Fix incorrect reference to non-existent tailOption method

### LINKED ISSUE #2338: tailsOption do
Written in this part of the  scala3 book [collections-methods](https://docs.scala-lang.org/scala3/book/collections-methods.html)
> tail throws an java.lang.UnsupportedOperationException if the list is empty, so just like head and headOption, there’s also a tailOption method, which is preferred in functional programming.
But we do not have a method with  name tailOption in the collections

Dev-CodeBox:
Fixes #2338

Removes incorrect reference to a non-existent `tailOption` method and
documents safe alternatives for working with collection tails.

### DIFF
--- a/_overviews/scala3-book/collections-methods.md
+++ b/_overviews/scala3-book/collections-methods.md
@@ -385,7 +385,10 @@ Just like `head`, `tail` also works on strings:
 
 {% endtabs %}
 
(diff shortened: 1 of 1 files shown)


SethTisue:
https://github.com/scala/docs.scala-lang/pull/3261 includes a commit that fixes some of the 404s. but yes, this and other PRs are mergeable regardless

PR #3260, opened 2026-01-13
Add pattern matching explanation to string interpolation page

Sporarum:
I had a lot of trouble finding documentation about this subject, so here is my contribution towards making it better !

### DIFF
--- a/_overviews/scala3-book/string-interpolation.md
+++ b/_overviews/scala3-book/string-interpolation.md
@@ -427,7 +427,7 @@ pt match case p"$a,$b" => a + b // a = 1, b = 2
 {% endtab %}
 {% endtabs %}
 
-We'd create a custom `p`-extractor which extracts two coordinates by first implementing a `StringContext` extension with something like:
+We'd create a custom extractor which extracts two coordinates, and make it act as a member `p` of `StringContext` with something like:
 
 {% tabs custom-interpolator-2-pattern class=tabs-scala-version %}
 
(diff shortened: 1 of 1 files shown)


_overviews/scala3-book/string-interpolation.md
```diff
@@ -365,6 +366,200 @@ of custom interpolators as above can allow for powerful syntactic shorthand, and
...
+
+Note however there are not extractors by default for the `f` and `raw` interpolators, so neither `case f"..."` nor `case raw"..."` will work.
+(Unless a library provided them.)
+
+Anytime the compiler encounters a processed string pattern of the form:
+
+{% tabs example-pattern %}
+{% tab 'Scala 2 and 3' for=example-pattern %}
+```scala
+id"string content"
```
bishabosha: jusgement by taste of course as this may imply not nestable (i hope not)
```suggestion
case id"string content" =>
```

Sporarum: I hesitated as well, it's hard to talk about patterns for this reason, they only appear inside `case`, but it is not technically a part of them

So I don't know what's best ?

_overviews/scala3-book/string-interpolation.md
```diff
@@ -365,6 +366,208 @@ of custom interpolators as above can allow for powerful syntactic shorthand, and
...
+{% endtab %}
+
+{% tab 'Scala 3' for=custom-interpolator-2-pattern %}
+```scala
+import scala.language.implicitConversions
+
+given Conversion[StringContext, PointHelper] = sc => PointHelper(sc)
+
+class PointHelper(val sc: StringContext):
+  object p:
```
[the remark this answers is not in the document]
Sporarum: What do you think of what I replaced it with ?
I tried to condense it further by not creating an extractor class, but did not manage to by using structural types, since those also require a way to do dynamic dispatch
See also the SLC I opened as a consequence: https://contributors.scala-lang.org/t/slc-optional-traits-to-simplify-creation-of-extractor-objects/7343

bishabosha (approved):
very nice!

PR #3270, opened 2026-01-22
Add Higher-Kinded Types section to Scala 2 Tour

Dev-CodeBox:
This PR adds a new section `Higher-Kinded Types` to the Scala 2 Tour documentation.

It includes:
- Higher-Kinded Types (HKT) with Functor example
- Type Bounds (upper/lower) with sample code
- Type Projections with example classes

This addresses issue #1269 and provides simple explanations and examples for learners.

### DIFF
--- a/_tour/higher-kinded-types.md
+++ b/_tour/higher-kinded-types.md
@@ -1,17 +1,14 @@
 ---
-layout: multipage-overview
-title: Advanced Type Features
-partof: scaladoc
-overview-name: Scaladoc
-num: 4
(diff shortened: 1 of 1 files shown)


_overviews/higher-kinded-types.md
```diff
@@ -0,0 +1,61 @@
+---
+layout: multipage-overview
+title: Advanced Type Features
+partof: scaladoc
```
bishabosha: why is this in the Scaladoc guide and not Tour of Scala?

PR #3292, opened 2026-04-01
JDK compat page: JDK 26 support landed in Scala 3.8

SethTisue:
I didn't bother adding a whole JDK 26 section, though we probably should at some point.


SethTisue:
oops, I confused myself about the meaning of the columns. it's fine as it stands

the table correctly shows that JDK 26 support never landed in the 3.4–3.7 version range, and that's fine

at some point Scala 3.3.8 will become available and it will support the whole 8–26 spectrum.

PR #3299, opened 2026-04-30
Document the new fuel limit

SolalPirelli:


_overviews/compiler-options/compiling-deeply-nested-code.md
```diff
@@ -0,0 +1,63 @@
...
+This depends on where you're running the compiler, _not_ on what target you're compiling for.
+For instance, if you're running the Scala compiler on your laptop using `scala-cli` to target Scala.js, you should follow the JVM instructions.
+
+On the **JVM**, pass `-Xss...` to the JVM, for example `-Xss16M`.
+How to do this depends on what IDE and/or build tool you are using. For sbt, add it to `.jvmopts`.
+
+On **Scala.js**, you cannot increase the stack space, because neither JavaScript nor WebAssembly allow this.
+You can try increasing the fuel limit anyway, but you may have to instead limit your use of recursive types.
+
+On **Scala Native**, ...TODO what?...
```
SolalPirelli: @WojciechMazur do we support running the compiler itself with Scala Native? What's the stack space story there?

_overviews/compiler-options/fuel-and-stack-overflows.md
```diff
@@ -0,0 +1,63 @@
...
+title: Fuel and Stack Overflows
+versionSpecific: false
+---
+
+## Fuel and Stack Overflows
+
+Scala lets you, the programmer, write deeply nested types if you find it necessary to model your business logic.
+
+Because types can refer to other types in ways that recurse without the base case being obvious, it's possible to write incorrect code whose recursion never terminates,
+or correct code whose recursion is so deep that the compiler could run out of stack space while compiling.
```
mbovel: I would not mention “to model your business logic” here, I don't think it brings anything to this discussion; the reason for which one might use deeply nested types is not our concern.

I would also not mention "base case", I'd expect this term not to be known by non-academic folks.

```suggestion
Scala's type system is expressive enough to describe types with deeply nested or recursive structure. When the compiler operates on such types, it may recurse very deeply and eventually exhaust its allocated stack. This can happen with incorrect code whose recursion never terminates, but also with correct code whose recursion is simply too deep.
```

_overviews/compiler-options/compiling-deeply-nested-code.md
```diff
@@ -0,0 +1,63 @@
...
+This ensures the compiler does not run out of stack space, which would cause it to crash instead.
+
+**In Scala 2 and Scala 3.8 or earlier**, the compiler attempts to catch stack overflows, which is not always reliable.
+
+### Did I hit a compiler bug?
+
+For **Scala 3.9 and newer**:
+- If you get a `StackOverflowError`, and you did **not** set `-Xmax-fuel`, nor any JVM option related to stack space (when running on the JVM),
+  yes, that's a compiler bug, because the compiler should run out of fuel before hitting a stack overflow.
+  Please [report it](https://github.com/scala/scala3/issues/new/choose).
```
mbovel, line 22: I am still not convinced we can nor should make that promise.

_overviews/compiler-options/compiling-deeply-nested-code.md
```diff
@@ -0,0 +1,60 @@
...
+This means your types contain unusually deep recursion.
+
+First, think about whether this is intentional. Perhaps you accidentally wrote a match type that recurses infinitely because it has no base case, for instance.
+
+If it is intentional, you must increase the fuel limit, **and** increase the stack space if you are on a platform that allows this, see below.
+
+### Increasing the fuel limit (Scala 3.9+)
+
+To increase the fuel limit, pass `-Xmax-fuel:...` to the compiler.
+You can check the default limit with ...TODO how?...
```
SolalPirelli: would be nice to not have to tell people "run scalac -X and look for the right option"

SethTisue:
Maybe "Compiling Deeply Nested Code" (the goal) would be a better title than "Fuel and Stack Overflows" (the means)?

SolalPirelli:
@SethTisue good idea, done

_overviews/compiler-options/compiling-deeply-nested-code.md
WojciechMazur: Not yet, I heard there was some advance in running full compiler, with macros, on Scala.js. In the future we might try to do the same with Native, but I don't think it's in the current scope of work. 

However, Scala Native can control stack space natively (for the main thread) with system specific settings (altough default typically quite large, like 16MB), and for java.lang.Thread with env variable (SCALANATIVE_THREAD_STACK_SIZE=4M  simulating -Xss on the JVM) so not an issue for us. 
SN has some experimental support for detecting and throwing StackOverflowError (altough there is some runtime cost currently for methods that are known to cycle)

SolalPirelli: OK, let me remove this bit here then, thanks for the explanation

_overviews/compiler-options/compiling-deeply-nested-code.md
```diff
@@ -0,0 +1,58 @@
...
+Recursive operations in the compiler consume fuel, and once a predefined fuel limit is reached, compilation fails.
+This ensures the compiler does not run out of stack space, which would cause it to crash instead.
+
+**In Scala 2 and Scala 3.8 or earlier**, the compiler attempts to catch stack overflows, which is not always reliable.
+
+### Did I hit a compiler bug?
+
+For **Scala 3.9 and newer**:
+- If you get a `StackOverflowError`, and you did **not** set `-Xmax-fuel`, nor any JVM option related to stack space (when running on the JVM),
+  yes, that's a compiler bug, because the compiler should run out of fuel before hitting a stack overflow.
```
bishabosha, line 21: is parser guarded by fuel?

_overviews/compiler-options/compiling-deeply-nested-code.md
bishabosha: `scala -X` is cross portable

_overviews/compiler-options/compiling-deeply-nested-code.md
SolalPirelli: everything that currently uses `handleRecursive` on main to catch stack overflow is being migrated to use the fuel, doesn't look like this includes parser files though? https://github.com/scala/scala3/pull/25937/changes

_overviews/compiler-options/compiling-deeply-nested-code.md
SolalPirelli: https://github.com/scala/scala3/pull/26052

SolalPirelli:
```suggestion
You can check the default limit by calling the compiler with `-Xmax-fuel:help`.
```

SolalPirelli:
I need this to merge the fuel PR since the error message refers to it, so I'll merge now, but of course feedback is always welcome if anyone has suggestions to make this clearer.

PR #3305, opened 2026-06-18
Update grammar and wording in the `Iterable` trait tutorial

neuroevolutus:
One small change in particular that may be worth double-checking was the replacement of a double hyphen in the Markdown source with an em dash. The change precedes the words "dropping the ordering..." in the bullet for "Element retrieval".


neuroevolutus:
I think the HTMLProofer step of the CI is failing (at least in part) due to a redirect from an old LAMP EPFL GitHub link. Edit: Never mind; the proofer runs seem to be flaky since that site seemed to yield no error on a subsequent run.

Also, the top-level page for the scala-exercises.org site seems to lead to a non-English page. It looks dubious, but, then again, I may be wrong since I don't know the language in which it's written.

The Gluon site is fine although the https://gluonhq.com/products/mobile/attach/ page in particular no longer exists.

I'm a bit befuddled since I haven't changed any of the links.

SethTisue:
Yeah you can just ignore the htmlproofer failures, I'll deal with it.

